PR #127, opened 2017-03-18
Get custom colors from appstream and set the appinfoview banner.

nathandyer:


src/Core/Package.vala
```diff
@@ -316,6 +318,27 @@ public class AppCenterCore.Package : Object {
...
+    public string? get_color_primary_text () {
+        if (color_primary_text != null) {
+            return color_primary_text;
+        }
+
+        color_primary_text = component.get_custom_value ("x-appcenter-color-primary-text");
+
+        return color_primary_text;
+    }
+
```
danrabbit: extra white space

src/Views/AppInfoView.vala
```diff
@@ -211,6 +226,27 @@ namespace AppCenter.Views {
             });
         }
 
+        private void reload_css () {
+            var provider = new Gtk.CssProvider ();
+            try {
+                string color_primary = package.get_color_primary ();
+                if (color_primary == null) {
+                    color_primary = "rgba(66,83,129,255)";
```
danrabbit: since this is 255, is there a reason to use RGBA instead of hex? Probably just #68758e is good

src/Views/AppInfoView.vala
```diff
@@ -18,9 +18,17 @@
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
 
+const string BANNER_STYLE_CSS = """
+    .banner {
+        background-color: %s;
+        color: %s;
+    }
+    """;
```
danrabbit: We might need to at least add 
```
.banner .button {
	background-color: @base_color;
}
```

PR #128, opened 2017-03-18
Error install notifications

### LINKED ISSUE #47: No error feedback
When installation fails for some reason, there should be some kind of dialog that would show error details, currently there's no feedback at all, nothing happens when the action failed, e.g: try to install Gimp.

Launchpad Details: [#LP1598608](https://bugs.launchpad.net/bugs/1598608) Adam Bieńkowski - 2016-07-03 18:41:21 +0000

donadigo:
Fixes #47 

Show a notification when there was an error to the user, the notification is shown always when an error happens not depending on the window state. Furthermore, the notification code was moved to MainWindow to provide easier access to application and the window itself. It's now possible to have notifications sent for different actions in one place in the code.


src/MainWindow.vala
```diff
@@ -311,4 +314,44 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+                    }                    
+                } else {
+                    // Check if permission was denied or the operation was cancelled
+                    if (error.matches (IOError.quark (), 19) || error.matches (Pk.ClientError.quark (), 303)) {
+                        break;
+                    }
+
+                    var close_button = new Gtk.Button.with_label (_("Close"));
+
+                    var dialog = new MessageDialog (_("Application installation failed"), _("There was an error installing %s:\n%s").printf (package.get_name (), error.message), "dialog-error");
```
danrabbit: Please use Title Case here for the primary text.

Is there a reason that we're not using the more specific "There was an error installing %s" for the primary text? Seems like that would flow better and then just use the error message as secondary text.

PR #133, opened 2017-03-19
Homepage redesign and new app banner implementation.

nathandyer:
Still waiting on the production server URL from @btkostner, but otherwise this should be ready for review


src/Widgets/Banner.vala
```diff
@@ -0,0 +1,167 @@
...
+            name_label = new Gtk.Label ("Name");
+            name_label.get_style_context ().add_class ("h1");
+            summary_label = new Gtk.Label ("Summary");
+            summary_label.get_style_context ().add_class ("h2");
+            description_label = new Gtk.Label ("Description");
+            description_label.get_style_context ().add_class ("h3");
+            name_label.xalign = 0;
+            summary_label.xalign = 0;
+            description_label.xalign = 0;
+            description_label.margin_top = 25;
```
danrabbit: Can we group these instead of having them mixed up? As in everything to do with name_label together, everything to do with summary_label together etc.

We also need to set wrap = true and a max character width (something between 40 and 60)

src/Widgets/Banner.vala
```diff
@@ -0,0 +1,167 @@
...
+
+            icon = new Gtk.Image ();
+            icon.icon_name = "system-software-install";
+            icon.pixel_size = 128;
+            icon.xalign = 1;
+            icon.margin = 25;
+            content_box = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 12);
+            var vertical_box = new Gtk.Box (Gtk.Orientation.VERTICAL, 0);
+
+            var read_more_button = new Gtk.Label (_("Read More"));
```
danrabbit: Can we add the character › at the end here :)

src/Core/Houston.vala
```diff
@@ -0,0 +1,71 @@
...
+        session = new Soup.Session ();
+    }
+
+    private Json.Object process_response (string res) throws Error {
+        var parser = new Json.Parser ();
+        parser.load_from_data (res, -1);
+
+        var root = parser.get_root ().get_object ();
+
+        if (root.has_member ("errors") && root.get_array_member ("errors").get_length () > 0) {
```
donadigo: Just call `get_array_member` only once.

src/Homepage.vala
```diff
@@ -0,0 +1,75 @@
...
+    public class Homepage: Gtk.Box {
+
+        public signal void package_selected (AppCenterCore.Package package);
+
+        public Widgets.Banner newest_banner;
+        public AppCenter.Views.CategoryView category_view;
+        private Gtk.ScrolledWindow scrolled_window;
+
+        public Homepage () {
+
```
donadigo: Remove this empty line.

donadigo:
The branch throws these fatals:
[FATAL 23:20:58.677708] app_center_core_houston_process_response: assertion 'res != NULL' failed
[FATAL 23:20:58.677735] [Json] json_object_get_array_member: assertion 'object != NULL' failed
[FATAL 23:20:58.677746] [Json] json_array_get_elements: assertion 'array != NULL' failed

I think these shouldn't happen even if houston is down or not responding.

src/Core/Houston.vala
btkostner: Without checking if it `has_member` we get JSON assert errors

PR #137, opened 2017-03-19
Add margin to bottom of app header

btkostner:
This doubles the margin to the bottom of the app view header. This balances the look.


danrabbit:
Probably what we should do here instead is use Gtk.Grid instead of box and set row_spacing to 24. This is a bit more future proof for adding other sections

PR #147, opened 2017-03-21
Banner simplification

danrabbit:


src/Widgets/Banner.vala
```diff
@@ -105,63 +104,52 @@ namespace AppCenter.Widgets {
...
+            add (grid);
         }
 
         public void set_brand () {
-            name_label.label = "AppCenter";
-            summary_label.label = "An open, pay-what-you-want app store";
-            description_label.label = "Try first, then pay what you want. Get the apps that you need, for a price you can afford.";
-
-            background_color = DEFAULT_BANNER_COLOR_PRIMARY;
+            background_color = "#665888";
```
Philip-Scott: Remove the constant?

danrabbit: So there's actually two things going on here. One is the default color for apps that don't provide a background color and one is the background color for the AppCenter placeholder banner

btkostner: There might be a case in which we have to go back to the default AppCenter banner (if the user installs all the banner apps)

PR #151, opened 2017-03-22
Use more modern way to create Threads

tintou:


src/Dialogs/StripeDialog.vala
```diff
@@ -344,9 +344,8 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         }
     }
 
-    private async void on_pay_clicked () {
-        SourceFunc callback = on_pay_clicked.callback;
-        ThreadFunc<void*> run = () => {
+    private void on_pay_clicked () {
```
Philip-Scott: Removed async, is that correct?

tintou: Yes, it's not used at all here

PR #157, opened 2017-03-24
Allow uninstallation when an update is available

### LINKED ISSUE #9: Old apps need to be updated before they can be uninstalled [$25]
In AppCenter, when wanting to uninstall an app that is not of the latest version, no option to uninstall is presented until after the app is up to date. This means that to remove an app, you're basically first required to download the app. It would be good to have a way to uninstall without updating first, especially for larger apps.

Launchpad Details: [#LP1667321](https://bugs.launchpad.net/bugs/1667321) Yhilan Ko - 2017-02-23 14:41:42 +0000

<bountysource-plugin>

---
There is a **[$25 open bounty](https://www.bountysource.com/issues/43123412-old-apps-need-to-be-updated-before-they-can-be-uninstalled?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Philip-Scott:
Adam was right. Stacks were actually more messy. And this will more easily allow so that i can lazy-load things in the future (So not all the UI get's build if it's not needed)

Fixes #9


src/Widgets/AbstractAppContainer.vala
```diff
@@ -71,9 +70,6 @@ namespace AppCenter {
             package_name = new Gtk.Label ("");
             image = new Gtk.Image ();
 
-            cancel_button = new Gtk.Button.with_label (_("Cancel"));
-            cancel_button.clicked.connect (() => action_cancelled ());
```
Philip-Scott: Yeah, we had two cancel buttons for some reason ;P

src/Widgets/AbstractAppContainer.vala
```diff
@@ -163,88 +157,57 @@ namespace AppCenter {
...
-                    open_button.hide ();
+                    action_stack.set_visible_child_name ("progress");
                     break;
 
                 default:
                     assert_not_reached ();
             }
         }
 
+        private void set_button_state (Gtk.Widget widget, bool show) {
```
donadigo: Since this function isn't using anything from the instance, it could perhaps be `static`?

I guess, you could also make use of `visible` instead of `show ()` and `hide ()`:
widget.visible = show;

But it's really optional, you can keep how it is, though I would like to change `show` to `show_all`

src/Widgets/AbstractAppContainer.vala
```diff
@@ -165,88 +159,52 @@ namespace AppCenter {
...
-                    open_button.hide ();
+                    action_stack.set_visible_child_name ("progress");
                     break;
 
                 default:
                     assert_not_reached ();
             }
         }
 
+        private static void set_button_state (Gtk.Widget widget, bool show) {
```
danrabbit: Can we rename this to something like `set_button_visibility` ? "state" is too open ended. It doesn't really make sense to say "state = false". State could be anything

davidmhewitt:
I've re-tested the bug that @donadigo mentioned and can confirm that when cancelling an un-install of an upgradeable app, the update button is still visible afterwards. So it should be good to go now.

PR #158, opened 2017-03-24
Homepage: fixed infinte loop if candidate_package = null

GiltosMx:


donadigo:
@tintou is right. You should `break` instead of going further.

GiltosMx:
Done, you're right, missed it. I think it's fixed now, let me know if anything else comes up.

PR #173, opened 2017-03-25
StripeDialog: Fix CC validation false negatives

### LINKED ISSUE #162: Card check is too strict [$50]
The current card number check appears to be hitting false negatives; i.e. my personal Visa debit card is not seen as valid. We might be too strict or have a bug with the client-side validation.

Philip-Scott:
Fixes #162

Reason of the bug: Just take the last digit at the end. A 10 was causing it to fail since it was comparing it to a 0

### DIFF
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -326,7 +326,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
             sum += number;
         }
 
-        return 10 - (sum % 10) == hash;
+        return (10 - (sum % 10)) % 10 == hash;
     }
 
     private void on_response (Gtk.Dialog source, int response_id) {


Philip-Scott:
Putting this here for the future:

```
public class CCValadator {
    public static bool is_card_valid (string numbers) {
        var char_count = numbers.char_count ();

        if (char_count < 14) return false;

        int hash = int.parse (numbers[char_count - 1].to_string ());

        int j = 1;
        int sum = 0;
        for (int i = char_count -1; i > 0; i--) {
            var number = int.parse (numbers[i-1:i]);
            if (j++ % 2 == 1) {
                number = number * 2;
                if (number > 9) {
                    number = number - 9;
                }
            }

            sum += number;
        }

        return (10 - (sum % 10)) == hash;
    }

    public static void main () {
        string[] cc_list = {
            "4532939667690661",
            "4485720698638136",
            "4024007150791803",
            "4929373336459491",
            "4485023285523484",
            "5229683747303003",
            "5348660838050244",
            "5445580219970657",
            "5147683293954042",
            "5397447123179627",
            "6011531498173333",
            "6011173459062500",
            "6011207551626331",
            "6011817127808880",
            "6011723938239051",
            "375556057611178",
            "340425695008944",
            "371585358922756",
            "345152639402596",
            "344425604997530",
            "4024007163674335",
            "4539765369685144",
            "4916202919943040",
            "4024007161262075"
        };

        foreach (var cc in cc_list) {
            if (!is_card_valid (cc)) {
                error ("CC not passed: %s", cc);
            }
        }

        stdout.printf ("Test compleate, all cards passed\n");
    }
}
```

Issue #130, opened 2017-03-19
[$25] Restart Required should be an infobar not a header


danrabbit:
At the moment the Restart Required messaging in the updates view is a list header. Instead it should be an infobar at the top of the view

codygarver:
There is a bounty for this issue [on Bountysource](https://www.bountysource.com/issues/43200147-restart-required-should-be-an-infobar-not-a-header).

Philip-Scott:
@donadigo: if you could solve as well #131 with this it would be great for testing :)

davidmhewitt:
Closed via #175

PR #175, opened 2017-03-25
Make the restart button appear in the InfoBar

donadigo:
Makes the restart button appear in the InfoBar instead of the header. **This again needed some refactoring how headers work and is in large portion just a port of this branch on LP + a rework how restart is checked:** https://code.launchpad.net/~donadigo/appcenter/sorting-headers/+merge/317787

Previously AppCenter updated the list every 20 milliseconds, but this is now gone with `set_header` method of `Gtk.ListBox`. This was also used to update the restart state, which is why it's now reworked.  Instead of refreshing it's state with fixed interval, we use `FileMonitor` to monitor if the `reboot-required` file is present. The `AppList` only binds the `restart_required` property to it's `InfoBar`.

You can test the restart detection with these commands:
1. Make AppCenter show the restart button:
`sudo touch /var/run/reboot-required`

2. Make AppCenter hide the restart button:
`sudo rm /var/run/reboot-required`

### DIFF
--- a/src/Core/Client.vala
+++ b/src/Core/Client.vala
@@ -23,6 +23,7 @@ public class AppCenterCore.Client : Object {
     public bool connected { public get; private set; }
     public bool updating_cache { public get; private set; }
     public uint task_count { public get; private set; default = 0; }
+    public bool restart_required { public get; private set; default = false; }
 
     public AppCenterCore.Package os_updates { public get; private set; }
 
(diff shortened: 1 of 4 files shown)


donadigo:
@danrabbit done.

danrabbit:
@donadigo you're so fast :astonished:

davidmhewitt:
Fixes #130

PR #176, opened 2017-03-25
Only show progress bar for updating apps

### LINKED ISSUE #166: [$50] Don't display update progress bar when "waiting" or "finished"
In the updates view, we should only display a progress bar when it's an app that's being updated. If its state is waiting or finished, we don't need a progress bar or label.

davidmhewitt:
Fixes #166.

Hide the progress bar for apps that are in waiting or complete status on the update screen.

Also changes the sorting order so that the apps in the list are displayed in the order they are going to be updated in after clicking "Update All". This allows the user to follow the progress of multiple updates down the screen rather than it jumping about due to the previous alphabetical order and the apps not being updated alphabetically.

### DIFF
--- a/src/Views/AppListView.vala
+++ b/src/Views/AppListView.vala
@@ -159,6 +159,13 @@ namespace AppCenter.Views {
 
         [CCode (instance_pos = -1)]
         protected override int package_row_compare (Widgets.AppListRow row1, Widgets.AppListRow row2) {
+            bool a_is_updating = row1.get_is_updating ();
+            bool b_is_updating = row2.get_is_updating ();
+
+            if (a_is_updating || b_is_updating) {
+                return a_is_updating ? -1 : 1;
+            }
+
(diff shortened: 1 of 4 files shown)


davidmhewitt:
@danrabbit I left that one in because it felt semi-useful, but I'm happy to remove it. Should I get rid of "Waiting for package manager lock" too? It could potentially be in that state for a while though, which may be confusing.

danrabbit:
Hm yeah that's a good question. I guess I would say that the difference is one is waiting for user interaction and one is waiting on the system. So "Waiting for authentication" is probably less valuable because we show the auth dialog as a form of feedback, but "Waiting for package manager lock" is important feedback because there's nothing the user can do here and no other indication of what is happening.

davidmhewitt:
@danrabbit Done!

danrabbit:
Nice! I feel like the switch makes this code more legible as well :) This works as expected as far as I can tell. Thanks!

PR #180, opened 2017-03-26
Use right reversed elementary scheme

### LINKED ISSUE #8: use the right reversed elementary Scheme
We do not own pantheon.org, so our GSettings living under org.pantheon does not follow RDNN (https://en.wikipedia.org/wiki/Reverse_domain_name_notation).

In Loki+1, we should switch to using io.elementary for shell/OS components, and com.github.elementary for first-party apps.

Things to update:
 * Application id
 * Desktop filename
 * AppStream filename and id
 * GSettings path and id
 * DBus interfaces

Launchpad Details: [#LP1594598](https://bugs.launchpad.net/bugs/1594598) Cassidy James Blaede - 2016-06-20 22:26:20 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123404-use-the-right-reversed-elementary-scheme?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Hyuchia:
Fix #8

## Changelog:
- Replace all org.pantheon schemes to io.elementary
- Replace all paths /org/pantheon to /io/elementary
- Rename files to match the new scheme


danrabbit:
We'll want to check with @codygarver first. I think he wanted to wait to do this until we start making breaking changes for Juno

danrabbit:
Yeah, since we don't technically own pantheon.org I think `io.elementary` is the more proper and accurate RDNN

Issue #35, opened 2017-03-16
Uninstall apps from context menu in Slingshot


elementaryBot:
Slingshot should have the ability to uninstall an app from the app's context menu.

Launchpad Details: [#LP1211570](https://bugs.launchpad.net/bugs/1211570) Jung-Kyu Park - 2013-08-13 00:10:57 +0000

elementaryBot:
Agreed, this should be a feature along with quicklist in the context menu.

Launchpad Details: [#LPC](https://bugs.launchpad.net/slingshot/+bug/1211570/comments/1) Cassidy James Blaede - 2013-08-16 07:12:10 +0000

elementaryBot:
$15 bounty added at https://www.bountysource.com/issues/919741-remove-programs-directly-from-slingshot

Launchpad Details: [#LPC](https://bugs.launchpad.net/slingshot/+bug/1211570/comments/2) Cassidy James Blaede - 2016-09-13 01:21:23 +0000

elementaryBot:
Added AppCenter here in case there is some sort of coordination needed. Feel free to mark as invalid on that project if I am wrong.

Launchpad Details: [#LPC](https://bugs.launchpad.net/slingshot/+bug/1211570/comments/3) Cassidy James Blaede - 2016-09-13 01:22:45 +0000

elementaryBot:
Adding new menu item for each application is not very hard. What about interaction with AppCenter? If I press "Uninstall" on "convert" it must find package name "imagemagick" (maybe via apt?), than somehow pass this name to AppCenter, which must ask user a confirmation and then delete the package. Must I implement command-line option for AppCenter? The solution "gksudo apt-get remove ..." is not acceptable, is not it?

Launchpad Details: [#LPC](https://bugs.launchpad.net/slingshot/+bug/1211570/comments/4) Dmitry  DeXPeriX  Hrabrov - 2016-09-20 11:21:18 +0000

elementaryBot:
My guess is that we shouldn't depend on AppCenter here and just use packagekit directly. But I'll bring this to Corentin's attention for his opinion

Launchpad Details: [#LPC](https://bugs.launchpad.net/slingshot/+bug/1211570/comments/5) Daniel Fore - 2016-12-07 00:31:15 +0000

elementaryBot:
From Slack:

Corentin: Hmm it's probably better to use AppCenter because you can search the package with AppStream

[...]

PR #183, opened 2017-03-26
Add DBus interface for managing AppStream components

donadigo:
This branch adds a new own AppCenter DBus interface "org.pantheon.appcenter". This will enable other OS components to access basic functions like:
* Install, update, remove packages
* Finding AppStream components for desktop ID's
* Searching components for a specific query

The interface is pretty simple for now, but I would expect that in the future it would serve much more sophisticated objects and methods for interacting with AppCenter / AppStream.

Each function has it's comment as a short summary of what it does and returns to not cause confusion.

Note that, after calling e.g `install ()`, AppCenter won't launch, instead an authentication dialog will pop up and after it, the app will begin installing. After installation a notification will be sent / error dialog will be shown just like you would install it directly from AppCenter. Because we're using our own interface, calls like `install`, `update` or `uninstall` will cause displaying progress directly in the AppCenter's UI if the user would like to check the progress.

The `org.pantheon.appcenter` interface is exposed in session bus at path named `/org/pantheon/appcenter`. This can be tested e.g: with d-feet.


danrabbit:
Fixes #35

danrabbit:
Since #180 was merged, this should probably change to `io.elementary` instead of `org.pantheon`

donadigo:
@danrabbit done.

src/Application.vala
```diff
@@ -158,6 +182,50 @@ public class AppCenter.App : Granite.Application {
             });
         }
     }
+
+    private void on_operation_finished (AppCenterCore.Package package, AppCenterCore.Package.State operation, Error? error) {
+        switch (operation) {
+            case AppCenterCore.Package.State.INSTALLING:
```
davidmhewitt: It would be good if we could have an un-installed version of this so that when stuff has been un-installed in the background, you get a notification to say it has succeeded (or not).

src/Services/DBusServer.vala
```diff
@@ -0,0 +1,115 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Adam Bieńkowski <donadigos159@gmail.com>
+ */
+
+[DBus (name="io.elementary.appcenter")]
+public class DBusServer : Object {
+    private static DBusServer? instance;
+    public static unowned DBusServer get_default () {
```
davidmhewitt: Maybe construct your singleton using [GLib.Once](http://stackoverflow.com/questions/26832371/how-to-create-a-thread-safe-singleton-in-vala) for neatness?

src/Application.vala
davidmhewitt: It would be good to have an un-install version of this so that we can get a notification when things have been un-installed (or failed) in the background.

donadigo: I will leave it as is for now, I want to do it in another PR.

donadigo:
Please don't approve / merge this PR yet.

davidmhewitt:
@donadigo I think you were maybe closer to being right after the first change with `package.update_state()`. With this current version, if I uninstall a package using the DBus daemon, then install it again from synaptic or the command line, then try and uninstall it from the DBus daemon again, nothing happens.

External changes to the package system aren't picked up by packagekit, so I think it's probably important to check the state of a package before you run an action on it.

donadigo:
@davidmhewitt resolved.

PR #184, opened 2017-03-27
Talk to Houston async and update banner in a thread

### LINKED ISSUE #154: AppCenter taking a long time to start
My suspicion is that we're waiting for Houston API to return results before showing the main window. We should be showing the main window immediately and then updating the homepage after we get results

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43376550-appcenter-taking-a-long-time-to-start?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #154

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -108,9 +108,11 @@ public class AppCenter.App : Granite.Application {
         }
 
         if (main_window == null) {
-            client.update_cache.begin (true);
-
             main_window = new MainWindow (this);
+            // Start updating the cache 5 seconds after starting the main window
+            // If we do it instantly, we slow down the loading of the banner on the
+            // homepage considerably
+            schedule_cache_update (false, 5);
             main_window.destroy.connect (() => {
                 main_window = null;
             });
@@ -139,7 +141,7 @@ public class AppCenter.App : Granite.Application {
     }
 
     private uint cache_update_timeout_id = 0;
-    private void schedule_cache_update (bool cancel = false) {
+    private void schedule_cache_update (bool cancel = false, int time = SECONDS_AFTER_NETWORK_UP) {
         var client = AppCenterCore.Client.get_default ();
 
         if (cache_update_timeout_id > 0) {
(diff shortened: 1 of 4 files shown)


src/Application.vala
```diff
@@ -108,9 +108,11 @@ public class AppCenter.App : Granite.Application {
         }
 
         if (main_window == null) {
-            client.update_cache.begin (true);
-
             main_window = new MainWindow (this);
+            // Start updating the cache 5 seconds after starting the main window
+            // If we do it instantly, we slow down the loading of the banner on the
+            // homepage considerably
+            schedule_cache_update (false, 5);
```
danrabbit: Is there a reason for it to be 5s? I'm not a huge fan of magic numbers

davidmhewitt: Not really, it just prevents the banner thread and the main thread interfering when talking to PackageKit. I guess there's a couple of options:

One is to move that code out to Homepage.vala and trigger it after the banner has finished loading, but I guess you'd expect that code to be in Application.vala instead of hidden in one of the Widgets. The other is to just make the number less "magic" and give it a const.

davidmhewitt: And when I say interfering, I mean slowing each other down. Nothing crashes, just the less important check for update actions tend to get put at the front of the queue, and the banner takes a while to change as a result.

davidmhewitt:
@danrabbit Hopefully this solution is better. Instead of waiting for an arbitrary 5 seconds, the cache update is now trigger from a signal fired after the homepage has loaded.

src/Core/Houston.vala
```diff
@@ -42,29 +42,30 @@ public class AppCenterCore.Houston : Object {
...
+            } catch (Error e) {
+                stderr.printf ("Houston: %s\n", e.message);
             }
-        } catch (Error e) {
-            stderr.printf ("Houston: %s\n", e.message);
-        }
+            Idle.add (get_newest.callback);
+        });
 
+        yield;
```
eyelash: Why do you yield here right before return? I can't see any other yield statements in this method and only one yield statement right before return seems to be rather missing the point.

davidmhewitt:
@eyelash Look at the 2nd sample in the Vala async samples:
https://wiki.gnome.org/Projects/Vala/AsyncSamples

The yield is waiting for the callback from the async `queue_message` method.

eyelash:
@davidmhewitt I was just confused for a moment and realized what's going on seconds after I wrote my comment. Sorry for the noise.

PR #187, opened 2017-03-29
Add name to BACKERS.md

alexgleason:
At the risk of being vain here...

My name got lost in this. Maybe because I didn't create a user account on Indiegogo, I just put in my card info. If you want to check me on it, the email associated is alex@alexgleason.me. I can provide more info if you email me. Maybe it's worth seeing if there's a way to include others who supported the project this way.

Thanks again for all your hard work on this!


danrabbit:
Hey Alex, I haven't added all names to the backers.md just yet. Currently it only contains names for people who didn't sign up for any other rewards. That's why it isn't proposed for merging yet :)

PR #189, opened 2017-03-30
CategoryFlowBox: Refactor category making

Philip-Scott:
Let's refactor all the things!


tintou:
Just a tiny comment, it's better to pass `null` than "" when it's possible. ("" is a newly allocated string)

PR #192, opened 2017-03-31
Animate banner

### LINKED ISSUE #185: Banner transition is not animated [$25]
When the banner changes, this is not animated. Very visible when testing #184

---
[There is a bounty on this issue on Bountysource](https://www.bountysource.com/issues/43504790-banner-transition-is-not-animated)

donadigo:
Fixes #185.

Make the background transition from one color to another and move the contents from right to left.

* The entire animation duration can be controlled by changing `transition_duration` on the `stack` widget.
* The movement can be changed with `transition_type` also on the `stack` widget.
* The background is animated with CSS (`transition`) and also can be changed accordingly.

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -45,8 +45,9 @@ namespace AppCenter {
             newest_banner.margin = 12;
 
             newest_banner.clicked.connect (() => {
-                if (newest_banner.current_package != null) {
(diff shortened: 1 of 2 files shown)


danrabbit:
This is dope dude! Can we maybe slow down the transition a bit here? I think like 500-600ms looks a little nicer in this context

donadigo:
@danrabbit done.

PR #193, opened 2017-04-01
Refresh package cache when last-apt-update file changes

### LINKED ISSUE #68: AppCenter icons show update counts on Slingshot and Plank, even after updating via apt on Terminal
Whenever AppCenter shows that there are updates, I notice three things happen:

* The icon in Slingshot menu shows the update count.
* The icon for AppCenter temporarily stickies on the dock with the update count shown.
* There is a notification which can be dismissed at will.

I usually prefer to do my updates via the terminal so I can see what is changing. However, upon actually updating my packages via apt on Terminal, I've noticed that the AppCenter icon still sits in my dock (with the exact same update count as before), and so does the icon in Slingshot (same behaviour -- with the update count).

The intended behaviour would be that once the updates are done, the icon would leave the dock, and the counter on the Slingshot icon would go back to nil.

I have tried updating my package list again using apt from the terminal (which tells me all my packages are up to date), but the icons still show updates.

So far, I know a restart works fine.
Also, if you update via the AppCenter, it works fine.

I will report back on the behaviour once I see more updates on AppCenter and try updating it via apt on Terminal.

Launchpad Details: [#LP1627498](https://bugs.launchpad.net/bugs/1627498) Ruppal Singh - 2016-09-25 13:53:35 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123634-appcenter-icons-show-update-counts-on-slingshot-and-plank-even-after-updating-via-apt-on-terminal?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #68 

This branch depends on the following merge into the deb-packaging repo: https://code.launchpad.net/~davidmhewitt/appcenter/dpkg-hook/+merge/321634

I've added a hook into the post-invoke scripts for DPkg that modifies the /var/run/last-apt-updates file when a package operation has taken place. AppCenter watches for this file modification and updates its cache when a change has occured.

This catches all the cases where the package database is modified outside of AppCenter and should allow us to keep better track of what's installed and what isn't in cases where things are modified outside of AppCenter.


donadigo:
Hm, I feel like there's another way to do this, I think it is definitely worth to test [Pk.Control](https://valadoc.org/packagekit-glib2/Pk.Control.html) class. It looks like it provides some useful signals to control what's happening in PackageKit backend (transaction list changed or some packages updated).

If this doesn't work, then your solution is also suitable.

davidmhewitt:
@donadigo cool, thanks. At first glance, they look like very packagekit specific signals, so they may not fire when external applications make changes. But I'll give them a test tomorrow for sure, and let you know.

Would be much more elegant if they do work.

Conan-Kudo:
It seems weird to see Apt specific code being written into AppCenter... Isn't the point of this to leverage the PK control mechanisms? For most backends, PK is at least peripherally aware of when something else does something it would usually do.

davidmhewitt:
You're both right, this is the wrong way of doing this. There are signals in the Pk.Control class that provide this functionality. It's just a case of filtering out the ones that were fired by AppCenter itself and the ones that were fired by external programs. I'll close this merge.

PR #195, opened 2017-04-01
Use suggested price key

### LINKED ISSUE #144: Use appdata key for suggested price [$25]
Create a custom appdata key to allow developers to choose a suggested price

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43250331-use-appdata-key-for-suggested-price?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #144 

I have tested this with local metadata and it works, but none of the apphub metadata currently implements this key, so it will need testing with one of those once available.


src/Widgets/HumbleButton.vala
```diff
@@ -123,7 +124,7 @@ public class AppCenter.Widgets.HumbleButton : Gtk.Grid {
         });
 
         custom_amount.value_changed.connect (() => {
-            amount = (int) custom_amount.value;
+            amount = (int)(custom_amount.value * 100);
```
tintou: change int to uint

davidmhewitt:
@tintou Good spot, thanks. Fixed.

Philip-Scott:
From a convo we had at Denver, I think we agreed that we were just not going to do cents. But in case we are, make sure you put $1 as the lower bound since anything lower would just end up eaten in Fees

davidmhewitt:
@Philip-Scott Done!

danrabbit:
@davidmhewitt we don't want to do that because users need to be able to select 0. But what we don't want is to process a charge for 30c since it costs 33c to process a charge :p

davidmhewitt:
@danrabbit ah crap yeah, I forgot that was how you got a free app. So, is the appcenter key going to be in dollars or cents and then rounded to whole dollars?

danrabbit:
I think maybe @btkostner suggested the price be in cents because stripe API deals with it in cents? But yeah I'm not sure it makes sense for developers to be able to set non-integer prices

davidmhewitt:
@danrabbit Ok, I'll go on the assumption that it will be in whole dollars in the metadata for now.

btkostner:
Ya, stripe API deals all in cents. That's why I suggested it.

src/Widgets/AbstractAppContainer.vala
```diff
@@ -80,6 +82,19 @@ namespace AppCenter {
...
+                if (this.package == null || this.package.component == null || updates_view) {
+                    return DEFAULT_PRICE_DOLLARS;
+                } else if (suggested_amount_ == null) {
+                    suggested_amount_ = this.package.component.get_custom_value ("x-appcenter-suggested-price");
+                }
+
+                return suggested_amount_ != null ? suggested_amount_ : DEFAULT_PRICE_DOLLARS;
+            }
+        }
+
```
Philip-Scott: I feel like this code should be moved to `src/Widgets/HumbleButton.vala`. The default should be 0 if there is no Stripe key found. If there is one, then try to load the suggested price, and set it to 1 if non was set

davidmhewitt: Thanks @Philip-Scott , I agree about setting the price to 0 if no stripe key is found, I'll do that later. However, I think we should keep this code out of HumbleButton.vala . If we do that, it will depend on a Package object, while at the moment it's a generic class that just expects an amount. I think it's probably better to keep it generic, so it can be used in other cases if needed. We'll see what other people think.

Philip-Scott: you could pass it as a construction parameter to the HumbleButton then instead of forcing the amount at the end there :)

davidmhewitt: @Philip-Scott I've taken a look at this, and it seems as though the package metadata is not available at the point we construct the button, so the code makes sense as it is at the moment. The button is already set to free when no stripe key is available by the `action_button.can_purchase = payments_enabled;` line in the same place as setting the amount.

davidmhewitt:
@danrabbit Fixed!

src/Widgets/AbstractAppContainer.vala
```diff
@@ -73,7 +73,18 @@ namespace AppCenter {
...
                 }
 
-                return this.package.get_payments_key() != null;
+                return this.package.get_payments_key () != null;
+            }
+        }
+
+        private string suggested_amount_;
+        public string suggested_amount {
+            get {
```
donadigo: I don't think it is ever needed to check if package is `null`. If it could, then the package would be nullable. Also, all packages must have their component, since they can't be constructed without it, so this check is also not needed here. The Package class also takes care of handling when there is no such key, so you can also remove returning `DEFAULT_PRICE_DOLLARS`. What I would like to see here is just:

`public string suggested_amount {
   get {
             return package.get_suggested_amount ();
   }
}`

src/Widgets/AbstractAppContainer.vala
```diff
@@ -73,7 +73,18 @@ namespace AppCenter {
                     return false;
                 }
 
-                return this.package.get_payments_key() != null;
+                return this.package.get_payments_key () != null;
+            }
+        }
+
```
donadigo: Since the `suggested_amount` is cached within the Package itself, there is no need to have another private one that's basically doing the same thing.

davidmhewitt:
@donadigo Just seemed simpler to get rid of the suggested_amount property altogether since it was only used once. Let me know what you think.

PR #198, opened 2017-04-03
explicitly show the infobar label

### LINKED ISSUE #196: Label is hidden in restart infobar
well that's awkward
screenshot from 2017-04-02 13 01 01

danrabbit:
Fixes #196

It seems like the `no_show_all` is required so explicitly show the label to fix the issue


donadigo:
LGTM.

Issue #5, opened 2017-03-16
Clicking on Steam does not install the app


elementaryBot:
I searched for Steam in tke App Center, and it found it (although without the Steam icon). I clicked the blue install button, and it switched to a progress bar for a bit, but then the install button reappeared.

The app was not installed.

There were no other notifications or error messages.

Thanks.

Launchpad Details: [#LP1622229](https://bugs.launchpad.net/bugs/1622229) Anthony Shively - 2016-09-10 18:19:01 +0000

elementaryBot:
It affected me installing Firefox

Launchpad Details: [#LPC](https://bugs.launchpad.net/appcenter/+bug/1622229/comments/1) Faizul Anwar - 2016-09-10 19:34:58 +0000

elementaryBot:
Can confirm same behavior as shown in the video:: https://plus.google.com/111572794920772736742/posts/bG9H7wpY1Xr

Launchpad Details: [#LPC](https://bugs.launchpad.net/appcenter/+bug/1622229/comments/2) Gabriel_P - 2016-09-11 18:53:42 +0000

elementaryBot:
https://www.bountysource.com/issues/37793128-clicking-on-steam-does-not-install-the-app

Launchpad Details: [#LPC](https://bugs.launchpad.net/appcenter/+bug/1622229/comments/3) Maxim Taranov - 2016-12-18 04:23:29 +0000

davidmhewitt:
Resolved in #200

PR #200, opened 2017-04-03
Fix resolving packages for other architectures

donadigo:
Adds a check if the package array returned by PackageKit is empty, resolve packages again but this time using `NOT_ARCH` filter to find the same package for other architecture. I left a comment in the code briefly explaining why the change to avoid any confusion when dealing with it in the future. This should solve installation of e.g: Steam


danrabbit:
fixes #5

PR #208, opened 2017-04-06
Scroll banner through 5 newest apps

### LINKED ISSUE #164: Cycle banner through newest 5 apps [$25]
Now that I know we have at least 2 apps I can see that it stays on the newest one forever. We should instead cycle through a small set like the newest 5

---
There is a bounty on this issue in [Bountysource](https://www.bountysource.com/issues/43413149-cycle-banner-through-newest-5-apps)

davidmhewitt:
One thing to check is the licensing on `Switcher.vala`. I took the majority of that code from Slingshot. Do we need to change the license?

Fixes #164


danrabbit:
Licensing is good. It's 3 (or later) and so are we :)

The only thing I'm wondering about is how there's a little bit of jumpiness when the switcher first comes into existence. My first thoughts are either to put it in a gtk.revealer so that it's at least smoothly animated in or what if we packed the switcher inside the banner so that we don't have to move the other items at all?

davidmhewitt:
@danrabbit I did experiment with having the switcher inside the banner and it started causing a lot of issues around not being able to click it because the banner is actually a button, and then moving away from using a button caused a whole manner of style issues. So I went the easy route :wink: 

So I think I'd rather use a revealer unless you'd rather have me make a better effort at integrating it into the banner?

danrabbit:
@davidmhewitt naw revealer sounds good to me :)

src/Views/Homepage.vala
```diff
@@ -40,7 +43,15 @@ namespace AppCenter {
         construct {
             var houston = AppCenterCore.Houston.get_default ();
 
-            newest_banner = new Widgets.Banner ();
+            var switcher = new Widgets.Switcher ();
+            switcher.halign = Gtk.Align.CENTER;
+
+            switcher_revealer = new Gtk.Revealer ();
+            switcher_revealer.set_transition_type (Gtk.RevealerTransitionType.SLIDE_DOWN);
+            switcher_revealer.set_transition_duration (Widgets.Banner.TRANSITION_DURATION_MILLISECONDS);
```
danrabbit: Nice touch :1st_place_medal:

PR #212, opened 2017-04-07
AppInfoView: adjust screenshots

danrabbit:
* Reduce the height request to 500
* Set background color on whole screenshot stack
* Remove extra margin


Philip-Scott:
You should remove those white spaces on your code ;) Besides that, code is good but i honestly disagree with adding the background color on the image.

looking good > keeping the grid

PR #217, opened 2017-04-07
Update PackageKit cache when external package changes happen

### LINKED ISSUE #68: AppCenter icons show update counts on Slingshot and Plank, even after updating via apt on Terminal

davidmhewitt:
Fixes #68.

Add a handler for the updates_changed signal in PackageKit. This is fired when the package database for the system is modified in some way. Unfortunately, it's also fired when AppCenter performs any actions.

To avoid refreshing our cache all the time, we ensure that the updates_changed handler only does something when we aren't running any transactions. We assume any `updates_changed` signals coming within 2 seconds of performing actions within AppCenter are triggered by AppCenter itself and ignore them.

This already fixes the badge count being wrong after installing updates with apt, but will also help towards making sure we show the installed/update state of apps correctly even if they've been modified outside of AppCenter. This will need a little work to make sure we update the interface correctly, however.

### DIFF
--- a/src/Core/Client.vala
+++ b/src/Core/Client.vala
@@ -22,8 +22,17 @@ public class AppCenterCore.Client : Object {
 
     public bool connected { public get; private set; }
     public bool updating_cache { public get; private set; }
-    public uint task_count { public get; private set; default = 0; }
     public bool restart_required { public get; private set; default = false; }
+    private uint _task_count = 0;
+    public uint task_count {
+        public get {
+            return _task_count;
+        }
+        private set {
+            _task_count = value;
+            last_action = new DateTime.now_local ();
+        }
+    }
 
     public AppCenterCore.Package os_updates { public get; private set; }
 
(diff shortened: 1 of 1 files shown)


davidmhewitt:
@donadigo I've just tested the `has_tasks` method to see if we could detect whether it was AppCenter/PackageKit that triggered the `updates_changed` handler. Unfortunately, it was still triggered in a few cases due to AppCenter reducing the `task_count` down to 0 before the signal was triggered.

I have simplified the code somewhat though to make it clearer what's happening. Let me know what you think.

danrabbit:
I can confirm that it fixes #68

PR #219, opened 2017-04-08
Don't show spinner if packages are being updated

### LINKED ISSUE #218: No row headers while updating
This is weird, we shouldn't do this. The "searching for updates" header is not accurate in this context.

screenshot from 2017-04-07 20 39 30

davidmhewitt:
Fixes #218 .

The property that decides whether to show the headers or the "Searching for updates" spinner is bound to the `updating_cache` property of `Client.vala`.

This sometimes causes the spinner to come back when we're doing updates if PackageKit decides to refresh its cache as part of the update process. To prevent this, we return false in our local copy of `updating_cache` if we know that there are some updates happening in the list. This ensures that we always show the headers while updates are happening in the list, but the spinner functions as normal outside of those cases.

Also removed an `updates_on_top` variable that wasn't used anywhere while I was in that part of the code.

### DIFF
--- a/src/Views/AppListView.vala
+++ b/src/Views/AppListView.vala
@@ -85,7 +85,6 @@ namespace AppCenter.Views {
     /** AppList for the Updates View.  Sorts update_available first and shows headers.
       * Does not show Uninstall Button **/
     public class AppListUpdateView : AbstractAppList {
-        private bool updates_on_top;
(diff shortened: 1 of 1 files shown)


danrabbit:
I can confirm that this resolves the issue :)

PR #220, opened 2017-04-08
General fixes to the code

donadigo:
* Use native Vala retrieving of values supported by Gee: `.get (pkg_name)` -> `[pkg_name]`
* Make sure that finding packages for a specific component won't add `null` packages to the requested array
* Remove `async` from `search` method since it doesn't contain any asynchronous code


davidmhewitt:
Why do you think the values of `package_list` should be unowned? They're GObjects and should be memory managed by Vala.

donadigo:
You're right, Vala should take care of that, my bad.

donadigo:
@davidmhewitt done.

PR #245, opened 2017-04-24
Show alertview when no internet connection available

### LINKED ISSUE #4: Show an alertview when no Internet connection [$30]
AppCenter is useless offline so we should show an alertview with a link to network settings when there's no internet

One constraint to keep in mind: You still need to be able to access the "Updates" pane in order to see and remove installed apps

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123380-show-an-alertview-when-no-internet-connection?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

DDieter:
If no internet connection is available, the user is shown an alertview explaining that installing and updating apps is not possible.

Closes #4.


src/MainWindow.vala
```diff
@@ -151,11 +162,22 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         installed_view = new Views.InstalledView ();
         search_view = new Views.SearchView ();
 
+        network_alert_view = new Granite.Widgets.AlertView (_("Network is Not Available"),
+                                                            _("While disconnected from the network, installing and updating apps is impossible."),
```
btkostner: This line should read something more positive/constructive like `Connect to the internet to install apps.` instead of using the word "impossible."

src/MainWindow.vala
```diff
@@ -151,11 +162,22 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         installed_view = new Views.InstalledView ();
         search_view = new Views.SearchView ();
 
+        network_alert_view = new Granite.Widgets.AlertView (_("Network is Not Available"),
```
danrabbit: The "I" should be capitalized here for Title Case :)

DDieter: Ah yeah I'm not familiar with the rules of title case as we don't do that in the Dutch language. Fixed it in my last commit :)

src/MainWindow.vala
```diff
@@ -151,11 +162,22 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         installed_view = new Views.InstalledView ();
         search_view = new Views.SearchView ();
 
+        network_alert_view = new Granite.Widgets.AlertView (_("Network Is Not Available"),
+                                                            _("Connect to the internet to install or update apps."),
+                                                            "network-error");
+        network_alert_view.get_style_context ().remove_class (Gtk.STYLE_CLASS_VIEW);
+        network_alert_view.show_action (_("Network settings…"));
```
danrabbit: This should be title case "Network Settings…"

src/MainWindow.vala
```diff
@@ -95,6 +90,22 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         homepage.subview_entered.connect (view_opened);
         installed_view.subview_entered.connect (view_opened);
         search_view.subview_entered.connect (view_opened);
+
+        NetworkMonitor.get_default ().network_changed.connect (update_view);
+
+        network_alert_view.action_activated.connect (() => {
+            var list = new List<string> ();
```
danrabbit: This should use the settings url scheme instead as in `AppInfo.launch_default_for_uri ("settings://network", null);`

DDieter:
I was wondering if the current criteria for showing the alert view is strict enough? The alert isn't shown if 'NetworkMonitor' says the network is available, but I read in the documentation that this doesn't necessarily mean the user will be able to connect to the internet.

PR #248, opened 2017-04-27
UpdateManager

donadigo:
This branch implements new class in the Core of the AppCenter which is UpdateManager. This class is a helper object for the Client. For now it doesn't have much functionality in it, but it contains `get_updates` method which will prepare all the needed results for the Client to present them in the window. It is the Client's responsibility to process these as it is the owner of all `AppCenterCore.Package` packages and has all needed variables & methods to do so. In the future I expect to move more methods & functions to the `UpdateManager` class.

This branch also removes unneeded call to `get_updates` when ran in silent mode. Currently in trunk when the AppCenter starts in silent mode it is called *only* to know how many updates there are to present a notification. This is no longer the case and the updates are fetched once and presented immediately in the UI. It doesn't matter when they were fetched.

The `Client` now exposes `get_pk_client` static function for the `UpdateManager` which returns the main `Pk.Task` that AppCenter uses for all it's transactions.


danrabbit:
I'm going to block this until after release since we have a new release planned on Wednesday and we probably don't want to introduce major changes before then

donadigo:
Ready to review again.

src/Core/Client.vala
```diff
@@ -457,14 +404,13 @@ public class AppCenterCore.Client : Object {
     }
 
     private async void refresh_updates () {
-        updating_cache = true;
         task_count++;
-
+
```
jeremypw: Trailing spaces?

donadigo:
@jeremypw issues addressed.

PR #255, opened 2017-05-06
Only check if we need to reboot after doing some updates

### LINKED ISSUE #246: "Restart required" button shows before the update is finished [$50]
So the latest system updates started showing a "restart required" banner (screenshot below) and notification before the whole update has finished. The screenshot doesn't reflect the update in progress as it finished before I managed to take it.

I suspect if I click the button the restart will happen before the end of an update.

screenshot from 2017-04-27 11 04 30

---
There is a bounty on this issue [on Bountysource](https://www.bountysource.com/issues/44591619-restart-required-button-shows-before-the-update-is-finished)

davidmhewitt:
Fixes #246

### DIFF
--- a/src/Core/Client.vala
+++ b/src/Core/Client.vala
@@ -104,12 +104,12 @@ public class AppCenterCore.Client : Object {
     private void updates_changed_callback () {
         if (!has_tasks ()) {
             update_restart_state ();
-        }
 
-        var time_since_last_action = (new DateTime.now_local ()).difference (last_action) / GLib.TimeSpan.MILLISECOND;
-        if (!has_tasks () && time_since_last_action >= PACKAGEKIT_ACTIVITY_TIMEOUT_MS) {
-            info ("packages possibly changed by external program, refreshing cache");
-            update_cache.begin (true);
+            var time_since_last_action = (new DateTime.now_local ()).difference (last_action) / GLib.TimeSpan.MILLISECOND;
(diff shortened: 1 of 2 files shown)


davidmhewitt:
@donadigo @Philip-Scott 

Thanks for the reviews and testing, both of your issues should now be resolved.

src/Core/Client.vala
```diff
@@ -112,6 +102,10 @@ public class AppCenterCore.Client : Object {
     }
 
     private void updates_changed_callback () {
+        if (!has_tasks ()) {
+            update_restart_state ();
+        }
+
         var time_since_last_action = (new DateTime.now_local ()).difference (last_action) / GLib.TimeSpan.MILLISECOND;
         if (!has_tasks () && time_since_last_action >= PACKAGEKIT_ACTIVITY_TIMEOUT_MS) {
```
donadigo: I know I'm nitpicking now, but you could check `!has_tasks ()` only once in this function like this:

```
if (!has_tasks ()) {
    update_restart_state ();

    var time_since_last_action = (new DateTime.now_local ()).difference (last_action) / GLib.TimeSpan.MILLISECOND;
    if (time_since_last_action >= PACKAGEKIT_ACTIVITY_TIMEOUT_MS) {
       ...
   }
}
```

davidmhewitt:
@donadigo done! :)

PR #256, opened 2017-05-06
Synchronously register the DBus interface

donadigo:
This makes the DBus `io.elementary.appcenter` name appear when the server has fully initialized and is available for use by other programs. Previously the name (with Gtk interfaces) would appear almost instantly but the main server wouldn't be present and this created problems e.g: for Slingshot to connect to the main interface where all install / uninstall methods are exposed.


donadigo:
@tintou so previously, I tried to make it synchronous, but the Server needs to populate installed applications in the Core, and the method `get_installed_applications` is asynchronous. I can make a synchronous version in the Client just for this purpose but I don't really want to. Do you still want this to be synchronous? I am open to changing it.

tintou:
I would really prefer you to use either an async or a sync function but here you're blocking the main loop by waiting for the async function to end, just keep everything sync in this case

donadigo:
@tintou done.

PR #258, opened 2017-05-06
Add Recently Updated apps to homepage carousel

danrabbit:


src/Views/Homepage.vala
```diff
@@ -98,11 +99,62 @@ namespace AppCenter {
...
+                            }
+                        }
+                    }
+                    
+                    Idle.add (() => {
+                        foreach (var banner_package in packages_for_carousel) {
+                            recently_updated_carousel.add_package (banner_package);
+                        }
+                        recently_updated_revealer.reveal_child = true;
+                        main_window.homepage_loaded ();
```
davidmhewitt: We shouldn't be calling `main_window.homepage_loaded ()` twice. It's already called by the banner code. It causes a refresh of AppCenter's package cache which is slow and network/CPU intensive.

The reason why this exists is because the cache refresh was previously happening while we were trying to see if packages were installed. Because that was happening, we had to wait until that finished and it slowed the homepage load. So we deferred that until after the homepage was finished doing its stuff.

I imagine only calling it once in either the banner or the carousel will be fine. Otherwise, we need to come up with a more inventive solution to make sure it happens only once after both have loaded.

src/Views/Homepage.vala
```diff
@@ -98,11 +99,62 @@ namespace AppCenter {
...
+                        return false;
+                    });
+                    return null;
+                };
+                new Thread<void*> ("update-banner", run);
```
davidmhewitt: Can we rename this thread so that it doesn't match the other thread for the banner? Only really useful if we ever need to differentiate the two threads during debugging, but worth it nontheless.

danrabbit:
@davidmhewitt thanks for your review! Made the changes :)

PR #273, opened 2017-05-16
Add privacy info

### LINKED ISSUE #271: Link to Stripe's Privacy Policy
Since users are entering details into our app and it's not clear what privacy policy that data is subject to, we should find a way to link them to [Stripe's Privacy Policy](https://stripe.com/us/privacy).

danrabbit:
Fixes #271 

screenshot from 2017-05-16 11 09 18


src/Dialogs/StripeDialog.vala
```diff
@@ -139,10 +140,14 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
 
         get_content_area ().add (layouts);
 
-        var action_area = get_action_area ();
+        var privacy_policy_link = new Gtk.LinkButton.with_label ("https://stripe.com/us/privacy", _("Privacy Policy"));
```
cassidyjames: Should probably be the region-less https://stripe.com/privacy? I'm not 100% sure.

PR #276, opened 2017-05-17
Corrected semantically wrong word

bigfatbird:
Open means frei as in freedom if you translate it. But in this context you need the word Öffnen to launch an app.


danrabbit:
We use Weblate for translations: https://l10n.elementary.io/projects/appcenter

Please make translation suggestions there :)

PR #282, opened 2017-05-17
Provide a custom error message from stripe

### LINKED ISSUE #277: More detailed Stripe errors [$50]
We actually get more  helpful feedback than we're showing. We should provide distinct error message for things like a card declining vs a server error

---
There is a bounty on this issue in [Bountysource](https://www.bountysource.com/issues/45293475-more-detailed-stripe-errorrs)

### LINKED ISSUE #283: AppCenter Closes When Trying to Pay
The AppCenter has closed when trying to pay for both Notes-Up and Spice-Up.

I had the issue originally with Notes-Up since that was the app I tried paying for first and eventually I was able to pay. There was no issue with the download of Notes-Up. I have not been able to pay for Spice-Up so I just clicked pay later and it installed without issue.

tintou:
Fixes #277 

Should also fix #283

### DIFF
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -27,7 +27,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
                             + "&card[cvc]=%s&card[exp_month]=%s&card[exp_year]=%s"
                             + "&key=%s";
 
-    private const string INTERNAL_ERROR_MESSAGE = N_("An error occurred while processing the card. Please try again later. We apologize for any inconvenience caused.")
+    private const string INTERNAL_ERROR_MESSAGE = N_("An error occurred while processing the card. Please try again later. We apologize for any inconvenience caused.");
     private const string DEFAULT_ERROR_MESSAGE = N_("Please review your payment info and try again.");
 
     private Gtk.Grid card_layout;
@@ -395,12 +395,16 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
                 debug (e.message);
             }
 
-            if (error != null) {
-                show_error_view (error);
-            } else {
-                download_requested ();
-                destroy ();
-            }
+            Idle.add (() => {
+                if (error != null) {
+                    show_error_view (error);
+                } else {
+                    download_requested ();
+                    destroy ();
+                }
+
+                return GLib.Source.REMOVE;
+            });
 
             return null;
         });
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -461,71 +461,14 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
                             return _("The card has expired. Please try again with a different card.");
                         case "processing_error":
                             return _(INTERNAL_ERROR_MESSAGE);
-                        default:
-                            return _(DEFAULT_ERROR_MESSAGE);
-                    }
-                } else if (error_object.has_member ("decline_code")) {
-                    unowned string error_code = error_object.get_string_member ("decline_code");
-                    debug ("Stripe decline error code: %s", error_code);
-                    switch (error_code) {
-                        case "card_not_supported":
-                            return _("This card does not support this kind of transaction. Please try again with a different card.");
-                        case "currency_not_supported":
-                            return _("The currency is not supported by this card. Please try again with a different card.");
-                        case "duplicate_transaction":
-                            return _("The transaction has already been processed.");
-                        case "expired_card":
-                            return _("The card has expired. Please try again with a different card.");
-                        case "incorrect_zip":
-                            return _("The ZIP/Postal code is incorrect. Please try again using the correct ZIP/Postal code.");
-                        case "insufficient_funds":
-                            return _("You don't have enough funds. Please use an alternative payment method.");
-                        case "invalid_amount":
-                            return _("The amount is incorrect. Please try again using a valid amount.");
-                        case "incorrect_cvc":
-                        case "invalid_cvc":
-                            return _("The CVC number is incorrect. Please try again using the correct CVC.");
-                        case "invalid_expiry_year":
-                            return _("The expiration year is invalid. Please try again using the correct expiration date.");
-                        case "incorrect_number":
-                        case "invalid_number":
-                            return _("The card number is incorrect. Please try again using the correct card number.");
-                        case "incorrect_pin":
-                        case "invalid_pin":
-                            return _("The pin number is incorrect. Please try again using the correct pin.");
-                        case "pin_try_exceeded":
-                            return _("There has been too many pin attempts. Please try again with a different card.");
-                        case "call_issuer":
-                        case "do_not_honor":
-                        case "do_not_try_again":
-                        case "fraudulent":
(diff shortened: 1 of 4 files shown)


lewisgoddard:
I'd question these strings. "payment info number"? Why not just say card number?

tintou:
Because there was never a mention of "card" so I assumed it was on-purpose ¯\_(ツ)_/¯

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +430,42 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+        unowned string error_type = error_object.get_string_member ("type");
+        debug ("Stripe error type: %s", error_type);
+        switch (error_type) {
+            case "card_error":
+                unowned string error_code = error_object.get_string_member ("code");
+                debug ("Stripe error code: %s", error_code);
+                switch (error_code) {
+                    case "incorrect_number":
+                    case "invalid_number":
+                        return _("Please review your payment info number and try again.");
```
danrabbit: "The card number is incorrect. Please try again using the correct card number."

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +430,42 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+                    case "invalid_expiry_month":
+                    case "invalid_expiry_year":
+                        return _("Please review your payment info expiry and try again.");
```
danrabbit: "The expiration year is invalid. Please try again using the correct expiration date."

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +430,42 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+        debug ("Stripe error type: %s", error_type);
+        switch (error_type) {
+            case "card_error":
+                unowned string error_code = error_object.get_string_member ("code");
+                debug ("Stripe error code: %s", error_code);
+                switch (error_code) {
+                    case "incorrect_number":
+                    case "invalid_number":
+                        return _("Please review your payment info number and try again.");
+                    case "invalid_expiry_month":
```
danrabbit: "The expiration month is invalid. Please try again using the correct expiration date."

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +430,42 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+                    case "invalid_expiry_year":
+                        return _("Please review your payment info expiry and try again.");
+                    case "incorrect_cvc":
+                    case "invalid_cvc":
+                        return _("Please review your payment info CVC and try again.");
```
danrabbit: "The CVC number is incorrect. Please try again using the correct CVC."

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +430,42 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+                    case "expired_card":
+                        return _("The provided payment info has expired.");
```
danrabbit: "The card has expired. Please try again with a different card."

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +430,42 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+                    case "processing_error":
+                        return _("An payment processing error occurred, please retry later. We apologize for any inconvenience caused.");
```
danrabbit: "An error occurred while processing the card. Please try again later. We apologize for any inconvenience caused."

danrabbit:
According to @btkostner some of these errors might not ever be reported to AppCenter for security purposes. We should find out which ones and make sure we're in line with what we expect Houston to send to us

tintou:
AppCenter is asking Stripe directly here, so we are sure to get the Stripe errors here. The Errors from Houston are considered as "Processing system error"

tintou:
I also fixed a possible crash here because there were Gtk calls in a different thread which is forbidden by Gtk

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +437,43 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
 
         return data.str;
     }
+
+    private string get_stripe_error (Json.Object error_object) {
```
donadigo: You don't check if there's `type` member in the `error_object` so, this could lead to some fatals.

src/Dialogs/StripeDialog.vala
```diff
@@ -427,5 +437,43 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
+        unowned string error_type = error_object.get_string_member ("type");
+        debug ("Stripe error type: %s", error_type);
+        switch (error_type) {
+            case "card_error":
+                unowned string error_code = error_object.get_string_member ("code");
```
donadigo: The same as with the type member. Check if `error_object` actually has the `code` member.

src/Dialogs/StripeDialog.vala
donadigo: Can we make this static? Doesn't look like it's accessing any members from the instance. Also, this function should return an `unowned string` since `_` function you're using does return an `unowned`: https://valadoc.org/glib-2.0/GLib._.html

src/Dialogs/StripeDialog.vala
```diff
@@ -360,41 +367,44 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
...
-
-                    if (root_object.has_member ("errors")) {
-                        error = true;
+                if (root_object != null && root_object.has_member ("id")) {
+                    var token_id = root_object.get_string_member ("id");
+                    string? houston_data = post_to_houston (stripe_key, app_id, token_id, email_entry.text, (amount * 100).to_string ());
+                    if (houston_data != null) {
+                        debug ("Houston data:%s", houston_data);
+                        parser.load_from_data (houston_data);
+                        root_object = parser.get_root ().get_object ();
```
donadigo: Both `get_root` and `get_object` could return `null`. Since this is already in trunk, I won't force you to check for that, but it would be nice to see this checked.

src/Dialogs/StripeDialog.vala
```diff
@@ -360,41 +367,44 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
 
             var data = get_stripe_data (stripe_key, email_entry.text, (amount * 100).to_string (), card_number_entry.text, card_expiration_entry.text[0:2], year, card_cvc_entry.text);
             debug ("Stripe data:%s", data);
-            var error = false;
+            string error = null;
```
donadigo: This should be `string?` not `string`, the `error` is nullable.

btkostner:
@tintou check the houston channel. We were debugging a crash related to threading. Is it the same issue?

src/Dialogs/StripeDialog.vala
tintou: This part of the code needs some cleanup, but I would prefer it being done in another PR

danrabbit:
Everything works as expected for me! Just waiting for Adam's review :)

PR #298, opened 2017-05-20
Allow putting a slash for the date

### LINKED ISSUE #250: The date placeholder in the StripeDialog is misleading
Hey guys,

So I just tried to buy something in Appcenter (and I succeeded). However, the placeholder in the expiry [date field](https://github.com/elementary/appcenter/blob/51480c8013ec8b53ee7ec957213cf0813a14dbc2/src/Dialogs/StripeDialog.vala#L93) is a bit misleading.

The reason for this is that it leads you to believe the format is "xx/yy", when the actual accepted format is "xxyy".

tintou:
Fixes #250


danrabbit:
LGTM, but maybe @Philip-Scott can double check since he's the original author on this dialog :)

UX Approve :+1:

PR #320, opened 2017-05-25
Create heading for non-elementary apps in Search and Category view

### LINKED ISSUE #319: Show "made for elementary" apps first in categories [$50]
Aside from the "Recently Updated" section of the homepage, there isn't really a way to find apps made for elementary right now. In Categories those apps are drowned out by the Ubuntu repository apps. 

For now I think categories should have a "Designed for elementary" and "Other Apps" header.

---
There is a bounty on this issue in [Bountysource](https://www.bountysource.com/issues/45521747-show-made-for-elementary-apps-first-in-categories)

davidmhewitt:
Fixes #319. Currently looks like the following with "Non-Curated Apps" as the heading above non-elementary apps.

screenshot from 2017-05-25 17 36 52

There's quite a lot of discussion around the actual wording of that heading, so I'm open to feedback.


cassidyjames:
I'm +1 for `Non-Curated`, at least for now. I think that most-clearly explains the implications to the user: These apps haven't gone through the review process which means they might well be garbage, they might be fine; we can't know

Conan-Kudo:
I'm not a fan of the `Non-Curated` wording. For one, curation could be happening at different levels, so you can't make that particular assumption.

cassidyjames:
@Conan-Kudo as far as the user of elementary OS is concerned, these apps have not been curated by the team behind the elementary OS AppCenter. Non-Curated is as precise as we could come up with, but we're open to further suggestions that clearly communicate the difference in apps.

PR #322, opened 2017-05-26
Order installed packages by date

### LINKED ISSUE #36: Sort installed tab by date installed [$20]
Especially since we're currently not sorting out libraries, I think it would be more helpful to sort by date installed than alphabetically.

Launchpad Details: [#LP1512912](https://bugs.launchpad.net/bugs/1512912) Daniel Fore - 2015-11-04 00:39:57 +0000

DDieter:
Searches for the latest installation/update timestamp in the transaction logs and uses this information to order the packages in the update view (newest first). Still uses the package name when this information is not present (pre-installed and never updated packages).

Closes #36.


danrabbit:
afaict, this does what it says it does :) Looks like @donadigo or @Philip-Scott are recommended here for code review

src/Core/Client.vala
```diff
@@ -310,23 +313,40 @@ public class AppCenterCore.Client : Object {
...
+        populate_timestamp (package, pk_package, transactions);
         package.update_state ();        
     }
 
+    private static void populate_timestamp (AppCenterCore.Package package, Pk.Package pk_package, Gee.Collection<Pk.TransactionPast> transactions) {
+        foreach (var pk_transaction in transactions) {
+            var lines = pk_transaction.get_data ().split ("\n");
+
+            for (var i = 0; i < lines.length; i++) {
+                if (lines[i].has_prefix ("installing") && lines[i].contains(pk_package.get_name ())) {
```
jeremypw: Suggest assign pk_package.get_name() outside loop to avoid repeated fetch.

src/Views/AppListView.vala
```diff
@@ -231,6 +231,21 @@ namespace AppCenter.Views {
                 return a_is_driver ? - 1 : 1;
             }
 
+            string? a_timestamp = row1.get_timestamp ();
+            string? b_timestamp = row2.get_timestamp ();
+
```
jeremypw: Suggest string a_timestamp = row1.get_timestampe () ?? "" and same for b_timestamp, in order to avoid possibly comparing null with null.

jeremypw:
Not a review, but added a couple of comments to the diff.

DDieter:
I've made some changes as you suggested. The only concern I have with this branch is it may have some performance issues when there are many transactions to search through. So if anyone has any suggestions let me know.

src/Core/Client.vala
```diff
@@ -277,6 +277,7 @@ public class AppCenterCore.Client : Object {
                 var pk_package = package.find_package ();
                 if (pk_package != null && pk_package.get_info () == Pk.Info.INSTALLED) {
                     package.installed_packages.add (pk_package);
+                    populate_timestamp (package, pk_package, get_install_transactions_sync ());
```
donadigo: Note that calling `get_install_transactions_sync` could be a bottleneck on some systems. Calling it in different places is unnecessary because really it will return the same results as the previous calls and will make AppCenter's backend load time slower than it already is. You should get those transactions only once at the beggining of Client initialization and then use them throughout the Client class.

donadigo:
Before I can approve, I made one comment about performance issues that you're facing and possibly how to get rid of them. These should be resolved.

DDieter:
There's still an issue where updating the package won't reorder it to the top until AppCenter is restarted. Closing this PR for now until I can fix this.

PR #325, opened 2017-05-27
Load more screenshots

### LINKED ISSUE #228: Show more than one screenshot [$25]
We're only loading one screenshot right now. We gotta load moar

---
There is a bounty for this issue in [Bountysource](https://www.bountysource.com/issues/43939045-show-more-than-one-screenshot)

DDieter:
Loads all the available screenshots when viewing package info. The screenshots are added to a stack and navigation is made possible with a switcher.

Screenshots marked as `DEFAULT` will be shown before other screenshots and the image with source quality is used. When one or more screenshot cannot be loaded it will place the error label in the stack instead of the image.

Closes #228.

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -287,32 +287,47 @@ namespace AppCenter.Views {
             new Thread<void*> ("content-loading", () => {
                 app_version.label = package.get_version ();
 
-                string url = null;
-                uint max_size = 0U;
-
                 if (screenshots.length == 0) {
                     return null;
                 }
 
+                List<string> urls = new List<string> ();
+
                 screenshots.foreach ((screenshot) => {
                     screenshot.get_images ().foreach ((image) => {
-                        if (max_size < image.get_width ()) {
-                            url = image.get_url ();
-                            max_size = image.get_width ();
+                        if (image.get_kind () == AppStream.ImageKind.SOURCE) {
+                            if (screenshot.get_kind () == AppStream.ScreenshotKind.DEFAULT) {
+                                urls.prepend (image.get_url ());
+                            } else {
+                                urls.append (image.get_url ());
+                            }
+
+                            return;
                         }
                     });
                 });
 
-                if (url != null) {
-                    set_screenshot (url);
+                foreach (var url in urls) {
+                    load_screenshot (url);
                 }
+
+                Idle.add (() => {
+                    if (app_screenshots.get_children ().length () > 0) {
+                        screenshot_stack.visible_child = app_screenshots;
+                        screenshot_switcher.update_selected ();
+                    } else {
+                        screenshot_stack.visible_child = app_screenshot_not_found;
+                    }
+
+                    return GLib.Source.REMOVE;
+                });
 
                 return null;
             });
         }
 
         // We need to first download the screenshot locally so that it doesn't freeze the interface.
-        private void set_screenshot (string url) {
+        private void load_screenshot (string url) {
             var ret = GLib.DirUtils.create_with_parents (GLib.Environment.get_tmp_dir () + Path.DIR_SEPARATOR_S + ".appcenter", 0755);
             if (ret == -1) {
                 critical ("Error creating the temporary folder: GFileError #%d", GLib.FileUtils.error_from_errno (GLib.errno));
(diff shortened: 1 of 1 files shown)


danrabbit:
I can confirm that this seems to work as expected. For testing purposes, Hourglass has multiple screenshots.

:+1: on UX

src/Views/AppInfoView.vala
```diff
@@ -281,32 +281,48 @@ namespace AppCenter.Views {
...
                 app_version.label = package.get_version ();
 
-                string url = null;
-                uint max_size = 0U;
-
                 if (screenshots.length == 0) {
                     return null;
                 }
 
+                Gee.Deque<Gtk.Widget> screenshot_widgets = new Gee.LinkedList<Gtk.Widget> ();
```
Philip-Scott: Why did you declare `screenshot_widgets` as a Gee.Deque if you are creating a new Gee.LinkedList?

src/Views/AppInfoView.vala
```diff
@@ -281,32 +281,48 @@ namespace AppCenter.Views {
...
+                    if (screenshot.get_kind () == AppStream.ScreenshotKind.DEFAULT) {
+                        screenshot_widgets.offer_head (screenshot_widget);
+                    } else {
+                        screenshot_widgets.offer_tail (screenshot_widget);
+                    }
                 });
 
-                if (url != null) {
-                    set_screenshot (url);
+                foreach (var widget in screenshot_widgets) {
```
Philip-Scott: This line, and the following modifying the already added widgets should be made on the main thread, if not you can get crashes.

You can do this using the `Idle.add (() => {});` line the one you removed on the set_screenshot function

src/Views/AppInfoView.vala
```diff
@@ -281,32 +281,48 @@ namespace AppCenter.Views {
...
                 screenshots.foreach ((screenshot) => {
-                    screenshot.get_images ().foreach ((image) => {
-                        if (max_size < image.get_width ()) {
-                            url = image.get_url ();
-                            max_size = image.get_width ();
+                    Gtk.Widget screenshot_widget = null;
+                    var index = 0;
+                    var images = screenshot.get_images ();
+
+                    while (screenshot_widget == null && index < images.length) {
```
Philip-Scott: foreach's are easier to understand :) You can add a `break` if the `widget != null`, and you also avoid having to declare the index

src/Views/AppInfoView.vala
Philip-Scott: This should be a list of strings with the URLs we will try to load on load_screenshot

src/Views/AppInfoView.vala
```diff
@@ -281,32 +281,48 @@ namespace AppCenter.Views {
...
+                screenshot_stack.visible_child = app_screenshots;
+                screenshot_switcher.update_selected ();
+
                 return null;
             });
         }
 
         // We need to first download the screenshot locally so that it doesn't freeze the interface.
-        private void set_screenshot (string url) {
+        private Gtk.Widget load_screenshot (string url) {
```
Philip-Scott: Don't return the widget here :)

src/Views/AppInfoView.vala
```diff
@@ -323,30 +339,39 @@ namespace AppCenter.Views {
...
+            try {
+                var image = new Gtk.Image ();
+                image.width_request = 800;
+                image.height_request = 500;
+                image.icon_name = "image-x-generic";
+                image.halign = Gtk.Align.CENTER;
+                image.pixbuf = new Gdk.Pixbuf.from_file_at_scale (fileimage.get_path (), 800, 600, true);
+
+                return image;
+            } catch (Error e) {
```
Philip-Scott: Creating and adding the image to the stach should be done here, but on an Idle block instead.

src/Views/AppInfoView.vala
```diff
@@ -323,30 +339,39 @@ namespace AppCenter.Views {
...
+                critical (e.message);
+                return screenshot_load_error_label ();
+            }
+        }
 
-                return GLib.Source.REMOVE;
-            });
+        private Gtk.Label screenshot_load_error_label () {
```
Philip-Scott: This should only run once if no screenshot was able to be loaded :) That way you don't end up with a stack of images and errors, or even yet, a stack full of errors. Remember to do it on an Idle block

Philip-Scott:
Looks good to me! Thanks for making those changes :)

PR #330, opened 2017-06-07
Add carousel for trending apps

danrabbit:
Server side de-duplication was just merged so this should be :+1:

### DIFF
--- a/src/Core/Houston.vala
+++ b/src/Core/Houston.vala
@@ -96,6 +96,34 @@ public class AppCenterCore.Houston : Object {
         return app_ids;
     }
 
+
+    public async string[] get_trending () {
+        var uri = HOUSTON_API_URL + "/newest/downloads";
+        string[] app_ids = {};
+
+        debug ("Requesting trending applications from %s", uri);
+
+        var message = new Soup.Message ("GET", uri);
+        session.queue_message (message, (sess, mess) => {
+            try {
+                var res = process_response ((string) mess.response_body.data);
+                if (res.has_member ("data")) {
+                    var data = res.get_array_member ("data");
+
+                    foreach (var id in data.get_elements ()) {
+                        app_ids += ((string) id.get_value ());
+                    }
+                }
+            } catch (Error e) {
+                stderr.printf ("Houston: %s\n", e.message);
+            }
+            Idle.add (get_trending.callback);
+        });
+
+        yield;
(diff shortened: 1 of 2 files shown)


src/Core/Houston.vala
```diff
@@ -98,6 +98,34 @@ public class AppCenterCore.Houston : Object {
...
+                var res = process_response ((string) mess.response_body.data);
+                if (res.has_member ("data")) {
+                    var data = res.get_array_member ("data");
+
+                    foreach (var id in data.get_elements ()) {
+                        app_ids += ((string) id.get_value ());
+                    }
+                }
+            } catch (Error e) {
+                stderr.printf ("Houston: %s\n", e.message);
```
Philip-Scott: You should use error, critical or warning instead of stderr. warning should be fine on this case since it won't kill the app :)

src/Views/Homepage.vala
```diff
@@ -153,6 +153,58 @@ namespace AppCenter {
...
+            trending_grid.margin_top = 12;
+            trending_grid.attach (trending_label, 0, 0, 1, 1);
+            trending_grid.attach (trending_carousel, 0, 1, 1, 1);
+
+            var trending_revealer = new Gtk.Revealer ();
+            trending_revealer.add (trending_grid );
+
+            houston.get_trending.begin ((obj, res) => {
+                var updated_ids = houston.get_updated.end (res);
+                List<AppCenterCore.Package> packages_for_carousel = new List<AppCenterCore.Package> ();
```
Philip-Scott: GLib list or GeeList? (Should use Gee if possible, and you can use var at the beginning

src/Views/Homepage.vala
```diff
@@ -153,6 +153,58 @@ namespace AppCenter {
...
+            trending_grid.margin = 2;
+            trending_grid.margin_top = 12;
+            trending_grid.attach (trending_label, 0, 0, 1, 1);
+            trending_grid.attach (trending_carousel, 0, 1, 1, 1);
+
+            var trending_revealer = new Gtk.Revealer ();
+            trending_revealer.add (trending_grid );
+
+            houston.get_trending.begin ((obj, res) => {
+                var updated_ids = houston.get_updated.end (res);
```
davidmhewitt: This should be:
`var trending_ids = houston.get_trending.end (res);`

PR #332, opened 2017-06-08
Banners are now saved locally

### LINKED ISSUE #210: Cache banners [$80]
That way you theoretically only see the placeholder once and then we update from the last cached set

---
There is a bounty on this issue in [Bountysource](https://www.bountysource.com/issues/43879325-cache-banners)

<bountysource-plugin>

---
There is a **[$80 open bounty](https://www.bountysource.com/issues/43879325-cache-banners?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tintou:
Fixes #210
Please not that the banners aren't loaded directly because it spends a real huge amount of time in `candidate_package.update_state ();` at startup

# What does this branch do
 - Houston cache is loaded every day
 - The cache is stored in a KeyFile
 - Add a `clear ()` function to `Carousel` and `Banner` containers in order to update them
 - Replaced some `parent.remove(child)` with `child.destroy()`, the destroy signal is watched by parents so it's safer
 - Houston fetching is now completely asynchronous

# How to test
Launch AppCenter, it launches normally
Disconnect to your network
Re-launch AppCenter, everything should be back again

### DIFF
--- a/src/Core/Houston.vala
+++ b/src/Core/Houston.vala
@@ -17,85 +17,154 @@
 */
 
 public class AppCenterCore.Houston : Object {
+    public signal void refreshed ();
+
     private const string HOUSTON_API_URL = "https://developer.elementary.io/api";
+    private const string NEWEST_CATEGORY = "Newest";
+    private const string UPDATED_CATEGORY = "Updated";
+    private const string APPS_KEY = "apps";
 
     private Soup.Session session;
+    private GLib.KeyFile cache_file = null;
 
     construct {
         session = new Soup.Session ();
     }
 
-    private Json.Object process_response (string res) throws Error {
-        var parser = new Json.Parser ();
-        parser.load_from_data (res, -1);
+    public string[] get_newest () {
+        if (cache_file == null) {
+            init_keyfile ();
+        }
 
-        var root = parser.get_root ().get_object ();
+        try {
+            return cache_file.get_string_list (NEWEST_CATEGORY, APPS_KEY);
+        } catch (Error e) {
+            critical (e.message);
+            return {};
+        }
+    }
 
-        if (root.has_member ("errors") && root.get_array_member ("errors").get_length () > 0) {
-            var err = root.get_array_member ("errors").get_object_element (0).get_string_member ("title");
+    public string[] get_updated () {
+        if (cache_file == null) {
+            init_keyfile ();
+        }
 
-            if (err != null) {
-                throw new Error (0, 0, err);
+        try {
+            return cache_file.get_string_list (UPDATED_CATEGORY, APPS_KEY);
+        } catch (Error e) {
+            critical (e.message);
+            return {};
+        }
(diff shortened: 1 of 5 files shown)


danrabbit:
I think a day is a little long, but I guess it makes sense if we're assuming that we're also updating metadata once a day. I think in the future I'd like to see this in the new update manager class so that we're updating this cache in the background and AppCenter can show an updated banner/carousel at launch time.

src/Core/Houston.vala
```diff
@@ -17,85 +17,154 @@
...
+        GLib.Timeout.add (timeout, () => {
+            refresh.begin ();
+            return GLib.Source.REMOVE;
         });
+    }
 
-        yield;
-        return app_ids;
+    private static string get_cache_file_path () {
+        return GLib.Path.build_filename (GLib.Environment.get_user_cache_dir (), "io.elementary.appcenter", "houston.conf");
```
donadigo: I'm not sure about the ".conf" extension, could this be renamed to something like ".cache" or ".data"?
Regarding the function: could this path constructing go into a `static_construct` to not construct the path everytime we need it?

src/Core/Houston.vala
```diff
@@ -17,85 +17,154 @@
...
 public class AppCenterCore.Houston : Object {
+    public signal void refreshed ();
+
     private const string HOUSTON_API_URL = "https://developer.elementary.io/api";
+    private const string NEWEST_CATEGORY = "Newest";
+    private const string UPDATED_CATEGORY = "Updated";
+    private const string APPS_KEY = "apps";
 
     private Soup.Session session;
+    private GLib.KeyFile cache_file = null;
```
donadigo: If the `cache_file` is nullable, it should contain `?` at the end of the type.
I guess the `KeyFile` could also be static here?

src/Core/Houston.vala
tintou: About the extension, the file is already in the .cache folder, I'm just using the commonly used extension for keyfiles, I could also remove the extension.

It could go into a static_construct function, but that would also mean that it would take more memory when the cache is only refreshed periodically. So here I've chosen the static function, but we can discuss it further if you wish

src/Core/Houston.vala
tintou: Yeah you're right with the nullable, I don't think that it would change anything to set it as static when the Houston class is already a singleton

donadigo: Ah, I haven't noticed that it is inded a singleton, well then, that's not required.

danrabbit:
Closing because conflicts and lack of movement. Please re-open if work here resumes

PR #339, opened 2017-06-21
AppInfoView.vala: code style and simplify

danrabbit:
Addresses a couple of concerns from Corentin on #338 

* Get rid of `set_up_package` override
* Move `parse_description` into construct
* cache `package.component` as `package_component`
* Remove whitespace below `var foo  = package_component.get_foo`


src/Views/AppInfoView.vala
```diff
@@ -219,13 +212,11 @@ namespace AppCenter.Views {
...
 
             open_button.get_style_context ().add_class ("h3");
-            reload_css ();
-        }
 
-        protected override void set_up_package (uint icon_size = 48) {
-            package_summary.label = package.get_summary ();
-            package_summary.ellipsize = Pango.EllipsizeMode.END;
-            base.set_up_package (icon_size);
+            base.set_up_package (128);
```
matzipan: I believe `base.` isn't needed here since there is not overriding function in the current class anymore.

src/Views/AppInfoView.vala
```diff
@@ -219,13 +212,11 @@ namespace AppCenter.Views {
...
+            parse_description (package.get_description ());
+
+            reload_css ();
```
matzipan: Before the change, `reload_css` was getting called before `set_up_package`. Does anything get influenced by the fact that it's getting called after it now?

PR #340, opened 2017-06-23
AppInfoView.vala: Add share button to footer

### LINKED ISSUE #190: Add "Copy Link" function
To easily share apps over the internets, copy the `appstream://` link to the clipboard

danrabbit:
Fixes #190 

Did this as "share" instead of link specifically in case we want to eventually make this a menu with more options like mail, twitter, etc

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -117,36 +117,47 @@ namespace AppCenter.Views {
             app_description.pixels_inside_wrap = 3;
             app_description.wrap_mode = Gtk.WrapMode.WORD_CHAR;
 
-            var links_grid = new Gtk.Grid ();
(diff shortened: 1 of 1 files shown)


cassidyjames:
I feel like this design is a regression since it requires tooltips to know for sure what each icon does. Was removing the labels intentional?
screenshot from 2017-06-23 13 05 55

cassidyjames:
Master for reference
screenshot from 2017-06-23 13 06 44

cassidyjames:
Can confirm that functionality-wise this seems good.

danrabbit:
Yeah I killed the labels because it makes each button a different width or if you make them homogeneous it makes the spacing look all fucky for the icons. How about like this instead?

screenshot from 2017-06-23 12 27 51

PR #345, opened 2017-06-24
AppInfoView: make share into a popover with email and twitter options

danrabbit:
screenshot from 2017-06-24 16 00 31

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -280,7 +280,7 @@ namespace AppCenter.Views {
                 share_popover.hide ();
             });
 
-            var canned_message = "Check out %s on AppCenter https://appcenter.elementary.io/%s".printf (package.get_name (), this.package.component.get_id ());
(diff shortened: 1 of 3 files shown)


src/Views/AppInfoView.vala
```diff
@@ -248,11 +272,32 @@ namespace AppCenter.Views {
...
-            share_button.clicked.connect (() => {
+            copy_link_button.clicked.connect (() => {
                 var clipboard = Gtk.Clipboard.get_for_display (get_display (), Gdk.SELECTION_CLIPBOARD);
                 clipboard.set_text ("https://appcenter.elementary.io/" + this.package.component.get_id (), -1);
 
                 toast.send_notification ();
+                share_popover.hide ();
+            });
+
+            var canned_message = "Check out %s on AppCenter https://appcenter.elementary.io/%s".printf (package.get_name (), this.package.component.get_id ());
```
Philip-Scott: This should probably be translatable

PR #346, opened 2017-06-25
Make share popover its own class, add more networks

danrabbit:
* Move the share popover out of AppInfoView into its own class
* Add Facebook and Google+ sharing


snwh:
Just a thought, but we can add even more sharing options in the form of reddit, pinterest and tumblr since they also have sharing links:

```
reddit_button.clicked.connect (() => {
    try {
        AppInfo.launch_default_for_uri ("http://www.reddit.com/submit?url=%s".printf (uri), null);
    } catch (Error e) {
        warning ("%s", e.message);
    }
    hide ();
});

pinterest_button.clicked.connect (() => {
    try {
        AppInfo.launch_default_for_uri ("http://pinterest.com/pin/create/button/?url=%s".printf (uri), null);
    } catch (Error e) {
        warning ("%s", e.message);
    }
    hide ();
});

tumblr_button.clicked.connect (() => {
    try {
        AppInfo.launch_default_for_uri ("https://www.tumblr.com/share/link?url=%s".printf (uri), null);
    } catch (Error e) {
        warning ("%s", e.message);
    }
    hide ();
});
```

harisvsulaiman:
I don't if that's the way contractor works, but it should be possible for apps to expose services as a target for the share widget, like how Android deals with it.
1.You install an app ( or pseudo app exposing new share targets)
2. share widget picks up the new service and displays it, along with defaults.

src/Widgets/SharePopover.vala
```diff
@@ -0,0 +1,165 @@
...
+        var reddit_button = new Gtk.Button.from_icon_name ("online-account-reddit", Gtk.IconSize.DND);
+        reddit_button.tooltip_text = _("reddit");
+        reddit_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+
+        var tumblr_button = new Gtk.Button.from_icon_name ("online-account-tumblr", Gtk.IconSize.DND);
+        tumblr_button.tooltip_text = _("Tumblr");
+        tumblr_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+
+        var pocket_button = new Gtk.Button.from_icon_name ("online-account-pocket", Gtk.IconSize.DND);
+        pocket_button.tooltip_text = _("Save to Pocket");
```
cassidyjames: I really don't think it makes any sense to add an AppCenter URL to a "read it later" service; it's not going to find any worthwhile content. :stuck_out_tongue:

snwh: @cassidyjames who are we to decide how people use Pocket. :wink: If anything it is a stop-gap for wishlisting apps

cassidyjames: It's literally an article saving service. Apps are not articles.

screenshot from 2017-06-29 20 07 56

screenshot from 2017-06-29 20 08 32

snwh: I don't think there's any harm in including Pocket, but we can remove it if you disagree.

PR #356, opened 2017-06-26
Create is_native in Package.vala and use in AppListView.vala

danrabbit:
Save some lines and kill the animals

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -24,6 +24,10 @@ public errordomain PackageLaunchError {
 }
 
 public class AppCenterCore.Package : Object {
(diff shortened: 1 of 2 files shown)


donadigo:
Instead of having that function in `AppListView`, you should make it a part of the `Package` itself.

davidmhewitt:
I'd be tempted to keep the origin strings as consts at the top of the file just so they're big capital letters and easy to spot when we come to change the repo names.

donadigo:
@davidmhewitt made a good point. I'll approve this after his requested change.

donadigo:
Looking good now! :)

PR #373, opened 2017-06-30
Don't show share popover for OS updates

### LINKED ISSUE #368: OS updates should not be shareable
You should not be able to share "OS updates"

danrabbit:
Fixes #368

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -254,6 +254,8 @@ namespace AppCenter.Views {
 
             add (overlay);
 
+            open_button.get_style_context ().add_class ("h3");
+
             if (!package.is_os_updates) {
                 var body = _("Check out %s on AppCenter:").printf (package.get_name ());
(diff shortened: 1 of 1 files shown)


codygarver:
Maybe Operating System Updates shouldn't have a page at all.

cassidyjames:
@codygarver I think we do want it as a potential place for a changelog, though.

tintou:
@cassidyjames that would be a lot of various changelogs.

donadigo:
I think what you should do here perhaps is to have a `can_share` method in the `Package` class and call that instead of throwing that statement there.  One thing that I also thought is that sharing drivers also currently doesn't work.

jeremypw:
If "shareability" is more complex than whether or not a package is an os-update then I agree with donadigo that this should be handled within Package and exposed as a single property.

ramirocantu:
Should drivers have a sharing link?

PR #379, opened 2017-07-01
Simplify view handling

### LINKED ISSUE #369: Cannot redirect to an app when in Updates view
Steps to reproduce:

1. Open AppCenter
2. Switch to updates tab
3. Open a link in a web browser like https://appcenter.elementary.io/com.github.ronnydo.colorpicker.desktop
3. Hit "open"
- AppCenter doesn't show any app

In addition to that, you can hit "Homepage" button. This will not switch anywhere, but it will show you a the stack switcher. When you hit "Homepage" button in the switcher again, the app will show but without the back button so you can't go back to homepage.

donadigo:
This PR simplifies how communication between the MainWindow and it's views happen. Previously the code logic for handling how information would be displayed in the header bar was in several places, and it was not exactly clear what it does in order to show / hide / change widgets. This branch attempts separating this code logic so that it is in it's respective views. In particular the `view_opened` and `view_return` methods don't contain any view-specific code now. they now only take the parameters from each widget about what should it display and do only that. 

This branch should indirectly fix #369.

To test:
There should be no different behaviour than what is now currently in master. What should be tested is if everything in the header bar is updated and works properly when you enter a new view.


donadigo:
@danrabbit fixed.

danrabbit:
Since we just landed a couple of crasher fixes, I'd like to see a new release before we merge this. Then we'll have plenty of time to test for regressions and not feel pressured about releasing :)

danrabbit:
It looks like the issue of being stuck is resolved, but the overall search scope still has a regression. Same reproduction steps:

1. Navigate inside a category (Office)
2. Search for an app that is not inside that category (Vocal)
3. Click "Search Apps" back button.

Expectation:
The scope of the search has been expanded to all apps
screenshot from 2017-07-04 09 41 22

Reality:
We're now viewing the office category without a filter
screenshot from 2017-07-04 09 41 38

donadigo:
@danrabbit should be fixed in the latest commit.

PR #382, opened 2017-07-02
Add a DBus service

### LINKED ISSUE #364: Uninstall dbus service not available if AppCenter is not available
The uninstall service Slingshot uses depends on AppCenter being running on the background BEFORE wingpanel starts up, and this is not always the case (Or it can get killed) and if any of this happens the menu item for uninstalling won't ever appear

We need a dbus .service file so that we can register AppCenter as a "on demand" service. Here is an example of one: https://github.com/Philip-Scott/Spice-up/blob/63cff4969937b5384d389e0e3971c66f924dca46/data/thumbnailer/com.github.philip-scott.spice-up.Thumbnailer1.service

donadigo:
Fixes #364.

Adds a DBus service to AppCenter so that it is registered as a name that can be activated by applications using this DBus API (in particular, Slingshot: https://github.com/elementary/applications-menu/pull/28).

Note: this possibly requires a change in the debian packaging so that the service file is actually installed.


donadigo:
Please *do not* merge / approve this branch yet. It looks like Exec key doesn't accept arguments, hence the AppCenter window opens up on startup.

donadigo:
Ready to review again.

PR #396, opened 2017-07-11
Better Stripe Dialog

### LINKED ISSUE #395: More descriptive subtitle in payment dialog
Right now the payment dialog just lists:

>**AppCenter**
>App Name

We could provide some more detail here and make this look a little nicer while we're at it. At the very least, we could reiterate the chosen payment amount, and add _what_ the user is doing to the app name. i.e.:

>**AppCenter**
>Pay **$5** for App Name

cassidyjames:
WIP. Will fix #395 when merged.


cassidyjames:
screenshot from 2017-07-10 22 00 23

PR #403, opened 2017-07-17
Add option to load local data

donadigo:
This branch adds a new command line `--load-local` option to load AppStream XML data files. This branch also changes that all options are now displayed when you execute `appcenter --help`.

The data is parsed into a package and the package is then added to the banner if parsing was successfull. Packages that are loaded locally have a `(local)` suffix in their name to distinguish them from the actual package with the same name. Local packages are also searchable and appear under native category although they cannot be installed / uninstalled nor updated.

Usage:
`appcenter --load-local /path/to/xml/file`

It might be worth to add this feature to docs so that devs can test their appdata if this gets merged into master.


src/Core/Package.vala
```diff
@@ -442,10 +449,11 @@ public class AppCenterCore.Package : Object {
     }
 
     public Pk.Package? find_package () {
-        if (component.id == OS_UPDATES_ID) {
+        if (component.id == OS_UPDATES_ID || is_local) {
             return null;
         }
 
+
```
danrabbit: extra newline here?

PR #404, opened 2017-07-18
add currency to stripe token creation

btkostner:
Because we always process in USD, we can add the currency field when we create the stripe token. This will stop some errors before it hit's houston's database, and will hopfully give a more descriptive error message.


btkostner:
Can you approve it now Mr Foré?

PR #409, opened 2017-07-22
Blacklist apps

### LINKED ISSUE #26: Add a blacklist
So we can hide some of the crap we don't want to show

Launchpad Details: [#LP1613461](https://bugs.launchpad.net/bugs/1613461) Daniel Fore - 2016-08-15 21:58:15 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123468-add-a-blacklist?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #26.

Adds a simple file `/etc/appcenter/appcenter.blacklist` that enables blacklisting specific AppStream component ID's from being added to packge list (that is also not showing in AppCenter's UI).

The file has a description of it's format and what it's used for. I've created a new `ComponentValidator` class to separate blacklist logic from the already big `Client` class. An example to blacklist QHaveDate application would be to add this line to the blacklist file:
```
qhavedate.desktop
```

The issue report also said that console applications should be disallowed automatically, so I've added this check in the `validate` method in `ComponentValidator`, it can be easily removed if needed though.


src/Core/ComponentValidator.vala
```diff
@@ -0,0 +1,65 @@
...
+            warning ("Could not get the contents of blacklist file: %s", e.message);
+        }
+    }
+
+    public bool validate (AppStream.Component component) {
+        if (component.get_kind () == AppStream.ComponentKind.CONSOLE_APP) {
+            return false;
+        }
+
+        if (component.get_id () in blacklist) {
```
Philip-Scott: Instead of a string array, you should use a hashmap. This will make searches faster by not having to loop though an array on each validation

src/Core/ComponentValidator.vala
```diff
@@ -0,0 +1,65 @@
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
```
Philip-Scott: You should kill this line :)

src/Core/ComponentValidator.vala
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private ComponentValidator () {
+        string blacklist_path = Path.build_filename (Build.CONFIGDIR, "appcenter.blacklist");
+
+        try {
+            string contents;
+            FileUtils.get_contents (blacklist_path, out contents);
+            parse_and_populate (contents);
+        } catch (FileError e) {
```
Philip-Scott: I don't remember if vala arrays are initialized from the start. If the file didn't exist and there's an error, would it crash in Line46? (`if (component.get_id () in blacklist)` )

donadigo: No, it won't crash. You can easily verify that: there is nothing added to `blacklist` currently because there are no entries in the blacklist file. The `blacklist` by default is an empty array.

PR #416, opened 2017-07-29
Remove blank appcenter.blacklist

codygarver:
Remove blank appcenter.blacklist so elementary-default-settings can install over it


codygarver:
The way we handle the wingpanel blacklist is as a file provided by the `elementary-default-settings` package. Because the `appcenter` package currently provides its own blank file, it makes `elementary-default-settings` conflict and refuse to install as an update.

tintou:
Is it possible to just not have it included into the packaging or maybe not install it with CMake, it seems important to me that AppCenter documents the way this blacklist file works

codygarver:
Deleted it using the packaging instead.

PR #421, opened 2017-08-01
Fix update notifications

### LINKED ISSUE #390: Badge notification in Plank isn't removed after updating
After running updates, the badge displaying the amount of updates available is not updated/removed. Appears to affect only the icon in Plank, not in Slingshot. Seems as though this issue can only be fixed by restarting Plank entirely. As far as I'm aware, it started occurring after version 0.2.4 was released.

screenshot from 2017-07-07 20-05-06

davidmhewitt:
Depends on #425. Fixes #390.

For some reason, calling the synchronous version of `get_installed_applications` on PackageKit was causing the `updates_changed` signal to no longer be emitted. Even connecting to the PackageKit D-Bus and listening to the signal on there didn't work.

I've wrapped the async version of the same method in a `MainLoop`. This should keep it synchronous but also prevents the PackageKit issue. I know it's a bit horrible but this is exactly how PackageKit creates sync versions of async methods:
https://github.com/hughsie/PackageKit/blob/master/lib/packagekit-glib2/pk-client-sync.c#L798

I've spent a lot of time trying to figure out why this bug is occuring. It **looks** like an issue in PackageKit, but I can't be 100% sure. I've created a small Vala example that reproduces the issue and submitted it as an issue to the PackageKit repo here: https://github.com/hughsie/PackageKit/issues/207

Requesting @donadigo for testing and review, since he knows this part of the code better than I.


Philip-Scott:
I actually couldn't confirm that it fixed the problem for me... Installed and killed AppCenter from your branch, and after running the updates the badge is still on Plank :/

davidmhewitt:
@Philip-Scott Thanks for testing! Unfortunately, it looks like there was another issue too. #425 will need merging before this one.

davidmhewitt:
The response to the issue I raised against PackageKit suggests we should go with this solution. Maybe @tintou can take a look as this approach was originally recommended against by him.

PR #424, opened 2017-08-01
Implements changelog information in AppInfoView

### LINKED ISSUE #296: Show changes in AppInfoView
Here in this example, there is an update available, but there is no information on whats new. Same for any other update provided by AppCenter.

screenshot from 2017-05-20 09 41 08

donadigo:
Fixes #296. Adds changelog information from AppStream to the AppInfoView. The implementation currently limits the number of releases shown to 5 but it can be easily changed in the declared constants.

If the package is not installed, the view shows only 1 package, otherwise, we calculate what index in the release list is the current installed version and add all the releases from the start of the list to the calulated index.

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -20,6 +20,9 @@
 
 namespace AppCenter.Views {
     public class AppInfoView : AppCenter.AbstractAppContainer {
+        private const int MIN_RELEASES = 1;
+        private const int MAX_RELEASES = 5;
+
         static Gtk.CssProvider? previous_css_provider = null;
 
         GenericArray<AppStream.Screenshot> screenshots;
@@ -28,6 +31,7 @@ namespace AppCenter.Views {
         private Gtk.Stack app_screenshots;
         private Gtk.Label app_version;
         private Gtk.ListBox extension_box;
+        private Gtk.ListBox release_box;
         private Gtk.Stack screenshot_stack;
         private Gtk.TextView app_description;
         private Widgets.Switcher screenshot_switcher;
@@ -161,6 +165,12 @@ namespace AppCenter.Views {
             }
 
             content_grid.add (app_description);
+
+            release_box = new Gtk.ListBox ();
+            release_box.selection_mode = Gtk.SelectionMode.NONE;
+            release_box.margin_top = 12;
+            release_box.hide ();
+            content_grid.add (release_box);
 
             if (package_component.get_addons ().length > 0) {
                 extension_box = new Gtk.ListBox ();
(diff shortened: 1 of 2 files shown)


src/Views/AppInfoView.vala
```diff
@@ -313,6 +323,45 @@ namespace AppCenter.Views {
             });
         }
 
+        private void load_releases () {
```
Philip-Scott: If you're going to make a new class for ReleaseRow, maybe this function could be refactored into it's own class, don't you think? :) 

Because one idea was to also have the Releases in the Updates tab, so maybe this could get reused in the future

codygarver:
I am going to do a new stable release of AppCenter very very soon, please don't merge this branch until after then.

src/Widgets/ReleaseRow.vala
```diff
@@ -0,0 +1,93 @@
...
+        unowned string version = release.get_version ();
+
+        uint64 timestamp = release.get_timestamp ();
+        if (timestamp != 0) {
+            var date_time = new DateTime.from_unix_utc ((int64)timestamp);
+            string format = Granite.DateTime.get_default_date_format (false, true, true);
+            string date = date_time.format (format);
+
+            if (version != null) {
+                label = _("<b>%s</b> • %s").printf (version, date);
```
danrabbit: Probably a more appropriate separate than `•` would be `–`

src/Widgets/ReleaseRow.vala
```diff
@@ -0,0 +1,93 @@
...
+
+        var grid = new Gtk.Grid ();
+        grid.margin_bottom = 6;
+        grid.attach (header_label, 0, 0, 1, 1);
+        grid.attach (description_label, 0, 1, 1, 1);
+
+        add (grid);
+    }
+
+    public ReleaseRow (AppStream.Release release) {
```
Philip-Scott: Just for consistency, can you move this to the top of the construct block? :)

src/Widgets/ReleaseRow.vala
```diff
@@ -0,0 +1,93 @@
...
+        Object (release: release);
+    }
+
+    private static string format_release_header (AppStream.Release release) {
+        string label;
+
+        unowned string version = release.get_version ();
+
+        uint64 timestamp = release.get_timestamp ();
+        if (timestamp != 0) {
```
Philip-Scott: Maybe it would be simpler to test first if version != null :man_shrugging:

Philip-Scott:
I made some comments but it's much better though :)

src/Widgets/ReleaseRow.vala
donadigo: I attempted to check if the `version != null`, but here we want to check for the timestamp first so we can later define if we do not have the version and only have the date show up.

PR #428, opened 2017-08-02
Reconstruct obtaining Pk.Package

donadigo:
This branch reconstructs how `Pk.Package` is obtained within the core `Package` class. I would not like to merge this now, but some time that will let us fix things if something is broken.

The `find_package` was renamed to `obtain_package` and is now asynchronous. The method was restructured so that it will only need to be called one time per package. This change also enables us to remove threading to get more package details like version (most visible in `AppInfoView` class). We also do not need to care about adding UI changes to the `Idle.add` method now which removes possibilities for crashes. There are more smaller changes to adapt to the change like making the `load_screenshot` method asynchronous and moving try-catch clause to it's right place.

The most visible change of this branch is the homepage loading time. In master on my machine showing all trending, recent updated and the banner takes about to 9-10 seconds to complete. With this branch the homepage is populated almost instantly and takes 0-1 second to show all metadata.


danrabbit:
Performance boost confirmed. Good god.

donadigo:
It looks like unfortunately this speed boost was that we didn't call `search_names_async` in the Homepage (`obtain_package`) which takes about 1-2 for each application. This is and will still be a problem here. To get an acceptable speed for app population we will need to everything when in the silent mode.

cassidyjames:
@donadigo is this branch irrelevant then, or would you pick back up on the work here?

danrabbit:
Adam said close it in Slack :)

PR #434, opened 2017-08-03
Add command line option to fake updates

donadigo:
This PR adds a way to fake updates as they were real updates shown in AppCenter. There is a new `--fake-package-update` command line option (`-f` for shortcut). The argument passed to this option is a package name that is recognized by the PackageKit backend (in this case `apt`) like `gimp`, `inkscape`, `com.github.donadigo.eddy` or `switchboard-plug-pantheon-shell`. You can provide multiple packages to the option if you want. Note that this is **not** the same as the AppStream component ID which usually ends with `.desktop`.

When running AppCenter with this option, the update will show in the update list view as well as the updated badge with the fake updates included. You can even update the fake package, the installation will prompt you for authentication and after that, the package will be "updated" as any other normal package except that the installation will be instant and the package will change it's state to installed and appear under up to date header (no actual changes to the package will be done on the system).


donadigo:
@Philip-Scott:

1. This is normal, we parse the options with standard GLib functionality.
2. Resolved.

PR #444, opened 2017-08-09
Support inline comments

### LINKED ISSUE #438: Blacklist does not support comments
If a line in the blacklist has a comment then the line (and possibly subsequent lines) get ignored.

Example:
```
display-im6.desktop # imagemagick
```

donadigo:
Fixes #438.

Makes this type of line valid:
```
some-id.desktop # This is an inline comment
```


src/Core/ComponentValidator.vala
```diff
@@ -57,13 +57,12 @@ public class AppCenterCore.ComponentValidator : Object {
...
-                continue;
-            }
-            
-            string token = line.strip ();
-            if (token != "") {
-                blacklist.add (token);
+            string[] tokens = line.split ("#");
+            if (tokens.length > 0) {
+                string id = tokens[0].strip ();
+                if (id != "") {
```
jeremypw: This assumes the id does not contain a # character.   Not sure if that matters.

donadigo: @jeremypw this doesn't matter since the string is already splitted with "#". So things like `foo.desktop#comment` will pass.

jeremypw: My point was that something like `my#tageditor.desktop # comment` might cause unexpected results but I am not sure whether app names which include a `#` are likely (or possible).  The id will be added to the blacklist as `my`.

Bat41: According to the docs, component IDs may contain a `#`

>  The `<id>` tag is a unique identifier for this component. It must contain only ASCII characters, dots, hyphens and numbers. Spaces are not allowed.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent

donadigo:
Okay, since component ID's can contain `#`, I'll close this as the current format features this character as a comment and I don't want to break that.

PR #445, opened 2017-08-09
Show newest changelog in the updates view

### LINKED ISSUE #24: Show changelog in updates [$100]
In the updates view, we should show the changelog for packages that have updates

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123462-show-changelog-in-updates?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #24.

This is work in progress, but I would still like to get feedback how can I fix some problems with the UI because I just can't get over them:
* The "Update" button stays always in the middle of the package row, even if you expand the version description
* When expanding the version description, the expander changes it's place and so the expander arrow changes it's place so that you can't click it again without moving the mouse.

Some changes in the code to make the PackageRow display the release information is moving some methods from `ReleaseListBox` and `ReleaseRow` to the core `Package` class so they can be used outside of these specific widgets.

### DIFF
--- a/src/Widgets/PackageRow.vala
+++ b/src/Widgets/PackageRow.vala
@@ -22,16 +22,16 @@ namespace AppCenter.Widgets {
     public class PackageRow : Gtk.ListBoxRow, AppListRow {
         AbstractPackageRowGrid grid;
 
-        public PackageRow.installed (AppCenterCore.Package package, Gtk.SizeGroup? size_group, bool show_uninstall = true) {
-            grid = new InstalledPackageRowGrid (package, size_group, show_uninstall);
+        public PackageRow.installed (AppCenterCore.Package package, Gtk.SizeGroup? info_size_group, Gtk.SizeGroup? action_size_group, bool show_uninstall = true) {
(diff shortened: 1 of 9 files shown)


danrabbit:
screenshot from 2017-08-09 16 12 14

This is less than ideal :p

src/Widgets/PackageRow.vala
```diff
@@ -72,57 +72,107 @@ namespace AppCenter.Widgets {
...
-                app_version.hexpand = true;
                 app_version.valign = Gtk.Align.START;
                 ((Gtk.Misc) app_version).xalign = 0;
 
-                attach (app_version, 1, 1, 1, 1);
+                release_description = new Gtk.Label (null);
+                release_description.selectable = true;
+                release_description.use_markup = true;
+                release_description.wrap = true;
+                release_description.margin_start = 14;
```
danrabbit: 12; multiples of 3 :p

PR #446, opened 2017-08-09
Do stuff to fix alignment

danrabbit:


danrabbit:
Fixed the size group issue. Seems like Gtk doens't like a widget to belong to more than one ;p

I can't confirm an issue with the expander, but I think this needs some further work anyways. It doesn't currently show any useful information in its collapsed state, just a duplication of the release version. It probably needs further thought.

Fixed the position of the action stack

PR #449, opened 2017-08-10
AppContainers: one class per file

danrabbit:
Moves each of these classes to its own file


src/Widgets/AppContainers/AbstractPackageRowGrid.vala
```diff
@@ -0,0 +1,68 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Corentin Noël <corentin@elementary.io>
+ */
+
+private abstract class AppCenter.Widgets.AbstractPackageRowGrid : AbstractAppContainer {
```
donadigo: There's no reason for this class to be private. Private classes are contained only within some other class.

src/Widgets/AppContainers/InstalledPackageRowGrid.vala
```diff
@@ -0,0 +1,132 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Corentin Noël <corentin@elementary.io>
+ */
+
+private class AppCenter.Widgets.InstalledPackageRowGrid : AbstractPackageRowGrid {
```
donadigo: The same here.

src/Widgets/AppContainers/ListPackageRowGrid.vala
```diff
@@ -0,0 +1,48 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Corentin Noël <corentin@elementary.io>
+ */
+
+private class AppCenter.Widgets.ListPackageRowGrid : AbstractPackageRowGrid {
```
donadigo: The same here.

PR #454, opened 2017-08-13
Apps by author

### LINKED ISSUE #451: Show other apps by author
screenshot from 2017-08-11 11 52 23

I started prototyping this a bit here: https://github.com/elementary/appcenter/tree/apps-by-author

- [x] I think in the interest of keeping the AppInfoView from growing forever it would probably be a good idea to break this out into its own widget that takes an author as a construction argument.

- [x] Something that I think I'd also like to do as a pre-req is to move the function for opening the app page from the homepage into the carousel itself

- [x] We'll need to also handle if there are no other apps by this author

donadigo:
Fixes #451.

Adds a bottom bar to the app page where all other apps by the same author are displayed. The PR should be reviewed carefully since it touches e.g: how the HeaderBar UI is updated.


src/Views/AppInfoView.vala
```diff
@@ -261,6 +263,36 @@ namespace AppCenter.Views {
             grid.attach (content_grid, 0, 1, 1, 1);
             grid.attach (footer_grid, 0, 2, 1, 1);
 
+            if (package.get_author () != null) {
+                var other_apps_header = new Gtk.Label (_("Other Apps by %s:").printf (package.get_author_title ()));
```
danrabbit: I guess in other headers we don't use ":"

src/Core/Package.vala
```diff
@@ -386,6 +388,37 @@ public class AppCenterCore.Package : Object {
...
+
+            if (project_group != null) {
+                author = project_group;
+            }
+        }
+
+        return author;
+    }
+
+    public string get_author_title () {
```
danrabbit: Can we use properties instead of `get` methods please? It's more vala style to have `package.author_title`

src/Core/Package.vala
```diff
@@ -386,6 +388,37 @@ public class AppCenterCore.Package : Object {
         return latest_version;
     }
 
+    public string? get_author () {
```
danrabbit: Can we use properties instead of `get` methods please? It's more vala style to have `package.author`

src/Widgets/Carousel/AuthorCarousel.vala
```diff
@@ -0,0 +1,39 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Adam Bieńkowski <donadigos159@gmail.com>
+ */
+
+public class AppCenter.Widgets.AuthorCarousel : Carousel {
+    private const int AUTHOR_OTHER_APPS_MAX = 5;
```
danrabbit: Does it makes sense to have a max here?

src/Widgets/Carousel/Carousel.vala
```diff
@@ -18,19 +18,32 @@
...
-                    max_children_per_line: 5,
-                    min_children_per_line: 2);
-        }
+    public signal void package_activated (AppCenterCore.Package package);
+
+    construct {
+        child_activated.connect (on_child_activated);
+    }
+
+    public Carousel () {
```
danrabbit: Is there a reason you moved this below `construct`? Typically we put it above

donadigo:
@danrabbit issues addressed.

danrabbit:
It looks like navigation does get a little fucky when you navigate too many levels down:

1. Navigate: Accessories > Calculator > Music > Calendar
2. Click back twice

Expectation: You're at Calculator
Reality: You're at Accessories

Philip-Scott:
@danrabbit This is actually an external problem and doesn't have to do with this branch. You can test it by:

- Audio > Vocal
- https://appcenter.elementary.io/com.github.philip-scott.spice-up.desktop 
- https://appcenter.elementary.io/com.github.danrabbit.harvey
- Click back, you're still in Audio

danrabbit:
Yeah that's true. I think that this feature really easily exposes the bug though. I'll dismiss my review so this branch isn't blocked, but my personal stance I think is that we should fix that navigation issue first before merging this

danrabbit:
Looks like even after that recent merge, the back navigation can still get messed up easily

PR #455, opened 2017-08-13
Don't create popovers everytime in HumbleButton

### LINKED ISSUE #447: HumbleButton creates popover everytime it's created
The `HumbleButton` should not create a popover everytime it is constructed. This leads to this:
popovers

donadigo:
Fixes #447.

Does what it says. Creates a popover only when the arrow is actually clicked.


Philip-Scott:
Why not go a step further and also only construct the contents of the popover along with it? :)

PR #457, opened 2017-08-14
Set AppInfoView icon size 60.

kmltrk:
Because low resolution icons not looking good.I think is better now.


danrabbit:
I'm going to close this as a "won't fix". This is an issue with appstream that's being resolved there.

kmltrk:
Ok i understand.Thanks for interest.

PR #475, opened 2017-09-11
Fix navigation bugs

DDieter:
Fixes some navigation regressions and issues caused by the possibility to instantly view a certain package using an AppStream URL.

- Fix navigation issue described in #454
- Fix search bar becoming sensitive in AppInfoView after internet connection returns
- Changed search behaviour to improve navigation and avoid some minor issues
- Fix issue where opening a URL like https://appcenter.elementary.io/com.github.danrabbit.harvey while searching would have no effect until exiting the search view


src/MainWindow.vala
```diff
@@ -44,6 +44,8 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
     private int homepage_view_id;
     private int installed_view_id;
 
+    private static int valid_query_length = 3;
```
tintou: use a const here instead of static, this requires a rename to uppercase "VALID_QUERY_LENGTH"

Issue #486, opened 2017-09-21
Add Updates quicklist action


cassidyjames:
It'd make it show up in Slingshot as an entry when you search Updates, and would be handy in the dock.

skyiker:
ghbdtn

PR #497, opened 2017-10-09
Add quicklist action to show updates

zeebok:
This addresses #486 and adds a quicklist action to show updates.


gcjuan:
Could be added a different translation of this quick action depending on the language of the system?
I have changed "Check for updates" on my system to Spanish "Buscar actualizaciones" modifying the file "/usr/share/applications/io.elementary.appcenter.desktop" but it is not the most professional way I mean...
I installed Firefox 57 and it has the correct quick actions depending on the system language (as an example).

PR #503, opened 2017-10-24
Show an error when updating cache failed

donadigo:
**Depends on a new release of Granite.**

Shows an error info bar with "Try Again" button when updating the cache fails for some reason.


danrabbit:
Wouldn't this error mean no package operations can be taken until it gets resolved? An info bar seems kind of passive for an error of this magnitude.

donadigo:
@danrabbit no, since the user will still be able to see installed apps in some cases.

donadigo:
The branch is now ready to review.

danrabbit:
screenshot from 2017-10-25 13 06 30

With this much text, I think this should be a dialog.

snwh:
definitely too verbose for an infobar

cassidyjames:
I feel like this is pretty heavy, technical information to be showing to the user by default, as well. I'd be okay with an error infobar if it was:

>Failed to fetch updates. [Details] [Try Again]

or something

donadigo:
@cassidyjames @danrabbit done.

PR #506, opened 2017-10-26
Make SharePopover widget code more readable

ypoz:
Okay, here's more polished SharePopover widget code, with bulky and repetitive code moved into helper methods. Its crystal clarity will astound generations to come, if I do say so myself.

Now why I don't want to make helpers into widgets. This just seems counterproductive to me: a widget has to hold its state and have its signals, so there'll be twice as much code, and extra plumbing. Unless there's a pressing need to use those buttons elsewhere I strongly suggest leaving the code as it is. Lean code and stuff, you know.

I've also spotted a possible problem: the sharing URIs are not escaped, so crafted package metadata may or may not have unwanted consequences. This should be a quick fix. Should I post it here or make a separate PR?


danrabbit:
One thing I really don't like about these kinds of helper methods is that they are very impure functions. You're acting on things that haven't been passed in to the function, such as the popover, body text, and the uri. This makes it a lot harder to follow along and if modifications are made in the future they'll need to be done in multiple functions in order to keep things from breaking

src/Widgets/SharePopover.vala
```diff
@@ -31,135 +31,76 @@ public class SharePopover : Gtk.Popover {
...
-            } catch (Error e) {
-                warning ("%s", e.message);
-            }
-            hide ();
-        });
-
-        facebook_button.clicked.connect (() => {
+    Gtk.Button make_share_button (string name, string uri_template, string icon_name) {
+        var share_button = new Gtk.Button.from_icon_name (icon_name, Gtk.IconSize.DND);
+        share_button.tooltip_text = name;
```
danrabbit: In general, it's probably best to name these variables the same as the properties they correspond to for clarity. So instead of `name` this should be `tooltip_text`

src/Widgets/SharePopover.vala
```diff
@@ -20,146 +20,87 @@
 public class SharePopover : Gtk.Popover {
     public signal void link_copied ();
 
-    public string body { get; set; }
-    public string uri { get; set; }
+    public string shared_text { get; set; }
+    public string shared_link { get; set; }
```
jeremypw: Make setter private - these properties do not change over the popover lifetime.

src/Widgets/SharePopover.vala
```diff
@@ -20,146 +20,87 @@
...
+    Gtk.Button make_share_button (string tooltip_text, string uri_template, string icon_name) {
+        var share_button = new Gtk.Button.from_icon_name (icon_name, Gtk.IconSize.DND);
+        share_button.tooltip_text = tooltip_text;
+        share_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+        share_button.clicked.connect (() => {
             try {
-                AppInfo.launch_default_for_uri ("https://www.tumblr.com/share/link?url=%s".printf (uri), null);
+                // Make the text and the uri safe to include in an uri
+                string escaped_link = GLib.Uri.escape_string (shared_link);
+                string escaped_text = GLib.Uri.escape_string (shared_text);
```
jeremypw: This checking should be done in the constructor.

src/Widgets/SharePopover.vala
```diff
@@ -20,146 +20,87 @@
...
-
-        telegram_button.clicked.connect (() => {
-            try {
-                AppInfo.launch_default_for_uri ("https://t.me/share/url?url=%s".printf (uri), null);
-            } catch (Error e) {
-                warning ("%s", e.message);
-            }
+        return share_button;
+    }
+
```
jeremypw: Instead of this helper function, make a private class "ShareButton" with this as a method and pass in the (sanitized) shared_text and shared_link properties on construction.

danrabbit:
Closing because this branch appears to have been abandoned

PR #517, opened 2017-10-30
Get DBus path from CMake variable

codygarver:


tintou:
I don't think that it's necessary, discussion welcome

danrabbit:
Closing this since Pop doesn't actually change this in their fork

Issue #433, opened 2017-08-03
Show optional details for Operating System Updates


cameralibre:
This is a feature request  - so I'm unsure if it should be posted here or Bountysource, or StackExchange. I couldn't find a clear explanation of where to go on the website. 
I'm happy to post a bounty, but I'm unsure if the lack of this feature is a conscious design decision.

**I would like there to be optional 'details' information for the 'Operating System Updates' catchall in AppCenter.**

I would like to know what I'm updating so that:
- I have a general understanding of the state of development of the programs I'm running
- to find out if some new update of a favorite program has landed
- to know how critical/useful the updates are, if they are likely to affect the performance of my system, and therefore if it is important to update now, or if it can wait. 

Currently I switch to the terminal and check with apt, but I would much prefer to see this information in the AppCenter - it is more convenient, immediate and the information is presented in a structured, visually clear way.

But perhaps showing this information goes against elementary's design approach of simplifying the interface, in which case I guess a bounty wouldn't be welcome. 

Can somebody confirm if this is possible, or if it is a conscious decision not to provide this information?

peteruithoven:
Part of this seems to be covered in the following issue https://github.com/elementary/appcenter/issues/296 ?

cassidyjames:
This seems related to changelogs, but not the same. One way to solve this might be to basically set a "changelog" for the OS updates to a list of packages and versions? That's fairly technical but would be comprehensive.

cassidyjames:
Thoughts from @elementary/ux?

danrabbit:
Yeah there's probably not going to be a good way to get a human-readable changelog for this. I know @davidmhewitt had a branch that did the package list thing at one point.

I think maybe a good middle ground would be showing something generic on the UpdatesListView like "Security & Stability updates. Platform libraries" etc. And then in the appinfoview show a list of packages and versions.

harveycabaguio:
I don't know how feasible it would be but it might be smart if the medium article for the  bigger updates could be embedded to this page.

cassidyjames:
@jackpot51 implemented something like this in the Pop!_Shop fork.

Relevant commits I see:
- https://github.com/system76/pop-shop/commit/8051303927972b16edd41c4d846af28642843002
- https://github.com/system76/pop-shop/commit/56029b79f3cdd5014de5ff04107c05d5d8ec7a0d
- https://github.com/system76/pop-shop/commit/7c31b3d75129fe20a46e3cf1f9a12e60d2b9cc78

[...]

PR #518, opened 2017-10-30
Add package information to Operating System Updates details

jackpot51:
This adds information about each package included in Operating System Updates - version and description.

It also fixes a bug where the Operating System Updates can appear more than once, if there are updates and `sudo apt update` is run while the AppCenter is open. This fix makes the Operating System Updates always visible, even when there are no updates, which may not be desired.

Translation will be required for the component(s) with updates string.


jackpot51:
This fixes https://github.com/elementary/appcenter/issues/433

jackpot51:
Sweet!

PR #522, opened 2017-10-31
Create package index class

donadigo:
This branch begins a series of branches to eventually speed up the AppCenter's backend to match the performance from #428. 

The PR adds a simple, new `PackageIndex` class which is used for storing `Pk.Package`s and some useful signals like `added` and `removed`. The class itself will be probably useful when constructing UI's for e.g: the current installation queue. The UI will be just selecting the right package index and will reflect it's state. The `ready` property is used to notify all objects that the index is now ready to read and is populated.

The first set of refactor is the installed packages list. Now, they are stored in the `PackageIndex` as a property in the `Client` class. Notice the list is populated only once through `create_package_indexes` on construction in `Application` class. In master, different parts of AppCenter could call `get_installed_applications` which explicitly called `get_packages_sync` and took time to get the same list of installed packages over again.


danrabbit:
Gonna close this since it now pretty severely conflicts with master and hasn't had any movement in a long time

PR #528, opened 2017-11-06
Get settings base from CMake variable

codygarver:


cassidyjames:
@tintou the idea was to make it easier for downstreams. But pop-os/shop has reverted to just using the same namespace, so :man_shrugging:

tintou:
@cassidyjames Yeah, that's because there is no reason for downstream to change this

danrabbit:
Closing since the only downstream that would use this doesn't want to use this

PR #537, opened 2017-11-12
Escape XML characters in OS description

### LINKED ISSUE #532: Operating System Updates showing html tags
image

donadigo:
Fix #532.

Escapes XML characters out of the package name, summary and the version so that we don't have errors while parsing those by the XML parser.


tintou:
Better use this to escape things : https://valadoc.org/glib-2.0/GLib.Markup.escape_text.html

donadigo:
@tintou fixed.

tintou:
In this context, GLib.Markup.printf_escaped is even better

donadigo:
@tintou done, thanks for the suggestion.

PR #550, opened 2017-12-02
Allow translating app options

Thalexander:


src/Application.vala
```diff
@@ -17,13 +17,13 @@
 public class AppCenter.App : Granite.Application {
     public const OptionEntry[] APPCENTER_OPTIONS =  {
         { "show-updates", 'u', 0, OptionArg.NONE, out show_updates,
-        "Display the Installed Panel", null},
+        N_("Display the Installed Panel"), null},
         { "silent", 's', 0, OptionArg.NONE, out silent,
-        "Run the Application in background", null},
+        N_("Run the Application in background"), null},
```
jeremypw: Mixed style - use definite article consistently in accordance with elementary style.

src/Application.vala
```diff
@@ -17,13 +17,13 @@
 public class AppCenter.App : Granite.Application {
     public const OptionEntry[] APPCENTER_OPTIONS =  {
         { "show-updates", 'u', 0, OptionArg.NONE, out show_updates,
-        "Display the Installed Panel", null},
+        N_("Display the Installed Panel"), null},
```
jeremypw: "Display the installed applications" better?  I am not sure title case is required in this situation (check with Dan)

src/Application.vala
```diff
@@ -17,13 +17,13 @@
...
         { "silent", 's', 0, OptionArg.NONE, out silent,
-        "Run the Application in background", null},
+        N_("Run the Application in background"), null},
         { "load-local", 'l', 0, OptionArg.FILENAME, out local_path,
-        "Add a local AppStream XML file to the package list", "FILENAME" },
+        N_("Add a local AppStream XML file to the package list"), N_("FILENAME") },
         { "fake-package-update", 'f', 0, OptionArg.STRING_ARRAY, out fake_update_packages,
-        "Add the package name to update results so that it is shown as an update", "PACKAGES..." },
+        N_("Add the package name to update results so that it is shown as an update"), N_("PACKAGES...") },
```
jeremypw: Mixed style - use definite article consistently (" .. the update results ..").  Maybe a more concise description would be better, also recognise that the argument is an array with possibly more than one package - e.g "Add these packages to the list of packages to be updated".

tintou:
Okay, I'm calling the cops
@danrabbit

danrabbit:
Hm, I would go sentence case here since it's more like a tooltip than a button label

src/Application.vala
danrabbit: Sentence case is probably correct here since it's more like a tooltip than a button label. We usually don't use articles in verbal phrases like this though and since this goes to the "Updates" view we probably should refer to it that way, so maybe "Display updates view" ?

src/Application.vala
danrabbit: If you dropped the article you could do "Add package(s) to updates view for testing" or something. Probably a good idea to make it clear that this is about testing and essentially adds a dummy item

danrabbit:
Gonna close this since no movement since December

PR #556, opened 2018-01-02
Use request body to send the data

Philip-Scott:
Things to test:

- Make sure you can still make a purchase :)


src/Dialogs/StripeDialog.vala
```diff
@@ -440,10 +453,16 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         return data.str;
     }
 
-    private string post_to_houston (string _app_key, string _app_id, string _purchase_token, string _email, string _amount) {
-        var uri = HOUSTON_URI.printf (_app_id, _app_key, _purchase_token, _email, _amount);
+    private string post_to_houston (string _app_key, string _app_id, string _purchase_token, string _amount) {
```
btkostner: You will still want to include the email in the POST body so we can keep on sending the receipts. Forgot to put that in the example. My bad.

PR #561, opened 2018-01-08
Show humblebutton on attempting to update unpaid applications

### LINKED ISSUE #502: Don't install unpaid apps when clicking "Update All"
Relies on #306 

Trying to think of a fair way to "nag" people about paying for the apps they've previously downloaded for free:

A user downloaded an app that is a suggested price of $1 for free. A week later there's an update for the app. When the user visits the updates page, they click update all.

Instead of installing all updates, AppCenter should only install updates which are either free or already paid for.

Apps which have not been paid for should display their suggested price and require the user to either pay or once again enter $0 to get the update.

In this way, users can still choose not to pay for the apps, but it becomes less convenient not to pay. At this point in time, we can probably assume that the user is getting value from the app and it's a fair time to re-prompt them to support its development

davidmhewitt:
Fixes #502 

This is currently untested as I don't have any applications with a suggested price that need updates yet.

To test, please make sure you clear out the `/io/elementary/appcenter/settings/paid-apps` dconf key to imply you haven't paid for any apps first.

If this branch is working correctly: 

* The "Update All" button should only install updates for applications that are in the `paid-apps` array (and non AppCenter native apps). 

* The HumbleButton should be shown set to the applications suggested price for the ones that have not been paid for.


danrabbit:
It works!

The only thing that's weird is that in the case that the only updates left are unpaid apps, the "Update All" button is still sensitive, yet will do nothing.

davidmhewitt:
@danrabbit Makes sense!

I think I've fixed that? But I still don't have any updates to run!

danrabbit:
screenshot from 2018-01-08 17 04 51

Ignore that there are two sections for "Up to date". That's a bug in master unfortunately.

I think it would probably be better to just set the button to insensitive here than to try to remove it

davidmhewitt:
@danrabbit Yeah sorry, testing blind here with having no updates...

Think that should do the trick.

benwaffle:
@danrabbit Does this show that updates for torrential cost money, or can you get them for free through the dropdown

PR #563, opened 2018-01-10
Add more opportunities to pay

### LINKED ISSUE #281: Always display a way to pay [$60]
If someone already installed an app but wants to come back to the AppCenter listing, we should let them pay the developer there.

davidmhewitt:
Fixes #281 

This also depends on/incorporates the diff of #561 so that should be merged first. Blocked pending new stable release of AppCenter, however.

This adds a "Fund" button/link to the bottom of each app's page (where payments are available), it uses the popover from the HumbleButton to present payment options, but doesn't allow the $0 option.

Secondly, for apps that have been installed, not paid for, and don't require updates; the HumbleButton is shown in the updates view as a bit of a reminder you haven't paid for that application yet. Again, this doesn't allow the $0 option, but disappears once you've paid for the app.

This maybe requires some UX input, checking whether all these new ways of paying complement each other and aren't confusing.


src/Views/AppInfoView.vala
```diff
@@ -540,5 +552,34 @@ namespace AppCenter.Views {
...
+                                                          );
+
+                    stripe.download_requested.connect (() => {
+                        Settings.get_default ().add_paid_app (package.component.get_id ());
+                    });
+
+                    stripe.show ();
+                });
+
+                tooltip_text = _("Fund the development of this project");
```
danrabbit: I think we should use "app" here instead of "project". I feel like "project" carries hobby connotations and sounds less like a real valuable product

danrabbit:
I wonder if we need a button in the Fund popover in the footer for when you make a custom selection. It's not clear after you've changed the price there how to actually trigger the dialog with that custom price. @elementary/ux

cassidyjames:
@danrabbit yeah I can see the slight weirdness there since we don't have a split button like the install button. Would it make sense to switch to that sort of button, and have a default fund amount? 

i.e. `Fund $10 | ▲ `

That way we always have the "Fund" part of the button as an explicit action if people don't know they can hit Enter, plus it works more similarly to the existing Install button.

…I don't know how that'd look in the flat style, though.

cassidyjames:
After a conversation with @danrabbit we were wondering if we should split up the two ways to pay here into different branches.

>This adds a "Fund" button/link to the bottom of each app's page (where payments are available), it uses the popover from the HumbleButton to present payment options, but doesn't allow the $0 option.

This we're having a bit of trouble figuring out the exact UX for, and probably needs some more thought and prototyping.

>Secondly, for apps that have been installed, not paid for, and don't require updates; the HumbleButton is shown in the updates view as a bit of a reminder you haven't paid for that application yet. Again, this doesn't allow the $0 option, but disappears once you've paid for the app.

This part I'm all for and would love to merge.

cassidyjames:
After playing with this, I think the fund button is actually fine as is. It's pretty natural to type the amount and hit enter, and you can always back out of the payment dialog if that wasn't what you intended.

@danrabbit thoughts? I'm open to future improvements, but don't want to artificially delay this any more if it's better than the current experience, which I think it is.

cassidyjames:
Looks like the build is passing CI in Loki, but failing for Juno.

danrabbit:
That's @tintou's fault for naming his meson packaging branch as `deb-packaging-juno` instead of `deb-packaging-meson` :p

PR #567, opened 2018-01-16
Fix Weblate merge conflicts in nl.po

codygarver:


po/nl.po
```diff
@@ -416,19 +409,10 @@ msgstr ""
...
 msgid "The amount is incorrect. Please try again using a valid amount."
 msgstr "Het bedrag is onjuist. Probeert u het nogmaals met een geldig bedrag."
-
-<<<<<<< 
+ 
 #: ../src/Dialogs/StripeDialog.vala:531
 msgid "The pin number is incorrect. Please try again using the correct pin."
 msgstr "De pincode is onjuist. Probeer het opnieuw met de juiste pin."
-=======
-#: ../src/Dialogs/StripeDialog.vala:557
```
davidmhewitt: This looks like the correct source string and line, rather than the other one:

https://github.com/elementary/appcenter/blob/master/src/Dialogs/StripeDialog.vala#L557

PR #584, opened 2018-01-28
Allow critcal urgency releases to update unhindered

### LINKED ISSUE #583: Don't hold back updates with critical urgency
If a release of an unpaid and monetized app has a "critical" urgency, we should make sure not to hold back that update

davidmhewitt:
Fixes #583 

Not tested, but should be able to do something like this with the newest release in the appstream metadata:

    <releases>
    ​  <release version="1.2" date="2014-04-12" urgency="critical" />
    ​  <release version="1.0" date="2012-08-26" />
    ​</releases>


src/Core/Package.vala
```diff
@@ -91,6 +91,10 @@ public class AppCenterCore.Package : Object {
                 return false;
             }
 
+            if (get_newest_release ().get_urgency () == AppStream.UrgencyKind.CRITICAL) {
```
donadigo: Please check if `get_newest_release` didn't return `null` before checking it's urgency.

PR #605, opened 2018-02-10
Encode Stripe data

btkostner:
Encodes Stripe data in hopes to avoid utf-8 encoding issues


danrabbit:
I can confirm that this works :)

PR #612, opened 2018-02-11
Show a download size next to the install button for apps that aren't installed yet

### LINKED ISSUE #411: Display the sizes of Apps before download
Show the user the size of the app before downloading. This will be extremely helpful for users on metered networks or users with limited data plans

davidmhewitt:
Fixes #411 

This calculates the download size of an application plus the download size of its dependencies that aren't currently installed. It displays the result of the calculation as shown below. May be more appropriate to place it somewhere else, but that's a UX question.

image


peteruithoven:
In the updates screen the size is shown on the left of the Update all button:
screenshot from 2018-02-12 21 21 20

cassidyjames:
Hm, I think this would be communicated as well without the label, i.e. just "1.3 MB" instead of "Download size: 1.3 MB".

PR #613, opened 2018-02-12
[WIP] Speedup homepage by searching in parallel

davidmhewitt:
This depends on a fix for what I think is a bug in PackageKit, PR submitted here:
https://github.com/hughsie/PackageKit/pull/240

It is possible to test PackageKit from git without installing it, hit me up in Slack if you want to know how so you can test this branch. But it's probably worth waiting to see if the above PR is considered a bug and whether or not my PR is accepted or not first.

Testing on my machine brings the time taken to load the homepage with the 3 carousels fully populated down from around 25 seconds to around 4-5. We were previously searching PackageKit one package at a time, when I think in theory you should be able to dump an array of package names on it and have it return them all at once.

I still need to do some testing and tidying up, but I'm waiting on the outcome of the PackageKit PR first.


davidmhewitt:
Upstream PR has been merged, so we'll definitely be able to get a significant speedup by utilizing this in the future. Just awaiting a new stable release of PackageKit now.

davidmhewitt:
A new version of PackageKit has been released (1.1.10). I don't know if Ubuntu will take this version, but if this version was available, we could utilise the speedups this branch provides.

davidmhewitt:
Marking this as blocked as it looks like the bionic packages have stabilised on PackageKit 1.1.9 now, so unless we package 1.1.10 for Juno or backport my patches, we won't be able to merge this.

cassidyjames:
@codygarver would it be sane to package 1.1.10 in Juno?

davidmhewitt:
Just leaving myself a note here that this is a breaking change for older PackageKit so would affect at least Pop!_OS. Probably worth making a conditional compile or something....

davidmhewitt:
Closing this as it's now so far out of date with master that it's easier to just make a new branch from scratch.

PR #616, opened 2018-02-15
use houston to test appcenter

btkostner:
This replaces the docker images and uses houston's new v2 worker for travis tests.


danrabbit:
Since this includes validation for metadata we should probably remove the metadata tests here: https://github.com/elementary/appcenter/blob/master/CMakeLists.txt#L31

btkostner:
Should be good now. Once this gets merged, I'll work on adding it to the other projects.

btkostner:
:shipit: :confetti_ball: :tada:

PR #621, opened 2018-02-25
Enable HiDPI icons and 128px icons

tintou:
This won't really change anything in appcenter until this https://github.com/ximion/appstream/pull/179 is merged

### DIFF
--- a/debian/60appcenter
+++ b/debian/60appcenter
@@ -1,11 +1,23 @@
 Acquire::IndexTargets {
+    # The 64x64px icons
     deb::DEP-11-icons  {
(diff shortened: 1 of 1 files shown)


cassidyjames:
ximion/appstream#179 has been merged, we should make sure we have an up-to-date appstream and make sure this works as expected.

cassidyjames:
@tintou it looks like we still have 0.11.8 in Juno, though 0.12 was released recently. @codygarver can we get 0.12 into the OS patches?

codygarver:
appstream 0.12 is in Ubuntu's proposed repo.

Issue #641, opened 2018-03-03
Attempt to auto-format card numbers


cassidyjames:
It sounds like auto-formatting (or just auto inserting space if #640 is implemented) might be worthwhile. There's some info here: https://baymard.com/blog/credit-card-field-auto-format-spaces

isantop:
My Vala-fu is weak, but here's some code to do this in python. If someone can port it to Vala, consider it public domain:

```python3
#!/bin/python3

# Simple Credit Card number formatter

formats = {
    '1' : ['0','4','9','15'],
    '3' : ['0','4','10','14'],
    '37' : ['0','4','10','15'],
    '34' : ['0','4','10','15'],
    '4' : ['0','4','8','12','16']}

formats_length = {
    12 :  ['0','4','8','12'],
    14 :  ['0','4','8','12'],
    16 :  ['0','4','10','15'],
    16 :  ['0','4','8','12','16'],
    18 :  ['0','4','8','12','16','18'],
    19 :  ['0','4','8','12','16','19'],
    110 :  ['0','4','8','12','16','110']}

def get_format(num):
    global formats
    global formats_length
    formatting = 'len'
    for iin in formats:
        if num.startswith(iin):
            formatting = formats[iin]
    if formatting == 'len':
        formatting = formats_length[len(num)]
    list_index = 0
    num_list = []
    while list_index < len(formatting)-1:
        this_index = int(formatting[list_index])
        next_index = int(formatting[list_index + 1])
        num_list.append(num[this_index:next_index])
        list_index = list_index + 1
    
    if len(num) > 16:
        num_list.append(num[next_index:])
    return num_list
```

PR #646, opened 2018-03-04
stripe dialog: new CardNumberEntry widget

tintou:
Let's try this.
Fixes: #641


cassidyjames:
Tested with a VISA card and this worked fine. I don't have a bunch of other test cases to work with, though.

PR #649, opened 2018-03-05
Add logic to update badge

isantop:
This hooks up @danrabbit 's update badge to the logic behind the client, which allows it to be shown and hidden based on whether updates are available or not. It also updates the label text with the correct number of updates available.

This is the fixed PR.


danrabbit:
Does the thing :tada:

PR #652, opened 2018-03-07
Display HumbleButton for unpaid apps not requiring updates

davidmhewitt:
As per a comment by @cassidyjames , this splits out the part of PR #563 that does the following:

>  For apps that have been installed, not paid for, and don't require updates; the HumbleButton is shown in the updates view as a bit of a reminder you haven't paid for that application yet. Again, this doesn't allow the $0 option, but disappears once you've paid for the app.


davidmhewitt:
I've just realised that most of this diff isn't necessary for this change. I'm still splitting out the HumblePopover from the HumbleButton, which isn't necessary for this, as a vanilla HumbleButton is used in the updates view.

I'll cut this down when I'm back from work later, but marking as in progress for now.

PR #659, opened 2018-03-13
Increase default window height to 660

codygarver:


codygarver:
Another downstream change seen in `Pop!_Shop`.

PR #662, opened 2018-03-14
Make UpdateFail dialog a singleton

### LINKED ISSUE #650: Error dialogs spawn infinitely
Observable on Juno daily iso, see attached screenshot.

appcenter-infinite-failure-dialogs-bug

danrabbit:
Fixes #650


snowparrot:
To be honest I dont think it is a singleton...

donadigo:
@snowparrot yeah it's not, but it shouldn't be anyway.

PR #665, opened 2018-03-19
Add package appcenter-casper

codygarver:


codygarver:
Adapted from adapted from `pop-shop-casper`. Fix elementary/os#11.

Issue #328, opened 2017-06-03
Retain last used window position


kdekooter:
If you open AppCenter it is always displayed in the center of the screen (is that an eOS convention?) and quite small. I always have to enlarge it to be useful.

It would be nice if at least the window size is retained for the next time you start AppCenter.

danrabbit:
Bumping this report. It looks like we currently save size, but not position

PR #671, opened 2018-03-31
Added last used window position to saved settings

zeebok:
This pertains to #328 and adds a state to Settings for the window position. I used the default position setting from Code as I have not read about a convention for that but am open to suggestions of a better one.


schemas/io.elementary.appcenter.gschema.xml
```diff
@@ -15,6 +15,16 @@
       <summary>Window height</summary>
       <description>Window height</description>
     </key>
+    <key type="i" name="window-x">
+      <default>250</default>
```
mdh34: I would set this to -1 as default and do some checking in the code

schemas/io.elementary.appcenter.gschema.xml
```diff
@@ -15,6 +15,16 @@
...
+      <summary>Window X position</summary>
+      <description>Window screen position on the X-axis</description>
+    </key>
+    <key type="i" name="window-y">
+      <default>250</default>
```
mdh34: same here

src/MainWindow.vala
```diff
@@ -66,6 +66,7 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
                 this.maximize ();
                 break;
             default:
+                move (saved_state.window_x, saved_state.window_y);
```
mdh34: Add an if statement to check for defaults here so we don't move the window to 250x250 on the first launch - let the window manager handle that

artemanufrij:
could you test your branch... after I installed it appcenter starts always on start up (elementary)

mdh34:
@artemanufrij do you mean it sits in the dock with an update badge?

zeebok:
I made the changes but upon testing I am also seeing the issue of the window appearing at start-up. Quick glance doesn't really give me an idea why as silent mode is suppose to skip the creation of the window. I'll take a look into this a little more tomorrow.

Issue #498, opened 2017-10-09
Add support for fwupd


AlonsoEnrique:
Example: 
![h3wxye1](https://user-images.githubusercontent.com/1877511/31346868-22bb91d8-ace0-11e7-95e6-7fdcdc035749.png)

https://fwupd.org/

cassidyjames:
@fabianthoma are you still planning on tackling this?

danrabbit:
Gonna punt this feature since we're already end of April

cassidyjames:
Thinking about this again design-wise and consent-wise. I don't know that we want to put anything in onboarding (it's not relevant to the whole OS, just AppCenter). Instead, would an infobar in the Installed pane be okay? Something like:

> AppCenter can use a secure web service to update device and component firmware [Privacy Policy](https://fwupd.org/privacy) `Enable`

cassidyjames:
I guess something we need to consider: dismissing/declining to use fwupd. :stuck_out_tongue:

mmstick:
For what it's worth, the plan for Pop!_OS is to integrate fwupd support via a new firmware panel in GNOME Settings > Devices.

hughsie:
If it helps, Andrew is a Dell intern who's building a standalone power-user firmware install application: https://gitlab.gnome.org/hughsie/gnome-firmware-updater

jackpot51:
To add to @mmstick, I don't think our plan for fwupd in Pop!_OS conflicts with the plan to add fwupd to AppCenter. We would just not enable fwupd in AppCenter and have our GNOME Settings panel handle it.

jackpot51:
@hughsie I think elementary will need a Vala wrapper for libfwupd to accomplish this - or they would use the org.freedesktop.fwupd DBUS API directly.

hughsie:
I think in GUI apps using the D-Bus interface async is the correct thing to use; libfwupd is really a very thin wrapper around that. If Vala bindings for any of the convenience API (like for instance https://github.com/hughsie/fwupd/pull/1232 ) then we can easily add that.

danrabbit:
@cassidyjames I think opt-in makes sense here.

I wonder if we should make it more clear in the messaging that the service is "3rd party" or "non-curated"? Would it be inaccurate to say something to the effect of "manufacturer provided firmware updates"? I think that what we want to inform the user of here is probably two things:

1. There is some data collection that happens when you opt in here
2. These updates are provided by some other entity than elementary.

[...]

PR #672, opened 2018-04-02
WIP: Abstract package and Backend to support multiple sources

fabianthoma:
To support more Backends for Appcenter in the future and also to implement stuff like #498 we need to overthink how we built Appcenter.
I'm doing quite a bit of rearchitecting of how Appcenter Works here. Basically this does:

- [ ] Rip out all of Backend specific code from the UI side of things (some is left to do)
- [ ] Define interfaces for Packages and Backends (kind of done, might require more for frontend)
- [ ] Change the Packagekit implementation so it uses the new Backend and Package structure (underway)

I know this is a huge change and also a huge diff, but keep in mind that some stuff was moved to other files.
Keep in mind that this work is not done and not currently mergeable (in fact it does not even build right now).


danrabbit:
I'm gonna assume this is in progress since there's a lot of commented stuff, mixed indentation, and it doesn't build :p

danrabbit:
Gonna close this since it pretty severely conflicts with master

PR #686, opened 2018-04-16
Improving the view of

alainm23:
Only a small visual improvement

appcenter


tintou:
Asking to @elementary/ux to review it.
Code-wise I would prefer you to use https://valadoc.org/glib-2.0/GLib.Markup.printf_escaped.html to ensure that translations don't interferes with the markup

danrabbit:
I agree with Cassidy. I'm going to close this PR because we want to make sure we're using consistent heading styles

PR #692, opened 2018-04-22
Add hover effects for homepage cards

danrabbit:
This adds a bit of a hover effect for the big homepage banner and the category cards. It also uses new CSS properties to have categories be 16:9 instead of filling space.

Also DRY our box shadows a bit

Might just be me, but I'm noticing a bit of a stutter hovering some category cards.


cassidyjames:
The margins don't animate cleanly; it looks like GTK is haing trouble repositioning the background and foreground elements with the margin animation. And honestly, I like the subtler look with just the shadow animating, but the physical box size staying the same :smile:

PR #694, opened 2018-04-25
Drop the "The" from fallback author name

cassidyjames:
Simple change, but it shortens up the fallback without removing clarity.


isantop:
For every case where this makes it more awkward, there at least 100 where it's basically the same or actively better. :+1:

PR #698, opened 2018-04-27
Proprietary License support

cassidyjames:
~Initial terrible~ `LicenseRef-proprietary=url` support

As reported at https://github.com/pop-os/shop/issues/72 and being discussed at https://github.com/ximion/appstream/issues/107

~Please send help :stuck_out_tongue_winking_eye:~


src/Views/AppInfoView.vala
```diff
@@ -252,30 +252,45 @@ namespace AppCenter.Views {
...
-                        license_url = license_url + "mit";
-                        break;
+                string? license_copy;
+                string? license_url;
+                if (project_license.has_prefix ("LicenseRef")) {
+                    // i.e. `LicenseRef-proprietary=https://example.com`
+                    var split_license = project_license.split_set ("=", 2);
+
+                    var license_type = split_license[0].split_set ("-", 2)[1];
+                    var pretty_license_type = license_type.substring (0, 1).up () + license_type.substring (1);
```
cassidyjames: This feels the most dirty and language-specific; it uppercases the `proprietary` string, or whatever comes after the `LicenseRef-` and before the `=`. In practice, this works fine in Pop!_Shop in English, but it's not translatable and might be trying to be too clever.

src/Views/AppInfoView.vala
```diff
@@ -583,10 +598,9 @@ namespace AppCenter.Views {
             }
         }
 
-        class UrlButton : Gtk.Button {
-            public UrlButton (string label, string uri, string icon_name) {
+        class UrlButton : Gtk.Grid {
```
cassidyjames: This changes the `UrlButton` to a grid with an optional URI so that we can reuse it whether or not a URL is set. It also feels like an abuse, and I forget off the top of my head why it was done.

cassidyjames:
This is going to become more important with #971, as Flatpaks do often have this format for proprietary licenses. An example (of many) is VS Code.

cassidyjames:
@donadigo and I chatted briefly. We'd really prefer the expected strings here be in the AppStream spec, as the linked issue mentions. For now, however, we can [echo what GNOME Software does](https://gitlab.gnome.org/GNOME/gnome-software/blob/master/src/gs-details-page.c#L2402) and ship the strings in AppCenter along with their translated strings.

src/Views/AppInfoView.vala
```diff
@@ -266,30 +266,67 @@ namespace AppCenter.Views {
...
-                        break;
-                    case "LGPL-2.1":
-                    case "LGPL-2.1+":
-                        license_url = license_url + "lgpl-2.1";
-                        break;
-                    case "MIT":
-                        license_url = license_url + "mit";
-                        break;
+                string? license_copy;
+                string? license_url;
```
jeremypw: There is a possibility that this is not assigned before use.  Should assign explicitly to null.

src/Views/AppInfoView.vala
```diff
@@ -266,30 +266,67 @@ namespace AppCenter.Views {
...
+                        case "LGPL-2.1+":
+                            license_url = license_url + "lgpl-2.1";
+                            break;
+                        case "MIT":
+                            license_url = license_url + "mit";
+                            break;
+                    }
                 }
-                var license_button = new UrlButton (_(project_license), license_url, "text-x-copying-symbolic");
+                var license_button = new UrlButton (_(license_copy), license_url, "text-x-copying-symbolic");
```
jeremypw: `license_url` is type `string?` but the parameter is of type `string`.  These should be matched.

src/Views/AppInfoView.vala
```diff
@@ -266,30 +266,67 @@ namespace AppCenter.Views {
...
+                    }
+                } else {
+                    license_copy = project_license;
+                    license_url = "https://choosealicense.com/licenses/";
+
+                    switch (project_license) {
+                        case "Apache-2.0":
+                            license_url = license_url + "apache-2.0";
+                            break;
+                        case "GPL-2":
```
jeremypw: Include "GPLv2"?

src/Views/AppInfoView.vala
```diff
@@ -266,30 +266,67 @@ namespace AppCenter.Views {
...
+                        case "GPL-2.0":
+                        case "GPL-2.0+":
+                            license_url = license_url + "gpl-2.0";
+                            break;
+                        case "GPL-3":
```
jeremypw: Include GPLv3?

src/Views/AppInfoView.vala
```diff
@@ -266,30 +266,67 @@ namespace AppCenter.Views {
...
+                    if (split_license[1] != null) {
+                        license_url = split_license[1];
+                    }
+
+                    string license_type = split_license[0].split_set ("-", 2)[1].down ();
+                    switch (license_type) {
+                        case "public-domain":
+                            // TRANSLATORS: See the Wikipedia page
+                            license_copy = _("Public Domain");
+                            if (license_url != null) {
```
jeremypw: Do you mean to overwrite the url given in the project license?

src/Views/AppInfoView.vala
cassidyjames: This code is unchanged from before, and not directly related to this PR.

src/Views/AppInfoView.vala
cassidyjames: This code is unchanged from before, and not directly related to this PR.

src/Views/AppInfoView.vala
cassidyjames: Oops, I think this was intended to be `== null`

PR #720, opened 2018-06-04
Delay daemon autostart by 60 seconds

codygarver:


codygarver:
This prevents AppCenter from locking up apt as soon as you log in, which is really annoying when you try to use a Terminal first thing and have to sit around waiting on this to finish.

codygarver:
Also nice if you immediately open a web browser so your network can catch its breath first.

codygarver:
(Partially?) closed #721

codygarver:
Merged.

PR #722, opened 2018-06-05
Search by mimetype

donadigo:
Needed for https://github.com/elementary/files/issues/446.

When searching also searches for mimetypes so that e.g Files can invoke AppCenter to search by mimetype.

I've made up a little "type" URI scheme to distinguish the type and appstream URI's, example:
`io.elementary.appcenter type://application/pdf`

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -98,6 +98,11 @@ public class AppCenter.App : Gtk.Application {
             return;
         }
 
+        if (file.has_uri_scheme ("type")) {
+            string? mimetype = mimetype_from_file (file);
+            main_window.search (mimetype);
+        }
+
         if (!file.has_uri_scheme ("appstream")) {
             return;
         }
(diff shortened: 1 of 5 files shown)


src/Application.vala
```diff
@@ -307,6 +318,16 @@ public class AppCenter.App : Gtk.Application {
...
     }
+
+    private static string? mimetype_from_file (File file) {
+        string uri = file.get_uri ();
+        string[] tokens = uri.split (Path.DIR_SEPARATOR_S);
+        if (tokens.length < 2) {
+            return null;
+        }
+
+        return "%s/%s".printf (tokens[tokens.length - 2], tokens[tokens.length - 1]);
```
jeremypw: You should just be able to use negative indices here i.e. tokens[-2], tokens[-1].

src/Application.vala
```diff
@@ -98,6 +98,17 @@ public class AppCenter.App : Gtk.Application {
             return;
         }
 
+        if (file.has_uri_scheme ("type")) {
+            string? mimetype = mimetype_from_file (file);
+            if (mimetype != null) {
+                main_window.search (mimetype);
```
jeremypw: This performs a general search for the mimetype string, not specifically for an app that takes that mimetype as a parameter.  It may therefore throw up irrelevant results.

src/Application.vala
donadigo: This doesn't work for me, after I change it to use negative indices, the process segfaults.

donadigo:
@jeremypw issues addressed.

src/Application.vala
jeremypw: Yes, sorry - I was thinking of string slicing indices.  It doesn't work in Vala as an actual index.

PR #741, opened 2018-07-15
Update translation files

ryonakano:
The new strings which were created in #738 couldn't translate in Weblate because the po files remained old. I updated them.


ryonakano:
@tintou Thank you!

PR #785, opened 2018-08-29
Clear Packages.installed_packages on successful remove

### LINKED ISSUE #783: An uninstalled app cannot be reinstalled
- Commit: `19789e`

I've noticed that if you uninstall an application, then attempt to reinstall it, the "Free" button disappears and nothing happens.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62762749-an-uninstalled-app-cannot-be-reinstalled?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mmstick:
Unsure if this is the fix you want, but it seems to solve the issue. Packages are added to the `installed_packages` set, but never removed. This will clear the set after a successful remove operation.

Closes #783


cassidyjames:
This appears to function as expected. @davidmhewitt does this look alright to you?

cassidyjames:
@davidmhewitt I'll add that as a separate issue since this solves the current issue.

PR #788, opened 2018-08-29
Use threads to prevent UI from blocking on Pk operations

### LINKED ISSUE #787: UI blocks on opening an app info view while updating repositories
There are two issues:

1. If the system does not contain any apt lists (located in `/var/lib/apt/lists/`), the app center seems to not request for an apt update to fech them.
2. If the system requires to fetch updated dist files from repos, the UI will freeze when opening an app view, which can easily last many minutes on a fast connection after a fresh install that requires updating everything.

I think this may be a source of frustration for a lot of people on fresh installations. I've been told that someone just installed Ubuntu and was happier with GNOME Software because of their first experience with AppCenter being that it freezes.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62765526-ui-blocks-on-opening-an-app-info-view-while-updating-repositories?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mmstick:
Fixes #787 

First, there is an issue where attaining the download size of a package would cause the UI to block when opening that package's app info view. That's been resolved with a thread which will calculate the size in the background and signal readiness via an atomic integer, and a timeout signal that will wait to set the size label as soon as it's available.

Then there's the issue of performing an install / remove / update operation which would also block after clicking the action button to perform that operation. This will now be carried out in a background thread, using a mutex to ensure that other threads wait their turn for the package manager, and waiting on a atomic result for each action in another timeout signal.

As a result, when you start the app center now, you can click on multiple apps to queue them for install, and they will install one after another as they're ready.

A future consideration would be for Vala to support channels and passing Arc'd atomic integers through them. That would enable a single background thread to be used to process all events, rather than spawning a new thread for each action and requiring a mutex to hold the extra threads back.

This should be tested with #785 merged.


tintou:
Just my $0.2, you can use https://valadoc.org/glib-2.0/GLib.AtomicInt.html to safely change an integer from multiple threads

davidmhewitt:
@tintou AtomicInt is already used in this PR... although it looks to be not necessary in this case.

mmstick:
@tintou AtomicInt is a bit weird. It'd be safer if it was its own type rather than a namespace, and more evident that an integer is being used as an atomic integer.

@davidmhewitt I've merged your changes, though I don't quite understand how this works for Vala.

I'll change the button label soon.

davidmhewitt:
@mmstick A bit of an explanation about my changes, in case any of this is useful to you in the future:

In theory, the code should have been fine in the state it was in, as true async methods that have been written properly should not block the main thread. Unfortunately, it seems there must be some blocking code in PackageKit. So, the `yield` statements that were there before should have returned immediately and allowed the main thread to continue. Then when the callback is fired, code execution continues from that point.

I made two changes to your code, one was to remove the `AtomicInt` signalling and use a `yield` and callback instead. This allows removal of the periodic wakeups to check for readiness, and instead continues on the callback being fired on the main thread by `Idle.add`. It's still a form of inter thread communication, just a bit nicer to read and possibly slightly more efficient.

The second was to add a `MainLoop` to the inner thread of the update/install action handling. In the previous revision, your mutexes would have been ineffective as the method calls before the unlock were async and would have been allowing the mutex to unlock too early. Ideally, we'd use synchronous versions of the same methods there, but we don't have them, so we block the thread ourselves, preventing the mutex from being unlocked before the install/update is done.

The vala code translates to C code something a little like the example on this page:
https://developer.gnome.org/programming-guidelines/stable/threading.html.en

I hope that helps and wasn't too simplistic, but if you have any more questions, feel free to ask!

Thanks for the PR though, you did a great job of tracking down where the slow points were.

mmstick:
Okay, the queued label has been removed, and it will just show the progress bar with the cancel button instead.

mmstick:
@davidmhewitt It's done.

davidmhewitt:
I've just added some cancellable stuff to make sure we can properly cancel queued package operations and properly reflect the UI state as well. But I think this should be good to go.

I'll give it another quick check over in a little while and some more testing. Then we'll need to find someone else to do second opinion code review since I've pushed to this a couple times now.

mmstick:
I'm going to go ahead and just merge this downstream. The sooner we get the fix, the better.

mmstick:
@donadigo Technically, they are named.

```vala
ThreadFunc<bool> run = () {};
```

Though I don't think it makes sense to clutter your namespace with single-use functions. Functions should generally only be created for code that may have purpose in being reused elsewhere. Anonymous functions are rarely ever that. Doing otherwise can make it harder to debug an issue when you need to chase down function after function to before you can finally get to the root cause.

brs17:
Testing this patch downstream in the Pop!_Shop, I found the Pop!_Shop would freeze when attempting to install two applications "at once" and while the installations were going on, trying to select on one of the already installed codecs. I provided logs and whatnot there: https://github.com/pop-os/shop/pull/91

PR #796, opened 2018-09-04
Fix homepage banner

### LINKED ISSUE #795: Banner only shows some apps
I think it's only showing apps that are using the deprecated `com.github.foo.bar.desktop` ID format.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63004394-banner-only-shows-some-apps?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #795. Right now it just stops appending `.desktop` when looking for IDs, which fixes it for new apps but doesn't display any apps with the legacy format (like Clairvoyant and Taxi).

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -132,6 +132,9 @@ namespace AppCenter {
                         }
 
                         var candidate_package = AppCenterCore.Client.get_default ().get_package_for_component_id (package);
+                        if (candidate_package == null) {
+                            candidate_package = AppCenterCore.Client.get_default ().get_package_for_component_id (package + ".desktop");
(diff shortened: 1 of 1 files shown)


btkostner:
This works, but you should also remove the lines for the other homepage houston parts, as they also don't have `.desktop` in them anymore.

PR #798, opened 2018-09-08
Add agent-style icon to Stripe Dialog

cassidyjames:
Little UI love in preparation for saving payment details

screenshot from 2018-09-07 22 25 30


src/Dialogs/StripeDialog.vala
```diff
@@ -75,19 +75,31 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
     }
 
     construct {
+        var image = new Gtk.Image.from_icon_name ("payment-card", Gtk.IconSize.DIALOG);
+        image.valign = Gtk.Align.START;
+
+        var overlay = new Gtk.Overlay ();
```
danrabbit: I would prefer to keep all the things involving the overlay together, so you can move this all down to right above line 87. That way it's easy to see how the overlay is packed

PR #800, opened 2018-09-08
Payment methods

cassidyjames:
WIP prototype for now.

screenshot from 2018-09-08 15 48 28 2x
screenshot from 2018-09-08 15 48 59 2x

- [x] Try smaller icons
- [x] Add radio buttons
- [x] Abstract payment buttons into a widget
- [x] Figure out what to do with margins
- [ ] Figure out what to do with email address


src/Widgets/PaymentMethodButton.vala
```diff
@@ -0,0 +1,83 @@
...
+        var overlay = new Gtk.Overlay ();
+        overlay.add (radio);
+
+        if (removable) {
+            var remove_button = new Gtk.Button.from_icon_name ("edit-delete-symbolic");
+            remove_button.halign = Gtk.Align.END;
+            remove_button.margin_end = 6;
+            remove_button.tooltip_text = _("Remove");
+            remove_button.valign = Gtk.Align.CENTER;
+            remove_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
```
danrabbit: Probably should cache this style context. So `var foo = bar.get_style_context ()`

PR #805, opened 2018-09-13
Make Package-Finding Operations Async

mmstick:
Fixes a UI hang when opening the app info views of certain types of packages, such as installed GStreamer codecs.


src/Core/Package.vala
```diff
@@ -367,19 +367,28 @@ public class AppCenterCore.Package : Object {
...
+            Idle.add((owned)callback);
         }
 
-        description = component.get_description ();
-        if (description == null) {
-            var package = find_package ();
-            if (package != null) {
-                description = package.description;
+        new Thread<bool> ("appstream_description", () => {
+            description = component.get_description ();
```
donadigo: I think you can ommit getting the description from `AppStream.Component` in a thread, since it's just a property get while `find_package` is the call that actually blocks the UI.

mmstick: I'll try that out.

mmstick: Finished. Commit coming soon.

donadigo: What I meant is more something like this: 
```
description = component.get_description ();
if (description != null) {
    return description;
}

new Thread<bool> (... () => { ... });
```

mmstick: Will submit a change for that soon.

mmstick: Done.

src/Core/Package.vala
```diff
@@ -349,7 +349,21 @@ public class AppCenterCore.Package : Object {
...
             change_information.cancel ();
         }
-     }
+    }
+    
+    private async Pk.Package? find_package_async () {
+        SourceFunc callback = find_package_async.callback;
+
+        Pk.Package? package = null;
+        new Thread<bool> ("appstream_description", () => {
```
donadigo: Perhaps a "appstream-find-package" would better fit the function here?

PR #812, opened 2018-09-20
Implement Screenshot Cache

### LINKED ISSUE #627: Locally cache viewed screenshots
Instead of re-downloading screenshots every time, we should be more conscious about bandwidth and load time and cache _viewed_ screenshots (not all screenshots in AppStream, because that would be huge) to disk once they've been downloaded. We should be able to check cache headers to see if we need to re-download once they've been downloaded once.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55668759-locally-cache-viewed-screenshots?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mmstick:
- Screenshots are now stored in the user's cache directory ($HOME/.cache/appcenter/screenshots/)
- Their names are chosen as the hash of the originating URL, and the extension.
- When fetched, the `Last-Modified` header is set as the mtime of each file, if found.
- The mtime of each file is compared against the `Last-Modified` header of a HEAD request, to determine when to renew a file.
- The cache is defined to not exceed 100MB. If exceeded, the files with the oldest ctime will be removed, until the cache is under the limit
- Screenshots are fetched in parallel

Fixes #627


src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,197 @@
+public class AppCenterCore.ScreenshotCache {
+    private const int MAX_CACHE_SIZE = 100000000;
+
+    public string screenshot_path;
+
+    // Atomic integer for keeping track of the current screenshot usage.
+    private int screenshot_usage;
+
+    public ScreenshotCache () {
+        screenshot_path = GLib.Environment.get_user_cache_dir () + "/appcenter/screenshots";
```
codygarver: Instead of `appcenter` should use `Build.PROJECT_NAME`

cassidyjames:
At a glance this looks sane. Thanks @mmstick! I'll wait for another desktop developer to sign off, but this pleases me. :grin:

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,201 @@
...
+     * 
+     * A result indicating the success (0) will be returned as the result upon completion.
+     */
+    public async int fetch (string url, out File out_file) {
+        SourceFunc callback = fetch.callback;
+        int ext_pos = url.last_index_of (".");
+        string extension = url.slice ((long) ext_pos, (long) url.length);
+        string file_name = "%02x".printf (url.hash ());
+
+        string path = screenshot_path + "/" + file_name + extension;
```
donadigo: Use `Path.build_filename`.

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,201 @@
...
+                    stream = file.create_readwrite (FileCreateFlags.NONE);
+                }
+            } catch (Error e) {
+                debug (e.message);
+                result = -1;
+                return true;
+            }
+
+            if (download) {
+                stderr.printf ("downloading %s to %s\n", url, path);
```
donadigo: Should probably be a `debug ()` call or just remove it entirely.

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,201 @@
+public class AppCenterCore.ScreenshotCache {
+    private const int MAX_CACHE_SIZE = 100000000;
+
+    public string screenshot_path;
+
+    // Atomic integer for keeping track of the current screenshot usage.
+    private int screenshot_usage;
+
+    public ScreenshotCache () {
+        screenshot_path = GLib.Environment.get_user_cache_dir ()
```
donadigo: Use `Path.build_filename`.

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,201 @@
...
+            + "/"
+            + Build.PROJECT_NAME
+            + "/screenshots";
+
+        if (GLib.DirUtils.create_with_parents (screenshot_path, 0755) == -1) {
+            critical (
+                "Error creating the temporary folder: GFileError #%d",
+                GLib.FileUtils.error_from_errno (GLib.errno)
+            );
```
donadigo: Should probably return at that point?

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,201 @@
...
+        new Thread<bool> ("screenshot_cache", () => {
+            cache_removal_event_loop ();
+            return true;
+        });
+    }
+
+    // When `screenshot_usage` exceeds `MAX_CACHE_SIZE`, the oldest files (based on ctime) will be deleted.
+    private void cache_removal_event_loop () {
+        screenshot_usage = summarize_screenshot_usage ();
+        while (true) {
```
donadigo: Do we really have to run always here? Can we do cleanups only when actually necessary, when downloading new screenshots?

src/Core/Client.vala
```diff
@@ -40,6 +40,7 @@ public class AppCenterCore.Client : Object {
 
     public bool updating_cache { public get; private set; default = false; }
 
+    public AppCenterCore.ScreenshotCache screenshot_cache;
```
donadigo: I guess you should use { get; construct; } here.

src/Core/ScreenshotCache.vala
mmstick: It does only perform a cleanup when necessary, communicated via an atomic integer that's incremented from the threads that fetch new screenshots. You could combine them into an extra step that re-calculates the cache size after fetching screenshots for an app. Not sure how much longer that would make the fetching process take.

src/Core/ScreenshotCache.vala
mmstick: I haven't figured out how to get `debug ()` to print anything. Is there something special that needs to be done when compiling it?

cassidyjames: @mmstick when running, you can set the environment variable `G_MESSAGE_DEBUG` to the debug domain, or just set it to `all`. So `G_MESSAGE_DEBUG=all io.elementary.appcenter` in this case.

mmstick: @cassidyjames For whatever reason, neither `env G_MESSAGE_DEBUG=debug ./src/io.elementary.appcenter` nor `env G_MESSAGE_DEBUG=all ./src/io.elementary.appcenter` seem to work.

mmstick:
I think this is ready for review again. I've improved the way it handles errors, and set a 10 second timeout for the Soup sessions. This should prevent it from continually attempting to fetch a screenshot that doesn't exist.

mmstick:
Just wondering, but does anyone know how safe it is to share a `Soup.session` across multiple threads?

src/Core/ScreenshotCache.vala
cassidyjames: @mmstick my bad, I had a typo. It's `G_MESSAGES_DEBUG`.

cassidyjames:
@donadigo would you be able to take a look again when you get a chance?

src/Core/ScreenshotCache.vala
Philip-Scott: And I don't think we should crash here, just not use cache instead

cassidyjames: @Philip-Scott does critical crash by default? I didn't think it did.

Philip-Scott: Critical and error crash, doesn't it? If not we can always change it to warning

mmstick: I've changed it to warning, and will now return a null pointer in the event that it couldn't be retrieved.

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,273 @@
+public class AppCenterCore.ScreenshotCache {
+    private const int MAX_CACHE_SIZE = 100000000;
+
+    public string screenshot_path;
+    private Soup.Session session;
+
+    public static ScreenshotCache? new_cache () {
```
Philip-Scott: Why static instead of just making a new class? It's not really a singleton anyways :)

src/Views/AppInfoView.vala
```diff
@@ -519,8 +524,35 @@ namespace AppCenter.Views {
...
+                    int index = i;
+
+                    cache.fetch.begin (url, (obj, res) => {
+                        results[index] = cache.fetch.end (res, out file);
+                        screenshot_files[index] = file;
+                        completed++;
+                    });
+                }
+
+                cache.maintain ();
```
Philip-Scott: Maybe let's just do the maintain once when you either open or close the app instead of going though all the cache every time?

src/Views/AppInfoView.vala
```diff
@@ -519,8 +524,35 @@ namespace AppCenter.Views {
...
+
+                // TODO: dynamically load screenshots as they become available.
+                while (urls.length () != completed) {
```
Philip-Scott: I think this _could_ crash, since `completed` might be modified on two separate threads, on this one and in the cache.fetch async callback

src/Core/ScreenshotCache.vala
Philip-Scott: edit: Instead of just using `new` to make an instance of the cache? :)

mmstick: The `static` keyword allows the method to be called without requiring an instance to be constructed first. After all, this is seeking to create a new instance, not use an existing one. There's a chance that the cache can't be created, so this will avoid allocating the object if the object is going to be invalid anyway.

As for singletons in general, I'm strongly opposed to their use in software. It's just bad practice.

src/Views/AppInfoView.vala
mmstick: The time to run this is within the `~1ms` range.

src/Views/AppInfoView.vala
mmstick: I don't think `begin ()` actually creates a separate thread. The operation is purely done in a synchronous manner.

mmstick:
Any plans to move this forward?

src/Core/ScreenshotCache.vala
```diff
@@ -0,0 +1,273 @@
...
+     * 
+     * A result indicating the success (0) will be returned as the result upon completion.
+     */
+    public async int fetch (string url, out string out_file) {
+        SourceFunc callback = fetch.callback;
+        int result = 0;
+        string path = generate_screenshot_path (url);
+        var file = File.new_for_path (path);
+
+        new Thread<bool> ("fetching_screenshot", () => {
```
tintou: Any specific reason to prefer a thread here instead of using all the `async` variants of the methods there?

mmstick: Sounds like that would complicate this implementation greatly, and increases the amount of code being run in the main thread that doesn't have to be.

tintou: how so? it just means using the `yield` keyword pretty everywhere

mmstick: Though using `yield` would not have this process running purely in the background. Each callback would hog processing time from the main thread.

PR #813, opened 2018-09-20
WIP: Recover state w/ `dpkg --configure -a` when required

### LINKED ISSUE #114: update error
whene i try to do updates in appcenter, is blocks and it say's "waiting for package manager lock"

whene i try to update with terminal "sudo apt-get update" it say's
"Get:1 http://security.ubuntu.com/ubuntu xenial-security InRelease [102 kB]
Hit:2 http://be.archive.ubuntu.com/ubuntu xenial InRelease
Get:3 http://be.archive.ubuntu.com/ubuntu xenial-updates InRelease [102 kB]
Get:4 http://be.archive.ubuntu.com/ubuntu xenial-backports InRelease [102 kB]
Hit:5 http://ppa.launchpad.net/elementary-os/stable/ubuntu xenial InRelease
Hit:6 http://ppa.launchpad.net/elementary-os/os-patches/ubuntu xenial InRelease
Fetched 306 kB in 1s (159 kB/s)                    
E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem."

Launchpad Details: [#LP1670211](https://bugs.launchpad.net/bugs/1670211) BOUCHAERT Giani - 2017-03-05 23:03:43 +0000

mmstick:
I haven't yet tested this, so it would be nice if someone could get the system in a state where this would be required.

Closes #114


arshubham:
@mmstick You can test it by adding daily ppa on juno beta 1 not sure about beta 2, While updating photos it gives the error.

mmstick:
@arshubham I don't use elementary OS, so that's not an option. If you're getting the error though, it'd be nice if you could try this patch out.

davidmhewitt:
I've got to say that I don't much like this approach due it to being completely non-portable to non deb-based distributions and there are certainly a couple that package appcenter. So, at the very least this should be a conditional compile behind a meson flag.

But the fact that it relies on CLI commands also isn't brilliant. I know the output format of dpkg is super unlikely to change now, but it still feels brittle when there are official libraries to interact with apt and dpkg.

This sort of thing is the whole motivation behind PackageKit, and it looks like they already have repair system methods: https://valadoc.org/packagekit-glib2/Pk.Client.repair_system.html

From looking at the code from the PackageKit apt backend, it looks as though that method only does the equivalent of a `apt-get install --fix-broken`, which I don't think would fix this particular case as it's partially configured packages rather than broken dependencies. But I'll do some testing and have a look at submitting a PR to PackageKit as I'm sure they'd be interested in having a `repair_system` method that can repair some more stuff.

mmstick:
@davidmhewitt If you could get PackageKit to do that, that would be great, but we do need a solution today.

mmstick:
Going to add a dialog to ask the user to perform the repair, rather than doing it automatically before checking for updates. Will allow it to perform the repair automatically when upgrading / installing a package, though.

davidmhewitt:
After doing some more testing and looking at the code of PackageKit, I'm of the strong belief that it may already be able to resolve this issue and other issues like it with the `repair_system` methods.

I guess the only thing worth thinking about is the UX behind it. I think the sensible thing to do would be to assume that the system is broken in some way if an upgrade/removal/installation fails, but then do we automatically attempt a `repair_system` in that case or do we let the user choose?

There's some risk behind doing it automatically, because depending on how broken the packaging system is, there's always the risk of breaking it more by letting some algorithm try to automatically add/remove packages to try and bring it back into a sensible state. But at the same time it's quite a technical question that may confuse people. I guess the only benefit we're offering by not doing it automatically and letting the user choose is that the more technical users that understand `apt`/`dpkg` at the command line possibly have a somewhat better chance at resolving it manually and they can cancel the auto repair.

danrabbit:
@davidmhewitt Because of the potentially dangerous consequences, I think we should ask first and use a destructive action button with the word "Try" in it to make it clear that it could fail.

mmstick:
If this is being fixed upstream in packagekit, then this won't be necessary.

PR #818, opened 2018-09-24
Discard old app info views to save memory

mmstick:
When navigating back to the app list, the app view that was previously loaded will remain in memory forever. This will ensure to unload the app view when returning from an app view. This fixes one of the memory issues in the app center, though there's still some memory leaks in the UI that I haven't figured out yet.


mmstick:
@donadigo I did attempt to do that, but the issue I had was that destroying the widget in an `Idle.add ()` would sometimes cause a segfault, and I'm not sure why. Even if during the `Idle.add`, I were to fetch the child by name and check if it was null before attempting to destroy it.

donadigo:
I think this should work:
```
Timeout.add (transition_duration, () => {
    destroy ();
    return false;
});
```

mmstick:
@donadigo That's what causes a segfault.

mmstick:
I suspect it may be the `load_more_content` thread that's causing it.

donadigo:
I have experimented with this a little bit more, and the issue seems hard to track down. I commented out completely `load_more_content` and a lot more of stuff and this keeps happening.

mmstick:
@donadigo Managed to fix this with an `Idle.add ()` to the actions that follow removal of the widget in each `return_clicked ()` method. That will ensure that they're queued to occur after the removal of the widget, rather than happening before it.

donadigo:
@mmstick yeah I see, that's quite unfortunate.

jeremypw:
@donadigo: I've tweaked this using a timeout as you suggest and it seems to work.

jeremypw:
I was wondering whether a simpler solution might be to leave clearing out the stack until after the main window is closed (hidden).

mmstick:
@jeremypw The issue with waiting for the main window to be closed is that memory usage will climb for each app info view that is opened during that session.

src/Views/View.vala
```diff
@@ -18,6 +18,8 @@
  * Authored by: Corentin Noël <corentin@elementaryos.org>
  */
 
+public delegate void Fn ();
```
donadigo: `Fn` seems a little bit too ambigious. Perhaps `BeforeDestroyFn`?

mmstick: This is because it has no meaning besides serving as a generic type for any function / closure which takes no inputs and provides no outputs. It therefore may be used anywhere where that pattern makes sense. Was looking to emulate the `Fn()` trait in Rust.

jeremypw:
@mmstick But the memory loss would be limited (most users are not going to keep the window open for very long) and would still be reclaimed without having to log out/reboot.  And the codebase would be slighlty simpler and potentially more stable. However, if your method does work reliably then I am not averse.

I too would have preferred a more informative delegate name to help maintainability.  We are not writing a general purpose library here.

jeremypw:
I think it would be better not to merge this until the problems mentioned in #947 are sorted out rather than risk introducing further instability.

PR #819, opened 2018-09-24
Use dbus to reboot via GNOME, if possible

### LINKED ISSUE #766: Use org.gnome.SessionManager.EndSessionDialog
We should use this DBus interface instead of shipping our own dialog.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61917993-use-org-gnome-sessionmanager-endsessiondialog?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mmstick:
Closes #766 

I found no mention of an `EndSessionDialog`, but the `Reboot` method does precisely that, opening a shell dialog to request the user if they want to reboot or not. (also interesting that `RequestReboot` apparently doesn't exist, even though it's advertised)

```
node /org/gnome/SessionManager {
  interface org.gnome.SessionManager {
    methods:
      Setenv(in  s variable,
             in  s value);
      GetLocale(in  i category,
                out s value);
      InitializationError(in  s message,
                          in  b fatal);
      RegisterClient(in  s app_id,
                     in  s client_startup_id,
                     out o client_id);
      UnregisterClient(in  o client_id);
      Inhibit(in  s app_id,
              in  u toplevel_xid,
              in  s reason,
              in  u flags,
              out u inhibit_cookie);
      Uninhibit(in  u inhibit_cookie);
      IsInhibited(in  u flags,
                  out b is_inhibited);
      GetClients(out ao clients);
      GetInhibitors(out ao inhibitors);
      IsAutostartConditionHandled(in  s condition,
                                  out b handled);
      Shutdown();
      Reboot();
      CanShutdown(out b is_available);
      Logout(in  u mode);
      IsSessionRunning(out b running);
      RequestShutdown();
      RequestReboot();
    signals:
      ClientAdded(o id);
      ClientRemoved(o id);
      InhibitorAdded(o id);
      InhibitorRemoved(o id);
      SessionRunning();
      SessionOver();
    properties:
      readonly s SessionName = 'pop';
      readonly s Renderer = 'GeForce GTX 1070/PCIe/SSE2';
      readonly b SessionIsActive = true;
      readonly u InhibitedActions = 0;
  };
}
```


mmstick:
@donadigo It's a singleton now.

cassidyjames:
UX-wise this seems to work here in elementary OS. Hitting Restart in the infobar throws the shut down dialog. I wonder if we should rephrase anything since the dialog has more options than just restarting, or is it fine? cc @elementary/ux

src/Views/AppListUpdateView.vala
```diff
@@ -54,7 +54,7 @@ namespace AppCenter.Views {
             update_mutex = GLib.Mutex ();
             apps_to_update = new Gee.LinkedList<AppCenterCore.Package> ();
 
-            sc = new SuspendControl ();
+            sc = SuspendControl.get_default ();
```
donadigo: It isn't needed anymore to hold a reference to `SuspendControl`, just use `SuspendControl.get_default ()` in all places where that's needed.

src/SuspendControl.vala
```diff
@@ -41,6 +44,14 @@ public class SuspendControl {
         }
     }
 
+    public static SuspendControl get_default () {
```
donadigo: Could return `unowned SuspendControl`.

PR #820, opened 2018-09-24
Fix blacklist path

donadigo:
The CONFIGDIR meson path definition doesn't contain the slash at the beggining hence it currently fails to read the blacklist file. This fixes it.


donadigo:
@codygarver fresh Juno beta2 prints me: `WARNING **: 23:40:20.253: ComponentValidator.vala:38: Could not get the contents of blacklist file: Failed to open file “etc/io.elementary.appcenter/appcenter.blacklist”: No such file or directory` and it indeed doesn't get applied.

PR #826, opened 2018-09-27
Make Homepage's reference to MainWindow weak

davidmhewitt:
While this doesn't seem to be the source of some of the bigger memory issues, it definitely doesn't help having both classes hold strong references to each other.


cassidyjames:
I haven't thoroughly analyzed this or anything, but I've installed it locally and haven't noticed a difference for the worse.

tintou:
It seems that the variable is only stored to call `main_window.homepage_loaded ();`
It should be better to have a `page_loaded` signal that the homepage connects to, and thus breaking the circular reference

davidmhewitt:
Superseded by #848

Issue #836, opened 2018-10-04
Install / cancel button moves


peteruithoven:
After clicking the Install button, when it becomes the cancel button it moves down. It's probably a small layout issue caused by the new size text? 
![peek 2018-10-05 00-11](https://user-images.githubusercontent.com/523210/46506059-5af9ae80-c833-11e8-8a6a-c959b424e1d7.gif)

PR #841, opened 2018-10-05
Fix action button movement

peteruithoven:
Prevents pay / install button to move after click. Fixes: #836


src/Views/AppInfoView.vala
```diff
@@ -227,20 +227,22 @@ namespace AppCenter.Views {
             if (!package.is_local) {
                 var download_button_grid = new Gtk.Grid ();
                 download_button_grid.orientation = Gtk.Orientation.VERTICAL;
-                download_button_grid.valign = Gtk.Align.END;
+                download_button_grid.valign = Gtk.Align.START;
                 download_button_grid.halign = Gtk.Align.END;
                 download_button_grid.row_spacing = 3;
+                download_button_grid.margin_top = 40;
```
danrabbit: I'm not a huge fan of magic numbers. Where is this value from? Can we instead do something like get the allocation of a widget we're trying to align to or use a size group or something?

peteruithoven: I understand. I needed to Start align the Download grid so it wouldn't move because of the disappearance of the size label. I then needed to drop the grid down to visually align it with the Package name and version. See:
screenshot from 2018-10-11 13 45 25-line

If there is a better way I'd love to try.

jeremypw:
I have pushed an alternative fix (#862)

PR #842, opened 2018-10-06
Replace deprecated Component.get_desktop_id

### LINKED ISSUE #683: AppStream.Component.get_desktop_id is deprecated
Replaced by Launchable. See https://valadoc.org/appstream/AppStream.Component.get_desktop_id.html

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56974042-appstream-component-get_desktop_id-is-deprecated?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #683 

And clean up a little bit of formatting while we're here.

In some cases we still want the `Component.id` field (i.e. the bit that's in the `<id></id>` tags in the appstream file. We use that for matching appstream components uniquely and sending to Houston, but for code that's involved in the launching of appstream applications, we should check if there are any `<launchable>` entries first and use those to construct the desktop app info.

I've left in a fallback to trying to launch based on the `id` field as I don't think all apps will have the `launchable` for the forseeable future.

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -597,7 +597,21 @@ public class AppCenterCore.Package : Object {
             return app_info != null;
         }
 
-        string? desktop_id = component.get_desktop_id ();
+        var launchables = component.get_launchable (AppStream.LaunchableKind.DESKTOP_ID).get_entries ();
+        for (int i = 0; i < launchables.length; i++) {
+            app_info = new DesktopAppInfo (launchables[i]);
+            if (app_info != null) {
+                break;
+            }
+        }
+
+        if (app_info != null) {
+            app_info_retrieved = true;
+            return true;
+        }
+
+        // Fallback to trying Appstream ID as desktop ID for applications that haven't updated to the newest spec yet
+        string? desktop_id = component.id;
         if (desktop_id != null) {
             app_info = new DesktopAppInfo (desktop_id);
         }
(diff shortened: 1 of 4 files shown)


src/Core/Package.vala
```diff
@@ -597,7 +597,21 @@ public class AppCenterCore.Package : Object {
             return app_info != null;
         }
 
-        string? desktop_id = component.get_desktop_id ();
+        var launchables = component.get_launchable (AppStream.LaunchableKind.DESKTOP_ID).get_entries ();
+        for (int i = 0; i < launchables.length; i++) {
+            app_info = new DesktopAppInfo (launchables[i]);
+            if (app_info != null) {
```
vjr: Won't `app_info` always be non-null here, causing the break in the very first iteration of this loop?

davidmhewitt: Oops, yes! I need to do a lookup to see if that desktop file exists and break if it does. Thank you for spotting that.

davidmhewitt: Actually, this works exactly as intended.

It's a bit weird but the `DesktopAppInfo` constructor does indeed return `null` if the desktop file isn't found: https://valadoc.org/gio-unix-2.0/GLib.DesktopAppInfo.DesktopAppInfo.html.

I'll leave a comment in the code because it does look strange.

Issue #838, opened 2018-10-04
After 1 uninstall missing padding in uninstall button 


peteruithoven:
Reproduce:
1. Install app
2. Uninstall app
3. Install app

![screenshot from 2018-10-05 00 30 07](https://user-images.githubusercontent.com/523210/46506840-f8ee7880-c835-11e8-921b-096b92ba882d.png)

hannesschulze:
I think this issue isn't related to the padding of the button, but rather to the size group. Maybe the first time displaying the buttons, both of them use the size of the "Uninstall"-button. When displaying the buttons a second time, they just use the size of the "Open"-button. I think this is related to the SizeGroup that is being used to make these two buttons the same width.

![screenshot von 2018-10-11 22 03 51](https://user-images.githubusercontent.com/43932053/46830873-1aa1af80-cda2-11e8-902f-6332f82f70f0.png)

peteruithoven:
@hannesschulze I think you're right! PR coming up...

Issue #837, opened 2018-10-04
Install button moved and cropped after failure?


peteruithoven:
The pay / install button is moved and cropped to the right after an installation failed. 
![screenshot from 2018-10-04 23 41 31](https://user-images.githubusercontent.com/523210/46506546-f3446300-c834-11e8-9b17-282b6ae80219.png)

One easy way to reproduce it is to 
1. Start a new installation in Terminal `sudo apt install ...`
2. Not enter `Y` yet
3. Attempt to install the app

This will fail, a popup will be shown and from then on the pay / install button is moved and cropped. 

Pressing the Search (back) button will even show a moved / cropped install button in that view:
![screenshot from 2018-10-05 00 22 22](https://user-images.githubusercontent.com/523210/46506661-4b7b6500-c835-11e8-984b-3e9cd04e74a2.png)

The moving and cropping was also noticed in: #833

jeremypw:
I am not sure this is the right solution.  It results in buttons of differing sizes - how different depends on the locale.  Need UX input on this.  Need a way of fixing all the button sizes to the widest one (whether visible or not)

PR #856, opened 2018-10-12
Remove dysfunctional SizeGroup

peteruithoven:
Fixes: #838 and #837


jeremypw:
I am not sure this is the right solution. It results in buttons of differing sizes - how different depends on the locale. Need UX input on this. Need a way of fixing all the button sizes to the widest one (whether visible or not).

SizeGroups are known not work properly when some members are hidden with the 'visible' property (the 'ignore-hidden' property is deprecated) but they are supposed to work when widgets are hidden with a GtkStack.

peteruithoven:
#886 is probably the better solution, closing this one.

Issue #792, opened 2018-08-31
AppCenter translation is not showing in Portuguese (Brazil)


darthmaulbr:
AppCenter translation is not showing in Portuguese (Brazil).
It's fully translated at Weblate.

PR #857, opened 2018-10-14
Add Portuguese and Brazilian Portuguese to LINGUAS

leonardo-lemos:
This branch adds pt and pt_BR to LINGUAS and make those two languages work in AppCenter.


ryonakano:
Fixes #792

PR #865, opened 2018-10-17
Refresh AppStream pool and UI when package cache changes

### LINKED ISSUE #1: AppCenter is empty if apt update has never been run [$30]
Steps to reproduce:
1. Do a fresh install without connecting to the Internet
2. Fresh boot and connection to the internet
3. Open AppCenter

Expectation:
AppCenter fires up packagekitd and updates the software catalog transparently in the background when the session starts up

Reality:
AppCenter only shows data for the pre-installed packages. packagekitd never fixes it. AppCenter is useless until you open a Terminal and manually run apt update

Launchpad Details: [#LP1615270](https://bugs.launchpad.net/bugs/1615270) Daniel Fore - 2016-08-20 19:10:38 +0000

<bountysource-plugin>

---
There is a **[$30 open bounty](https://www.bountysource.com/issues/43123361-appcenter-is-empty-if-apt-update-has-never-been-run?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #1 and probably some other stuff

So there are a few parts to this:

- We trigger a reload of the AppStream cache after every PackageKit refresh, the assumption being that PackageKit probably downloaded a new package cache which probably contains new/updated AppStream metadata. This may also fix issues where the wrong version is displayed on an app info page. And will also likely allow new packages to drop into AppCenter without restarting it.
- We drop any cached list views for each category so that they're recreated with the new metadata when that category is next opened.
- If the trending or recently updated apps banners weren't visible before we refreshed the appstream cache, we try to reload them for the case where the AppCenter was empty on the first boot.

It's possible to reproduce the empty AppCenter issue by doing the following and then restarting appcenter:
```
sudo rm /var/lib/apt/lists/*dep11*
sudo appstreamcli refresh-cache
````

In master, AppCenter will be empty until restarting it a second time. With this branch, it'll automatically populate itself some moments after starting.


jeremypw:
I'll dogfood this a while.

cassidyjames:
@davidmhewitt can you fix conflicts?

PR #877, opened 2018-10-25
Add communication category

### LINKED ISSUE #571: Add "Communication" category
Instead of dumping this all just into "Internet", we could surface a "Messaging" metacategory on the home page. This would include "Chat", "InstantMessaging", "IRCClient", "VideoConference", "Email", "Telephony", "ContactManagement".

We can use the internet-chat icon. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/53862619-add-communication-category?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #571 

Could probably use some love with the styling. I'd like to use an `internet-chat-symbolic`, but it doesn't exist. So the current icon is just a placeholder.

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -174,20 +174,15 @@
 
 .category.communication {
     background-image: linear-gradient(
-        to bottom,
-        @BANANA_100,
-        @BANANA_300
+        175deg,
+        @LIME_300 5%,
+        @LIME_500
     );
-    border:none;
-    box-shadow:
-        inset 0 0 0 1px alpha (#fff, 0.10),
-        inset 0 1px 0 0 alpha (#fff, 0.90),
-        inset 0 -1px 0 0 alpha (#fff, 0.30),
-        0 0 0 1px alpha (#000, 0.15),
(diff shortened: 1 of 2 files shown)


danrabbit:
I'm not sure if I'm digging the yellow. What about something like this?

screenshot from 2018-10-25 15 59 19 2x
screenshot from 2018-10-25 15 59 48 2x

cassidyjames:
I'm not digging the reversed gradient tbh. Feels weirdly inset when everything else is subtly outset. I was trying to avoid another blue as well, and chose yellow to riff off of Mail. But yeah I wasn't super sold on the yellow either. Would green work to echo the chat icon?

danrabbit:
Green could indeed be cool

PR #881, opened 2018-10-28
StripeDialog.vala: Match Stripe's Regex for email validation

### LINKED ISSUE #880: Unable to pay after filling out credit card details
I tried to pay for Ideogram but the pay button never came alive:

image

Checked all my details, twice, so no idea what else I can do.

FWIW, I started the payment process from this button:

image

But uninstalling and reinstalling while choosing an amount to pay results in the same problem. Then I'm not even able to download the app. 🙄

wout:
The original regex was too strict, preventing users from using anything other than lowercase alphanumeric characters (no hyphens, underscores, periods, ...).

Fixes #880

### DIFF
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -322,7 +322,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         try {
             switch (entry) {
                 case 0:
-                    var regex = new Regex ("""[a-z|0-9]+@[a-z|0-9]+\.[a-z]+""");
+                    var regex = new Regex ("""([0-9a-zA-Z]([-\.\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})""");
                     email_valid = regex.match (new_text);
                     break;
                 case 1:
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -322,7 +322,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         try {
             switch (entry) {
                 case 0:
-                    var regex = new Regex ("""([0-9a-zA-Z]([-\.\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})""");
+                    var regex = new Regex ("""([0-9a-zA-Z]([-\.\+\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})""");
                     email_valid = regex.match (new_text);
                     break;
                 case 1:
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -322,7 +322,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         try {
             switch (entry) {
                 case 0:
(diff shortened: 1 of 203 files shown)


src/Dialogs/StripeDialog.vala
```diff
@@ -322,7 +322,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         try {
             switch (entry) {
                 case 0:
-                    var regex = new Regex ("""[a-z|0-9]+@[a-z|0-9]+\.[a-z]+""");
+                    var regex = new Regex ("""([0-9a-zA-Z]([-\.\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})""");
```
Philip-Scott: Since we're modifying the regex, we should add the `+` to the regex since at least for Google's emails, you can have `foo+something@gmail.com`

```suggestion
                    var regex = new Regex ("""([0-9a-zA-Z]([-\.\+\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})""");
```

wout: Right, forgot about that one.

src/Dialogs/StripeDialog.vala
```diff
@@ -322,7 +322,7 @@ public class AppCenter.Widgets.StripeDialog : Gtk.Dialog {
         try {
             switch (entry) {
                 case 0:
-                    var regex = new Regex ("""[a-z|0-9]+@[a-z|0-9]+\.[a-z]+""");
+                    var regex = new Regex ("""([0-9a-zA-Z]([-\.\+\w]*[0-9a-zA-Z])*@([0-9a-zA-Z][-\w]*[0-9a-zA-Z]\.)+[a-zA-Z]{2,9})""");
```
Philip-Scott: Actually, we should change it for the one suggested here instead of making our own regex :sweat_smile:  https://regular-expressions.mobi/email.html?wlr=1

```suggestion
                    var regex = new Regex ("""[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?""");
```

wout: If you're going to use an elaborate version, it should take [all the scenarios](https://en.wikipedia.org/wiki/Email_address#Syntax) in to account. There are a few things wrong with this one, for example:
- forward slashes are not escaped
- it does not allow the local-part to end in capitals (e.g. `my.name.MIT@domain.com`)
- it does not allow capitals in the domain part (e.g. `my.name@Domain.com`)
- it allows an invalid local-part (e.g. `my..name@domain.com`)
- it does not allow the valid version of the one above (e.g. `"my..name"@domain.com`)
- it does not allow comments (e.g. `my.name(comment)@domain.com`)

The only question is, will Stripe accept all those characters as well? Because if that is not the case, there is no use of allowing those exotic email addresses. I'm out all day but tomorrow I'll take a look into it to be sure all scenarios are covered correctly.

wout: I just contacted the stipe developers on freenode and as it tuns out, their email validation is quite loose:

> I think we do something quite basic: no space, an @, an extension, and that's mostly it

So instead of including every possible scenario, I think it would be better to go in the opposite direction and follow the same approach as Stripe: keep it simple. So then I would suggest `[^\s]+@[^\s]+\.[^\s]+` if that's ok with you. That's exactly what their developer suggests (no space, an @, an extension).

wout:
I changed the regex to match what Stripe is using, so this should now fix the issue.

PR #885, opened 2018-11-04
a little back to get_desktop_id () for the uninstall feature

torikulhabib:
a little back to get_desktop_id () to maximize the uninstall feature of the appmenu
I think this is very important for new users who use elementary OS


cassidyjames:
Right, we replaced it because of that. @torikulhabib can you describe the issue this is supposed to solve?

PR #886, opened 2018-11-04
Do not alter buttons visibility property - use stacks to hide.

### LINKED ISSUE #837: Install button moved and cropped after failure?

### LINKED ISSUE #838: After 1 uninstall missing padding in uninstall button

jeremypw:
Fixes #838 

May also fix #837 

This is an alternative to removing the size group (see #856 ).

The visibilities of buttons and progress widget are controlled entirely by GtkStack.visible_child in AbstractAppContainer instead of setting the 'visible' property.  The code is rather cumbersome as you have to add a dummy widget which does not get drawn (an EventBox was used) as an alternative to showing the button since it is not allowed to set the visible child to 'null'.  You cannot add all the buttons to one stack because more than one may be shown at the same time.

### DIFF
--- a/src/Widgets/AppContainers/AbstractAppContainer.vala
+++ b/src/Widgets/AppContainers/AbstractAppContainer.vala
@@ -28,6 +28,14 @@ namespace AppCenter {
         }
     }
 
+    public static void set_stack_visibility (Gtk.Stack stack, bool show) {
+        if (show) {
+            stack.set_visible_child_name ("CHILD");
+        } else {
+            stack.set_visible_child_name ("NONE");
+        }
+    }
+
     public abstract class AbstractAppContainer : Gtk.Grid {
         public AppCenterCore.Package package { get; construct set; }
         protected bool show_uninstall { get; set; default = true; }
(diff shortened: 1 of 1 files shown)


peteruithoven:
Looks like this also fixes #837 :tada: 
Great you've found a better solution.

PR #887, opened 2018-11-05
Alphabetize categories

cassidyjames:
Split from #577 to make future category work more sane.


cassidyjames:
@tintou they're already alphabetized elsewhere. This is just to keep the code more sane.

PR #893, opened 2018-11-06
Update video category CSS

danrabbit:
## BEFORE
screenshot from 2018-11-05 18 13 48 2x

## AFTER
screenshot from 2018-11-05 18 14 10 2x

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -278,12 +278,14 @@
 }
 
 .category.video {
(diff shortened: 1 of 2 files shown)


danrabbit:
Good call! Made the spotlight a circle always now as well

screenshot from 2018-11-06 09 57 33 2x

PR #898, opened 2018-11-06
Clean up CSS

cassidyjames:
Lint and reorganize the CSS for consistency. It's a big diff, but there should be **no visible change** to the styles.

- [x] Lint current CSS
- [x] Alphabetize the categories for sanity
- [ ] Add linter to build process?
- [x] Pull categories CSS out into a new file?
- [x] Use constants where possible to avoid magic numbers


cassidyjames:
I've added the stylelint rules to the data folder, but haven't integrated them into the build process or anything yet. I don't think that's required to merge this, though.

cassidyjames:
Alright @danrabbit this is probably good for a review. Even though I `git mv` files, in the overall diff it's not showing up, so it may be easier to step through some of the changes by commit to see the moves.

PR #904, opened 2018-11-07
Rewrite window state saving

danrabbit:
Uses the same methods as the new Mail

* Use (ii) for window size and position keys
* Move state restoring to `construct` block
* Use GLib.Settings instead of Granite.Settings
* Move window position even when maximized so that window position is restored when unmaximized
* Save state on configure event (throttled) so that state is saved even when the app is quit forcefully
* Only save window size and position when the window is not maximized so that we can restore those properties accurately


src/MainWindow.vala
```diff
@@ -214,24 +219,35 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+                Gdk.Rectangle rect;
+                get_allocation (out rect);
+                settings.set ("window-size", "(ii)", rect.width, rect.height);
+
+                int root_x, root_y;
+                get_position (out root_x, out root_y);
+                settings.set ("window-position", "(ii)", root_x, root_y);
+            }
+
+            return false;
```
tintou: Use `GLib.Source.REMOVE` instead of `false`

src/MainWindow.vala
```diff
@@ -214,24 +219,35 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+        });
+
+        return base.configure_event (event);
```
tintou: Can you try with `return false` instead? The documentation explicitly stands that:
> true to stop other handlers from being invoked for the event. false to propagate the event further.

danrabbit: This does really fucky things when you resize the window. Decorations get clipped and other weird things

tintou: Alright, thanks for trying :)

src/MainWindow.vala
```diff
@@ -132,7 +120,24 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+        settings = new GLib.Settings ("io.elementary.appcenter.settings");
+
+        int window_x, window_y;
+        var rect = Gtk.Allocation ();
+
+        settings.get ("window-position", "(ii)", out window_x, out window_y);
+        settings.get ("window-size", "(ii)", out rect.width, out rect.height);
+
+        if (window_x != -1 ||  window_y != -1) {
+            move (window_x, window_y);
```
jeremypw: Not sure this will work with all window managers, but as long as it works with gala (which it does at the moment) I guess thats OK.

src/MainWindow.vala
```diff
@@ -132,7 +120,24 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+        }
+
+        if (settings.get_boolean ("window-maximized")) {
+            maximize ();
+        }
+
         icon_name = "system-software-install";
+        set_allocation (rect);
```
jeremypw: Should use `resize ()` - gtk documentation states that `set_allocation ()` should not be called directly.  Note that if `resize ()` is used, then you need to use `get_size ()` when saving the size parameters, not `get_allocation ()`. I think one includes the shadow and one doesnt.

danrabbit: From the previous testing we had done, using `get_size ()` and `resize ()` lead to discrepancies with the window size on restore sometimes so you'd get windows that grow or shrink every time you restore it

jeremypw: Best to find out the reason why ...

src/MainWindow.vala
```diff
@@ -42,6 +42,9 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
     private Gtk.Grid network_view;
     private Gtk.Label updates_badge;
 
+    private GLib.Settings settings;
+
+    private uint configure_id;
```
jeremypw: Maybe name `configure_timeout_id` for better clarity?

danrabbit:
Superseded by #930

PR #905, opened 2018-11-07
Revamp Office CSS

cassidyjames:
image

Trying to avoid just a blank tile here. Each color represents an office app, like word processor, spreadsheets, presentation, drawing, and database.


snowparrot:
It looks dispropotional.

cassidyjames:
btw with the document icon:

screenshot from 2018-11-08 11 30 09 2x

snowparrot:
@cassidyjames 

The position of  word "Office" looks disproptional related to the colours on the bottom. But I cant give you any explaination.

cassidyjames:
@snowparrot maybe with the colors at the bottom, we should raise the word/icon up a bit to compensate—that way it would remain centered in the whitespace?

cassidyjames:
Here it is raised up a bit:

image

micahilbery:
Probably need an appropriately sized `x-office-document-symbolic` but other than that I like it.

cassidyjames:
@micahilbery do we want to block it for that, or merge this now and make sure we have an issue filed so it just gets better once that's released?

micahilbery:
:shipit: I don't think it's necessarily worth blocking it as it look good now. It just could look even better.

PR #908, opened 2018-11-07
Education more chalkboardy

cassidyjames:
A little more dusty, plus using palette colors.

screenshot from 2018-11-07 12 36 29

### DIFF
--- a/data/styles/categories.css
+++ b/data/styles/categories.css
@@ -147,10 +147,15 @@
 
 .category.education {
     background-image:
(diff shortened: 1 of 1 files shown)


danrabbit:
I feel like this makes the text harder to read. It feels a little more glowy than just like smeared chalk.

cassidyjames:
image

cassidyjames:
With @micahilbery's new eraser marks in place of text shadow:

image

PR #909, opened 2018-11-07
 Homepage: Add background pattern to communication card

danrabbit:
- [x] Requires a release of elementary icons with [this commit](https://github.com/elementary/icons/commit/00df0e80a56c11208e9752faf22c1bd9eb667175)

screenshot from 2018-11-07 11 12 35 2x


danrabbit:
Oh derp, I forgot I never proposed that

danrabbit:
This should probably be blocked still to make sure a new release of AppCenter doesn't come out with a missing icon

PR #912, opened 2018-11-09
Add graph to Finance tile

cassidyjames:
Playing with this direction:

image

Kinda in the style of something like this:

graph


cassidyjames:
@micahilbery commented in Slack that it might feel too much like Audio. We might try GResourcing a line graph SVG instead.

cassidyjames:
Closing for now since nobody seemed that excited about the design. We can always revisit.

PR #913, opened 2018-11-09
Update Internet styling

cassidyjames:
Use palette colors plus add a cloud to the background

internet-css


micahilbery:
I wonder if it would look cool to do multiple (2-3) overlapping background clouds.

PR #637, opened 2018-03-02
Adds uninstall buttons to the app listings.

isantop:
Adds an uninstall button to apps installed in the results view as well. 

Fixes #634

PR #636, opened 2018-03-02
Show Uninstall button in Updates view

isantop:
This PR adds uninstall buttons to the Updates view. They appear to be fully functional and work great without any additional changes. 

Fixes #632

Issue #634, opened 2018-03-02
Show uninstall buttons in search listings


cassidyjames:
If you're viewing a list of apps, is it useful to show an "uninstall" button alonside any apps that are installed? 

I'm not super convinced it should be on the category listings since if you're browsing for an app, you're probably not looking to uninstall one. 

On the other hand, if you're _searching_ for an app _in order to uninstall it_, it might be really handy to bubble that action up to the list instead of having to go into the app's page itself.

Brought up due to some WIP changes in Pop!\_Shop downstream. https://github.com/pop-os/shop/pull/54/files#diff-8a3694d1b620c25ea491262722dd3d6aR64

danrabbit:
My gut reaction is that on the categories page the user is browsing, so they're not necessarily looking for apps they've already installed.

I agree though that in the case of search they could be looking for an app they have installed already

cassidyjames:
Yeah I agree. The problem is that right now those are the same view.

isantop:
Should AppCenter show apps that are already installed when browsing the categories page?

cassidyjames:
@isantop I think so, since it might feel weird if you're browsing and expect to see an app that's installed and suddenly it's not any more. Basically, it's still on the store shelves even though you bought a copy. 

But I don't think uninstalling is the **primary** action for this list, and I don't really like mixing actions (especially opposite ones) in the same list.

danrabbit:
I'm gonna punt since this is an enhancement and not a release blocker

danrabbit:
I'm gonna make a call here, we probably don't need to show uninstall buttons in list views. This isn't an action that you're likely to be doing across multiple apps in sucession, unlike "Upgrade" or "Install". Uninstall is always available from AppInfo views

Issue #632, opened 2018-03-02
Show uninstall button on updates page


cassidyjames:
Splitting out from #548. It might be useful to show an Uninstall button for apps on the updates/installed page.

danrabbit:
Should this show only for apps that don't have updates?

cassidyjames:
@danrabbit probably, as I don't think we want two buttons in the row. And since we already separate them into sections, I think it'd feel natural.

snowparrot:
Please consider in later versions that updates are sometimes only possible if you pay something or fill in 0 Dollars. If this context there is a need for a second uninstall button as some beginners might think "I just can pay but I cant uninstall".

danrabbit:
I'm gonna punt since this is an enhancement and not a release blocker

danrabbit:
I'm gonna make a call here, we probably don't need to show uninstall buttons in list views. This isn't an action that you're likely to be doing across multiple apps in sucession, unlike "Upgrade" or "Install". Uninstall is always available from AppInfo views

PR #919, opened 2018-11-13
[WIP] Tweaked visibility of uninstall button in lists

peteruithoven:
Tweaked visibility of uninstall button in lists

## Category list
Don't show in category lists.
See: https://github.com/elementary/appcenter/pull/637#pullrequestreview-100963848
screenshot from 2018-11-14 00 28 58

## Search list
Show in category lists.
See (same comment): https://github.com/elementary/appcenter/pull/637#pullrequestreview-100963848 
screenshot from 2018-11-14 00 28 46

## In installed tab
Show in Installed tab, but not when update is available.
See: https://github.com/elementary/appcenter/pull/636#pullrequestreview-100964526
Not sure why it's weird to show two buttons.
screenshot from 2018-11-14 00 29 41

Not sure what to do for installed apps that show the payment option. Should that payment option even be shown when there is no update available? 
screenshot from 2018-11-14 00 29 47

Fixes:
- #632
- #634

Supersedes:
- #636 
- #637

ToDo:
- [ ] Fix spacing when uninstall is the only button (it's positioned a bit to the left currently).


cassidyjames:
>Not sure what to do for installed apps that show the payment option. Should that payment option even be shown when there is no update available?

I think I'd be fine removing the payment buttons when apps don't have updates; I think it has muddied whether or not a payment is actively being requested, and the Uninstall action is probably more relevant anyway.

peteruithoven:
I've now removed the payment button on the Installed page, for packages for which no updates are available. So this is the only change compared to the screens shared above:
screenshot from 2019-02-08 11-36-57

Removing the space is tricky, because, the logic that renders these rows is also reused for the App info view where you do want the spacing. I don't really feel like adding more hacks to make this work. This code needs a serious refactor, less inheritance, more composition, smaller files. 

We should also find a better way to have a set of buttons with the same size that we want to show and hide. Currently we use workarounds like:
``` vala
uninstall_button = new Gtk.Button.with_label (_("Uninstall"));
uninstall_button_stack = new Gtk.Stack ();
uninstall_button_stack.add_named (uninstall_button, "CHILD");
uninstall_button_stack.add_named (new Gtk.EventBox (), "NONE");
uninstall_button_stack.hhomogeneous = false;

action_button_group = new Gtk.SizeGroup (Gtk.SizeGroupMode.HORIZONTAL);
action_button_group.add_widget (action_button);
// other buttons that should have the same size

set_stack_visibility (uninstall_button_stack, false);

public static void set_stack_visibility (Gtk.Stack stack, bool show) {
  if (show) {
    stack.set_visible_child_name ("CHILD");
  } else {
    stack.set_visible_child_name ("NONE");
  }
}
```

jeremypw:
@peteruithoven Closing as has conflicts and no recent activity. Please feel free to reopen (as draft) if you wish to continue working on it.

PR #930, opened 2018-12-22
Use resize and get_size when saving and restoring window state

jeremypw:
based on #904 but using resize and get_size instead of manipulating allocation directly (contrary to Gtk documentation).

Steps are taken to ensure the results are correct when using a persistent app like appcenter.


jeremypw:
I am pretty sure it did work when pushed?!  However it now seems that the default configure-event handler is swallowing the signal so connect_after () does not work.  I have reverted to connect () and stopped respawning the settings timout to keep resizing smooth.

src/MainWindow.vala
```diff
@@ -132,8 +120,36 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+
         set_size_request (910, 640);
+
+        if (window_x != -1 ||  window_y != -1) {
+            move (window_x, window_y);
+        }
+
+        resize (window_width, window_height);
+
+        show.connect (() => {
```
danrabbit: I don't think this is necessary. In other apps we're overriding the configure event method as in https://github.com/elementary/appcenter/pull/904/files#diff-4089e5f35cbce620c0c1758fb579b361R222

jeremypw:
Refactored configure_event handling as suggested - thanks!  Also fixed a small bug whereby restoring position did not work on first use.

Issue #939, opened 2019-01-22
Duplicated entries in Trending & Recently Updated apps


megatux:
Happened a few times to me :
![captura realizada el 2019-01-22 09 29 03](https://user-images.githubusercontent.com/850020/51560363-138b4280-1e63-11e9-9c1c-026414011a87.png)
and reported by someone else, too :
![screenshot from 2019-01-23 00-05-11](https://user-images.githubusercontent.com/850020/51560375-19812380-1e63-11e9-9aa7-eac3a49f951c.png)

cassidyjames:
I occasionally get this as well. I'm assuming somehow the API is being hit twice, or the results are cached and then retrieved again? Regardless, we should probably just de-dupe the carousels anyway.

PR #941, opened 2019-01-23
Comment out the banner refresh hack

far-echo:
This is a quick fix for #939. There was a hack to trigger a refresh of banners if they failed to load at the first time, but it didn't give them any time to load. I guess the code was rewritten to be asynchronous at some point, but the hack wasn't updated to accomodate this. Handling network issues properly will require some refactoring, as the banner code doesn't even have a method to clear the banners list, so I've just commented the hack out to remain as a reminder.

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -160,10 +160,10 @@ namespace AppCenter {
                     }
                 }
 
-                // If the banners weren't populated, try again to populate them
-                if (!recently_updated_revealer.reveal_child || !trending_revealer.reveal_child) {
-                    refresh_banners ();
-                }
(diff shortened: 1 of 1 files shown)


far-echo:
Now that you mention it, I see that the code is plain wrong: it's hardcoded to load banners exactly once, so you probably have laptop users not getting updated banners for weeks, if not months, until the Appcenter itself is updated and restarted. I vaguely remember trying to do a rewrite of this code before, so I may just beat you to it.

danrabbit:
Closing this PR since it conflicts with master. OP, if you want to fix the conflicts and then re-open this PR, please feel free to :)

PR #945, opened 2019-01-26
Move PackageKit operations to AsyncQueue/ThreadPool

### LINKED ISSUE #789: Investigate removing installed_packages TreeSet
Regarding #785, @davidmhewitt commented:

>I do question the need for this TreeSet at all. Could probably save some RAM and get rid of it. May be worth looking into who put that installed_packages code in and seeing if they can remember if it's still needed for something.
>
>Seems to do very little.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62790808-investigate-removing-installed_packages-treeset?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #790: Investigate using threading in Client.vala
I'm not going to block #788 with this as it does seem to resolve most if not all of the hanging issues. This is more of a bigger refactoring to consider for the future.

It seems although Client.vala uses the async PackageKit methods, they still block the UI thread. Rather than wrapping method calls in threads potentially every time we use them like in #788, we should fix the problem at the source and make the Client.vala methods truly async by threading them and then reverting the AbstractAppView.vala changes in #788.

This should ensure that any future uses of these methods don't cause any future slowdowns.

Sidenote: Since PackageKit in theory does its own queueing, it should be OK to just keep spawning off new threads and letting PackageKit handle it. But maybe we want to look at using a single thread for package operations and use `GLib.AsyncQueue` too for efficiency.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62791921-investigate-using-threading-in-client-vala?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #938: Updates not listed?
The updates count in the Installed badge seems off with the contents off the Installed page:
screenshot from 2019-01-21 19 28 29 1

Apt does report there are updates available:
```
$ apt list --upgradable
Listing... Done
com.github.bartzaalberg.snaptastic/bionic 0.7.0 amd64 [upgradable from: 0.6.0]
com.github.calo001.fondo/bionic 1.2.0 amd64 [upgradable from: 1.1.4]
com.github.philip-scott.spice-up/bionic 1.8.0 amd64 [upgradable from: 1.7.0]
```

Restarting AppCenter (not the background process) doesn't change anything.
Also killing that background process (`killall io.elementary.appcenter`) results in the same. 
Running `sudo appstreamcli refresh-cache --force` doesn't help.

Going to a specific app's page does show it's upgradable:
screenshot from 2019-01-21 19 49 15

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68990780-updates-not-listed?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #939: Duplicated entries in Trending & Recently Updated apps

davidmhewitt:
Essentially, I've moved the PackageKit specific methods out of `Client.vala` and put them in their own PackageKit specific class. We know from experience that even the async PackageKit methods often block the main thread, which has lead to random inline thread lambdas being littered throughout the codebase.

To resolve this issue, the new `PackageKitClient` class uses an AsyncQueue which stacks up a bunch of PackageKit jobs as requested by the various bits of AppCenter, then a built-in ThreadPool unstacks these in turn, executes the job and returns the result via async methods. As PackageKit only performs one operation at a time internally anyway, it's fine to stack up the jobs like this. I may add some concept of priority to the queue so we can prioritize certain backend tasks over others.

Moving each of these methods in turn to a new class allows me to see where there isn't very good separation between the PackageKit backend and the UI and I can slowly start to resolve these issues so it's easier to work with other backends like Flatpak in the future.

Fixes #790 
Fixes #789 
Fixes #939 
Fixes #938


davidmhewitt:
This is now nearing completion. All of the methods that operate on the PackageKit backend have now been moved to the new class and stripped of any application logic.

A couple of issues are outstanding:
- The spinner isn't completely truthful anymore as it only spins when the old class is doing any work.
- There seems to be an issue with updates not showing properly on the "Installed" tab, but I think this exists in master (see #938)

As far as I can tell, everything else is working just as well, if not better than it did before. It'd be good to get a decent amount of testing on this branch to make sure there's nothing I've broken aside from the above. It definitely feels much faster and more responsive than master.

src/Core/PackageKitJob.vala
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: David Hewitt <davidmhewitt@gmail.com>
+ */
+
+public class AppCenterCore.PackageKitJob : Object {
+    public JobArgs? args;
+    public Value result;
```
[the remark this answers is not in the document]
davidmhewitt: Making this a property seems to cause some issues. I'll leave it as is for now.

src/Core/Client.vala
```diff
@@ -292,71 +206,58 @@ public class AppCenterCore.Client : Object {
...
-        new Thread<void*> ("parse-drivers-output", () => {
-            string[] tokens = output.split ("\n");
-            for (int i = 0; i < tokens.length; i++) {
-                string package_name = tokens[i];
-                if (package_name.strip () == "") {
-                    continue;
-                }
+        string[] tokens = output.split ("\n");
+        for (int i = 0; i < tokens.length; i++) {
+            string package_name = tokens[i];
```
tintou:
```suggestion
            unowned string package_name = tokens[i];
```

src/Core/Package.vala
```diff
@@ -368,17 +371,28 @@ public class AppCenterCore.Package : Object {
     private async Pk.Exit perform_package_operation () throws GLib.Error {
         Pk.ProgressCallback cb = change_information.ProgressCallback;
         var client = AppCenterCore.Client.get_default ();
+        var pk_client = AppCenterCore.PackageKitClient.get_default ();
+
+        Gee.ArrayList<string> packages_ids = new Gee.ArrayList<string> ();
+        foreach (var pkg_name in component.get_pkgnames ()) {
```
tintou:
```suggestion
        foreach (unowned string pkg_name in component.get_pkgnames ()) {
```

src/Core/PackageKitClient.vala
```diff
@@ -0,0 +1,489 @@
...
+        });
+
+        jobs.push (job);
+        yield;
+        return job;
+    }
+
+    private void get_package_by_name_internal (PackageKitJob job) {
+        var args = (GetPackageByNameArgs)job.args;
+        var name = args.name;
```
tintou:
```suggestion
        unowned string name = args.name;
```

src/Core/PackageKitClient.vala
```diff
@@ -0,0 +1,489 @@
+/*-
+ * Copyright (c) 2019 elementary LLC. (https://elementary.io)
```
tintou:
```suggestion
 * Copyright 2019 elementary, Inc. (https://elementary.io)
```

src/Core/PackageKitJob.vala
```diff
@@ -0,0 +1,87 @@
+/*-
+ * Copyright (c) 2019 elementary LLC. (https://elementary.io)
```
tintou:
```suggestion
 * Copyright 2019 elementary, Inc. (https://elementary.io)
```

davidmhewitt:
The following things need to be tested as they have been touched by this branch:
* Installing/updating/removing packages
* Suspend prevention during updates
* Driver detection
* List of packages in installed tab
* Update checking
* Download size display
* Banner loading
* The spinner indicating activity in the headerbar

Search, navigation, UI, strings, payments, restart notifications, and the backends that fetch info about apps for the various views all remain untouched.

danrabbit:
I can confirm that the following are working as expected:
- [x] Installing/updating/removing packages
- [x] List of packages in installed tab
- [x] Update Checking 
- [x] Download size display
- [x] Banner loading
- [x] The spinner indicating activity in the headerbar

I was not able to confirm:
- [ ] Suspend prevention during updates, because I didn't have enough updates for this to be a problem
- [ ] Driver detection, because there are no available additional drivers for this computer

PR #950, opened 2019-02-04
homepage: add conditional compilation for using dynamic content

decathorpe:
On fedora (and probably most other distributions as well), the appcenter Homepage is pretty useless - the huge top banner is static (since there's no server content), and recommendations are mostly empty most of the time, as well (only a few "appcenter applications" are available from the repos).

appcenter-fedora

This PR puts all Homepage widgets that depend on dynamic server content behind a conditional compilation flag in meson. The default behaviour isn't changed, but on distributions that aren't elementary, the widgets that are useless / static can be disabled - leaving only the Categories on the Homepage (which is hence named "Categories" in the interface in this case, as well).

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,5 +1,6 @@
 option('blacklist_file', type : 'string', value : 'appcenter.blacklist', description : 'The name of the file listing blacklisted applications')
 option('curated', type : 'boolean', value : true, description : 'Differentiate between apps created for elementary and not')
+option('homepage', type : 'boolean', value : true, description: 'Dynamic homepage using remote server data')
 option('icon', type : 'string', value : 'system-software-install', description : 'The name of the icon')
 option('libunity', type : 'boolean', value : true, description : 'Use libunity to show badges')
 option('name', type : 'string', value : 'AppCenter', description : 'The name of the application')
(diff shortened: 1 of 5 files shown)


decathorpe:
Side note: I didn't catch the shy "Trending" section on the screenshot above, but it sometimes appears. However, it's pretty useless on fedora, as I said - it's usually the same three apps (Quilter - if the Appcenter COPR is enabled, sequeler, and notejot).

decathorpe:
Hm, I've never worked with translations before, but nothing is removed - either the string is "Home" or it's "Categories", and in both cases it's always wrapped in `_()`. What would I have to change?

tintou:
They both have to appear once between `_()` also, I wonder why change the wording when Home fits here well…

TingPing:
> and in both cases it's always wrapped in _(). What would I have to change?

Add `N_()` on the static strings should work.

codygarver:
For the sake of simplicity, it's best to revert the Home change.

decathorpe:
@codygarver @tintou You're right, I'll change it back to use "Home" in both cases.

decathorpe:
Thanks! This will make my work for fedora easier (and others will probably benefit too).

PR #955, opened 2019-02-14
Package refactoring

davidmhewitt:
This removes the vast majority of PackageKit specific code from `Package.vala` to move closer to being able to be used as a more backend independent class in the future.


davidmhewitt:
The following things need to be tested as they have been touched by this branch:
* Installing/updating/removing packages
* Update checking
* List of packages in installed tab
* Download size display
* App information (version, description, title, etc...) fetching

danrabbit:
I don't want to lose momentum here, so... I can confirm that the following still works:

- [x] Installing/updating/removing packages
- [x] Update checking
- [x] List of packages in installed tab
- [x] Download size display
- [x] App information (version, description, title, etc...) fetching

PR #957, opened 2019-02-24
Client refactoring

davidmhewitt:
This builds on #955 so this should be merged first.

This moves the AppStream code that's currently in `Client.vala` to `PackageKitClient.vala`.

My reasoning:
* The AppStream code is currently looking at the systemwide AppStream data, which currently **is** the APT AppStream data.
* Other backends tend to put their AppStream data outside of the systemwide locations. For example, Flatpak has it's AppStream data in `/var/lib/flatpak/appstream`. This has to be parsed separately.
* This then frees up `Client.vala` to be the aggregator of different backends. A lot of the frontend code goes via this class, so it avoids too much rework of frontend code.

As a result, most of the methods in `Client.vala` are now just passthroughs to `PackageKitClient.vala`. But in future, will perform aggregated actions on all of the backends.

Things to test that have been touched by this branch:
* Update checking
* Fake package updates
* Local metadata
* Installed applications list
* Listing applications by category
* Searching for applications
* Listing packages by author


src/Core/UpdateManager.vala
```diff
@@ -20,56 +20,102 @@
...
+        os_updates = new AppCenterCore.Package (os_updates_component);
+    }
 
-                fake_packages = {};
+    public async uint get_updates (Cancellable? cancellable = null) {
+        var apps_with_updates = new Gee.TreeSet<Package> ();
+        uint count = 0;
+
+        unowned PackageKitClient client = PackageKitClient.get_default ();
+        var pk_updates = yield client.get_updates (cancellable);
```
danrabbit: Compiler warning about an unhandled GLib error ;)

PR #964, opened 2019-03-08
RestartDialog: Subclass Granite.MessageDialog

### LINKED ISSUE #682: RestartDialog: get_action_area is deprecated
We should replace it with a Granite dialog, then we can eventually address the deprecation and do the styling there instead.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56973986-restartdialog-get_action_area-is-deprecated?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #682


jeremypw:
I do not understand the purpose of this dialog.  It seems to show when there is an error thrown by `SuspendControl.get_default ().reboot ()`.  In effect this means if the user presses "Cancel" on the system reboot dialog then another dialog shows asking "Are you sure you want to restart?".  This does not make sense.

danrabbit:
> I do not understand the purpose of this dialog. It seems to show when there is an error thrown by `SuspendControl.get_default ().reboot ()`. In effect this means if the user presses "Cancel" on the system reboot dialog then another dialog shows asking "Are you sure you want to restart?". This does not make sense.

I thought it was supposed to be a fallback for when the system doesn't provide a restart dialog on dbus? But yeah I kind of feel like we'd probably be safe to remove it if you feel that way too

jeremypw:
You should at least check what kind of error has been returned by the system and only show this dialog when appropriate.

jeremypw:
@danrabbit Is this OK now?  I have addressed the issue with the fallback dialog.

danrabbit:
Proposed another branch removing the dialog instead

PR #243, opened 2017-04-24
Implement drivers (AppStream)

donadigo:
Implements drivers and fixes #12.

After discussion on the Slack channel I decided to give AppStream implementation of drivers a go. We no longer depend on any Ubuntu / Debian specific code. Everything is based on AppStream and PackageKit, however scanning modaliases on the system is pretty much a Vala version of the UbuntuDrivers component. For more details about the code and it's design, look into the diff, I left some explanation for each driver related method.

Note that currently there is no AppStream data to test this branch with and this is kind of a problem since this feature is being asked by lots of people. So how did I test it?
The way to test this is to fool AppStream that some components are drivers. All of the compiled data is stored in `/usr/share/app-info/yaml/pantheon_xenial-main_amd64.yml.gz` file. Open this file with archive manager (with admin), and edit the .yml file inside of it. Choose some component and change it's `Type` field from `desktop-app` to `driver`. Next you will need a supported modalias by your system. These can be found by e.g: executing `ubuntu-drivers debug` and choosing one of the `modalias` strings. After that you can paste this under the `Provides` field:
```
modaliases:
  - your_modalias
```

and save the file.

Sorry if this is really complicated but this is the only way I found to test this. Perhaps Matthias knows a better way.

PR #965, opened 2019-03-09
Move drivers logic to separate class

davidmhewitt:
Start to form the concept of separate backends and move the `ubuntu-drivers` logic into one.

The concept of backends is not complete yet as drivers are still technically part of the "PackageKit" backend. So we abuse the fact that AppCenter only has a PackageKit backend so far to prevent building a load more shims in the new drivers backend.

This will need to be done, but I'll do it as a separate PR to keep things easier to review.

Things to test:
* All things related to drivers
* The installed/update view's list of packages

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -72,7 +72,7 @@ public class AppCenter.App : Gtk.Application {
         var client = AppCenterCore.Client.get_default ();
         client.operation_finished.connect (on_operation_finished);
         client.cache_update_failed.connect (on_cache_update_failed);
-        client.update_check_finished.connect (on_updates_available);
+        client.installed_apps_changed.connect (on_updates_available);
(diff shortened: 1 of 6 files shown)


danrabbit:
Requesting a review from @donadigo here since he did some work with drivers from appstream https://github.com/elementary/appcenter/pull/243 and it might be relevant/affected.

I'm also curious about any implications this might have with fwupd support: https://github.com/elementary/appcenter/issues/498

davidmhewitt:
I'll try and give my take on those things ahead of time.

AppStream drivers should be fine once AppStream data exists for them. We'd just need the code from #243 updating to fit the refactoring.

I think fwupd would be a separate backend so if anything this refactoring should make it easier to add.

PR #966, opened 2019-03-10
Define an interface for backends

davidmhewitt:
This builds on #965 so that should be merged first.

I'm nearing completion of my refactoring now, there's just a little bit of work to do around making the progress callbacks less PackageKit specific. Once that's done in another PR, I'll start work on implementing a new backend for a containerized packaging format that we all know and love.

Things to test:
* Updating/Installing/Removing packages (particularly from the installed view)
* Details of updatable packages
* Total size of updates on updates view.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ You'll need the following dependencies:
 * gettext
 * libappstream-dev (>= 0.10)
 * libgee-0.8-dev
-* libgranite-dev (>=0.5)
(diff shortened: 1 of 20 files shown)


davidmhewitt:
I should add that this reduces the active memory consumption of AppCenter by around 10% due to not having to hold TreeSets of `Pk.Package` and `Pk.Detail` in memory anymore

Issue #609, opened 2018-02-10
Collapse non-curated list


cassidyjames:
If we have non-curated apps in a list, we should collapse them under the header. This deprioritizes them without straight up removing them from AppCenter.

Something up for discussion is if we want to _always_ collapse them, or only when there are curated apps in the same list. I imagine the latter would work best with downstreams and for search results where there are no curated apps.

danrabbit:
I'm gonna punt since this is an enhancement and not a release blocker

inputbim:
seperate category in appstore to only see curated apps? hard to discover curated apps without going to the website

matfantinel:
I think it's worth considering if this should be applied to all uncurated apps regardless of source (Ubuntu repos, Flathub, etc) or just to the .deb ones. Since Flathub is a source to many well-known apps like Slack, Spotify, and soon Firefox, and Flatpak provides an extra security layer, maybe it's okay to keep them displayed below the curated ones. Specially considering the user has knowingly added these sources manually via Sideload, unlike the Ubuntu repos

cassidyjames:
I'm still inclined to treat them the same as other non-curated apps. On elementary OS today, there are not that much stronger of security guarantees, as we're still using Xorg and thus these apps still aren't sandboxed.

danrabbit:
Closing since we don't show apps from Ubuntu in the default install and if someone adds a remote we shouldn't hide the remote that they intentionally added

PR #971, opened 2019-03-21
Flatpak support

### LINKED ISSUE #712: flatpak application support [$160]
Would be interesting an option that presents flatpak applications

<bountysource-plugin>

---
There is a **[$160 open bounty](https://www.bountysource.com/issues/58197944-flatpak-application-support?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #712 

Requires associated deb-packaging PR: #996 

This branch deals with the technical implementation of getting flatpak support into AppCenter (displaying, installing, removing, updating, etc...). It does not touch any UI code and any flatpak packages just show up alongside any from the apt repositories, which can be pretty confusing visually. PR #992 follows on from this branch to de-duplicate these listings and improve the UX.


davidmhewitt:
I've opened a PR in appstream that would help me solve the icon issue (actually caused by whatever generates appstream directory structures in flatpak repos). If this or an equivalent could be merged into appstream, it should be **much** easier to resolve this icon loading issue.

https://github.com/ximion/appstream/pull/224

cassidyjames:
@elementary/ux how do we want to handle colliding apps? @davidmhewitt how does this branch currently do it? I.e. if someone has a Flatpak repo that contains app `com.example.foo.bar`, and their Debian repo also contains `com.example.foo.bar`, do we show both? Do we deduplicate? How do we determine which to install?

I believe GNOME Software includes a "source" dropdown menu, where it shows something like "RPM (Fedora)", "RPM (fedora-testing)", "Flatpak (Flathub)", etc. and the user can choose which listing to view and install. I don't know that I love this, but I also don't know the best way to handle it.

davidmhewitt:
As it was the easiest thing to do, this currently just shows duplicate apps in all the results lists in a really confusing manner :stuck_out_tongue:

Apps are assumed to be curated if they're from the AppCenter debian repo for now, so they come up higher in search results than the flatpak equivalents. But you generally can't tell the difference if you're on the apps info page.

But, there are many more than two possible duplicate packages, even when we just have Flatpak and Debian backends...

You could have flatpak user installations, system installations, from multiple remotes, from multiple branches, so on and so forth

Even with just adding flathub and a GNOME flatpak repo, I already have like 4 of all the GNOME apps. 3 from various flatpak sources and 1 from ubuntu repos.

cassidyjames:
@davidmhewitt _would_ it make sense to try and combine the listings and then offer a toggle between the different specific versions? And have some sort of prioritization ordering where we choose the default?

davidmhewitt:
I think so. The way that currently makes most sense in my head is that search result and browse lists are de-duplicated with some way of selecting which one to install from the app info page.

But, on the installed tab, they're not de-duplicated with some kind of indicator of where they came from so that they can be updated separately if desired.

Then when clicking on one in the installed view, the app info view pre-selects the right source, otherwise we prioritize in this order (for the time being):
- Curated flatpak repo (when it exists)
- Curated Debian repo
- Flatpak from anywhere

davidmhewitt:
On second thoughts, I'd probably revise my above comment to have flatpak from 3rd party repos be the last option in the priority list as I've definitely seen some descriptions on flathub like, "You gotta do x to the sandboxing settings to make this work properly", so these apps would give a worse user experience than their debian equivalents.

Would it make sense to have a button/dropdown that said something like, "More versions"? Because, the users may not necessarily care about the packaging technology, but there are likely to be version differences between them (which is more useful/interesting). Some illustrative examples:

```
Dippi:
- Elementary curated (1.0.7)
- Flathub (1.0.6)
```

```
GNOME Calculator:
- Non-curated Ubuntu Repository (3.28)
- Flathub (3.32)
- Gnome nightly (commit: f6ed73a)
```

tintou:
@davidmhewitt I think that we shouldn't switch between sources and install from the curated one but upgrade using the same source as the previous version (so that if that's been installed from flathub it takes to one from flathub).
 * Deliberately choosing a different (and non supported?) version should be a command-line experience type of thing
 * If you switch from Flatpak to .deb you're likely to lose some settings
 * Downgrading a package might lead to issue (incompatible cache databases…)

cassidyjames:
@tintou that sounds sane, but the edge case I think is when there are multiple non-curated versions and no curated version. :smile: But yeah maybe at that point we just say, well, it's uncurated so we show it all. I think we want to more heavily deprioritize uncurated apps anyway, i.e. collapsing the list down when there are curated results (#609).

So:
- If a curated version of the id exists, don't show any non-curated version (users can use Terminal or whatever to manage unsupported versions)
- If no curated version of the id exists, just show everything as-is under the non-curated section
- Also de-prioritize non-curated results (#609) which will help hide that mess if people go adding duplicate Flatpak repos or whatever

cassidyjames:
Another thing we'll need to handle eventually is if there are multiple curated IDs… i.e. if Dippi comes to AppCenter Flatpak, then it'll exist in the deb repo and the Flatpak repo. In this case I think we prefer the Flatpak one since it's the direction we're going.

danrabbit:
Some of the progress feedback is a little excessive:
Screenshot from 2019-04-08 12 44 17@2x

davidmhewitt:
@danrabbit 

From http://docs.flatpak.org/en/latest/libflatpak-api-reference.html#FlatpakProgressCallback :
> `status`         A status string, suitable for display

That excessive "suitable for display" string and the progress percentage is all we get from `libflatpak`, there isn't an enum that has different states like Waiting, Downloading, Installing, etc...

So it's either the excessive string, nothing, or something generic like, "Working".

mmstick:
Parsing the untranslated string, and then arranging the data however you want, is one possible option.

danrabbit:
@davidmhewitt In that case, maybe we should drop the label and use a tooltip instead (for all backends ofc).

@elementary/ux thoughts?

davidmhewitt:
@mmstick Indeed, but it's also a pretty brittle option that's liable to break if flatpak alter their string format/messages, so I'd prefer not to go that route if possible.

cassidyjames:
Yeah I'd say something generic would be fine, like "Working", or just not showing a status and throwing the status into a tooltip in case people want more info. But I'd also like to report something against Flatpak to get us a nice enum of statuses so we can craft something a bit better in the future.

src/Utils.vala
```diff
@@ -39,6 +39,33 @@ namespace Utils {
...
+            warning ("Error while getting file age: %s", e.message);
+            return uint.MAX;
+        }
+
+	    if (info == null) {
+            return uint.MAX;
+        }
+
+	    uint64 mtime = info.get_attribute_uint64 (FileAttribute.TIME_MODIFIED);
+	    uint64 now = (uint64) get_real_time () / 1000000;
```
tintou: Wouldn't https://valadoc.org/glib-2.0/time_t.html make more sense than this very precise time?

davidmhewitt: `FileAttribute.TIME_MODIFIED` is a `uint64` and I personally prefer getting the mtime this way as we can use `FileAttribute` and proper GLib exception handling. If we wanted the mtime as time_t, I think we'd have to use something like https://valadoc.org/glib-2.0/GLib.Stat.html which isn't as clean IMO.

Let me know what you think, or if you had a different idea.

tintou: I'm more talking about the `now` value as you don't need the precision of `get_real_time ()`

src/Core/FlatpakBackend.vala
```diff
@@ -0,0 +1,860 @@
...
+                    assert_not_reached ();
+            }
+        }
+
+        return true;
+    }
+
+    private FlatpakBackend () {
+        worker_thread = new Thread<bool> ("flatpak-worker", worker_func);
+        appstream_pool = new AppStream.Pool ();
```
donadigo: Is there a way we could make this pool shared across backends? I'm not sure if creating the pool causes disk reads or not but it would be nice.

src/Core/FlatpakBackend.vala
```diff
@@ -0,0 +1,860 @@
...
+
+    public bool working { get; private set; }
+
+    private string local_metadata_path;
+
+    private bool worker_func () {
+        while (thread_should_run) {
+            working = false;
+            var job = jobs.pop ();
+            working = true;
```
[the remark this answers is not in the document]
davidmhewitt: `jobs.pop` is a blocking non-async call within that thread, so if there are no jobs then it waits there until there is a job, at which point `working` is set to true. It's the same as having `working = false` at the bottom of the loop, but I've moved it there instead as it's probably clearer.

src/Core/FlatpakBackend.vala
davidmhewitt: I would prefer to keep these separate so that there can be no conflicts of AppStream data between Flatpak and PackageKit. I know it's been designed to allow multiple packages with the same ID to come from different sources, but I'd had to do some custom metadata paths, XML wrangling and icon symlinking to get the Flatpak metadata to work properly, and I'd prefer to keep that separate.

src/Core/FlatpakBackend.vala
```diff
@@ -0,0 +1,853 @@
...
+        package_list = new Gee.HashMap<string, Package> (null, null);
+
+        local_metadata_path = Path.build_filename (
+            Environment.get_user_cache_dir (),
+            "io.elementary.appcenter",
+            "flatpak-metadata"
+        );
+
+        try {
+            installation = new Flatpak.Installation.system ();
```
donadigo: This should go into a `static construct` block instead of a normal construct.

src/Core/FlatpakBackend.vala
```diff
@@ -0,0 +1,853 @@
...
+
+    public async PackageDetails get_package_details (Package package) throws GLib.Error {
+        var details = new PackageDetails ();
+        details.name = package.component.get_name ();
+        details.description = package.component.get_description ();
+        details.summary = package.component.get_summary ();
+
+        var newest_version = package.get_newest_release ();
+        if (newest_version != null) {
+            details.version = package.get_newest_release ().get_version ();
```
donadigo: Could be simplified to `newest_version.get_version ()`.

src/Core/FlatpakBackend.vala
```diff
@@ -0,0 +1,853 @@
...
+                    delete_folder_contents (child, cancellable);
+                    child.delete ();
+                }
+            }
+        } catch (Error e) {
+            warning ("Error while cleaning up flatpak metadat directory: %s", e.message);
+        }
+    }
+
+    private void perform_xml_fixups (string origin_name, File src_file, File dest_file) {
```
donadigo: Could be static as well.

src/Core/UbuntuDriversBackend.vala
```diff
@@ -18,6 +18,18 @@
  */
 
 public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
+
+    private bool _working;
+    public bool working {
```
donadigo: Since this is only getting and setting the property, wouldn't it make sense to just do what is in the flatpak backend: `public bool working { get; private set; }`?

src/Core/BackendAggregator.vala
```diff
@@ -24,6 +24,28 @@ public class AppCenterCore.BackendAggregator : Backend, Object {
         backends = new Gee.ArrayList<unowned Backend> ();
         backends.add (PackageKitBackend.get_default ());
         backends.add (UbuntuDriversBackend.get_default ());
+        backends.add (FlatpakBackend.get_default ());
+
+        foreach (var backend in backends) {
+            backend.notify["working"].connect (() => {
+                // Trigger a notify on this class, value doesn't matter
+                working = true;
```
tintou: Wouldn't it make more sense?
```suggestion
                notify_property ("working");
```

davidmhewitt: It would indeed. Every day is a school day :wink: 

I was trying to figure out how `notify["working"] ()` worked and gave up without realising that `notify_property` exists :slightly_smiling_face:

tintou:
Thanks @davidmhewitt for this!

tintou:
Tested and works, we can fine-tune and fix regressions in later iterations

PR #972, opened 2019-03-30
Fix crashing issues when cancelling package operations

davidmhewitt:
During testing downstream, @isantop and @mmstick found and were able to reproduce some crashing issues when cancelling package operations:
https://github.com/pop-os/shop/pull/117#pullrequestreview-219171652

Since moving more of the backend logic into threads, some of the signals being fired back to the UI were now running in non-UI threads causing issues. If we wrap the firing of these signals in an `Idle.add` we can ensure they run on the main GTK thread.


tintou:
I'd really prefer adding a comment above the signals and using Idle.add inside of the consumer function so that we have less clutter and be clear that Gtk subclasses are the one that should care.

davidmhewitt:
@tintou done

PR #973, opened 2019-04-01
More package operation cancellation crash fixes

davidmhewitt:
Further testing by @mmstick found more crashing issues relating to crashing package operations. I wasn't able to reproduce all of these, but I have cleaned up more instances where UI operations could be called from other threads and this is apparently more stable.


mmstick:
Confirming that this fixes the x11 thread panics when clicking the cancel button as an application is being installed or removed.

PR #977, opened 2019-04-06
Check if appcontainer is destroyed before updating

davidmhewitt:
Cherry picked from my flatpak branch. Fixes an edge case issue where the container can be destroyed before the Idle code runs which causes issues.

### DIFF
--- a/src/Widgets/AppContainers/AbstractAppContainer.vala
+++ b/src/Widgets/AppContainers/AbstractAppContainer.vala
@@ -66,7 +66,7 @@ namespace AppCenter {
         private Mutex action_mutex = Mutex ();
         private Cancellable action_cancellable = new Cancellable ();
 
-        private bool destroyed = false;
+        private bool is_destroyed = false;
 
         private enum ActionResult {
             NONE = 0,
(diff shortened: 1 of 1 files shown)


tintou:
@danrabbit no, it's not, please review carefully, widget lifecycle should not be tracked this way.
@davidmhewitt The correct way to do it would be to remove the Idle source when `destroy` has been called:
```vala
uint state_source = 0U;
…
destroy.connect (() => {
    if (state_source > 0) {
        GLib.Source.remove (state_source);
    }
});
…
private void on_package_state_changed () {
    if (state_source > 0) {
        return;
    }

    state_source = Idle.add (() => {
        if (is_destroyed) {
            state_source = 0U;
            return GLib.Source.REMOVE;
        }

        update_state ();
        state_source = 0U;
        return GLib.Source.REMOVE;
    });
}
```
But note that this might also just be a side-effect of some `update_state` implementation doing wrong things because the reference should still be there

davidmhewitt:
Thanks @tintou , that looks like it makes more sense, I'll try that out and open a PR if it still fixes the issue I was experiencing.

The issues I was experiencing were segfaults coming from `InstalledPackageRowGrid.vala` in the overriden `update_state` method when setting properties on child widgets. Investigation showed that these child widgets were destroyed between the idle source being added and it running.

PR #980, opened 2019-04-07
Refactor progress callbacks to be backend agnostic

davidmhewitt:
I think this is the last bit of refactoring needed before the flatpak support is fully unblocked. With this, it'll be possible for backends to report operation progress without using any PackageKit specific objects.

Things to test would be any of the views involving progress bars, cancellation of operations and updating a batch of packages from the installed view.


src/Core/PackageKitBackend.vala
```diff
@@ -769,6 +795,59 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
         return (PackageDetails)job.result.get_object ();
     }
 
+    private void ProgressCallback (Pk.Progress progress, Pk.ProgressType type) {
```
donadigo: At first I didn't understand why this method was called in the callbacks of calls to PackageKit, we should perhaps rename it to something more descriptive like `update_progress_status ()` since it updates global state of the backend.

src/Core/ChangeInformation.vala
```diff
@@ -161,43 +88,22 @@ public class AppCenterCore.ChangeInformation : Object {
...
-                last_progress = progress.percentage;
-                double progress_sum = current_progress + last_progress;
-                this.progress = progress_sum / progress_denom;
-                progress_changed ();
-                break;
-            case Pk.ProgressType.STATUS:
-                status = (Pk.Status) progress.status;
-                status_changed ();
-                break;
+        if (this.progress != (progress / 100.0f)) {
```
donadigo: Is this needed to normalize this progress here instead of expecting from the backend to normalize it on it's own and then pass the normalized version to the method?

src/Widgets/AppContainers/AbstractAppContainer.vala
```diff
@@ -375,7 +375,7 @@ namespace AppCenter {
         protected virtual void update_progress_status () {
             Idle.add (() => {
                 progress_bar.text = package.get_progress_description ();
-                cancel_button.sensitive = package.change_information.can_cancel;
+                cancel_button.sensitive = package.change_information.can_cancel && !package.changes_finished;
```
donadigo: Can you clarify this change here?

davidmhewitt: This is a bug that I think exists in master where the cancel button in the update view is sensitive for packages that have finished updating, but other package updates are ongoing.

It's not really required as part of these changes, but just something I noticed while testing the `can_cancel` stuff.

src/Utils.vala
```diff
@@ -38,5 +38,82 @@ namespace Utils {
             dialog.show_all ();
         }
     }
+
+    public static string pk_status_to_string (Pk.Status status) {
```
tintou:
```suggestion
    public static unowned string pk_status_to_string (Pk.Status status) {
```

tintou:
Only a tiny thing to commit and it's ready to go to me :rocket:

src/Utils.vala
davidmhewitt: @tintou Thanks! Done :slightly_smiling_face:

PR #985, opened 2019-04-12
Fix broken builds due to Hindi translations

cassidyjames:


cassidyjames:
Fixed in https://github.com/elementary/appcenter/commit/14f743051212d1a74e6b68fe58fe36c446e96062

PR #989, opened 2019-04-14
Add a backend aggregator to simplify OS updates from multiple backends

davidmhewitt:
This creates a "backend" that represents all of the backends available to AppCenter.

This helps in the case of the OS Updates "package" which could contain updates from multiple backends. I'm currently thinking that the Flatpak runtimes will fall into OS updates as we'll be shipping an elementary runtime as part of the OS.

This will help with that as we can say the OS Updates package comes from the aggregated backend and then the individual backends will update the packages within that metapackage that belong to them. I've already tested this against the Flatpak branch, just cherry picking this out to keep that branch cleaner.

For testing, there isn't really much logical difference to the code here, it's just a bit of abstracting. So if OS updates and searching for apps still works, I'm confident this won't have any regressions elsewhere.


src/Core/BackendAggregator.vala
```diff
@@ -0,0 +1,151 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: David Hewitt <davidmhewitt@gmail.com>
+ */
+
+public class AppCenterCore.BackendAggregator : Backend, Object {
+    public Gee.ArrayList<unowned Backend> backends;
```
donadigo: While we're here, I don't see this array used anywhere in the code, at least not publicly by some other object. It is probably safe to remove this or at least have it return a read only view to the list.

src/Core/PackageKitBackend.vala
```diff
@@ -474,8 +474,17 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
 
         Pk.Exit exit_status = Pk.Exit.UNKNOWN;
         string[] packages_ids = {};
-        foreach (var pk_package in package.change_information.updatable_ids) {
-            packages_ids += pk_package;
+        foreach (var pk_package in package.change_information.updatable_packages) {
```
donadigo: I wonder if it would make sense to either sort `updateable_packages` or split the packages for different backends to avoid looping through all of them in each backends but I guess that's not the focus of this PR.

src/Core/ChangeInformation.vala
```diff
@@ -19,6 +19,12 @@
  */
 
 public class AppCenterCore.ChangeInformation : Object {
+
+    public struct UpdatablePackage {
+        Backend backend;
```
donadigo: Have you tried making this unowned?

src/Core/PackageKitBackend.vala
davidmhewitt: @donadigo That was a really good point, thanks! I've swapped out the data structure for a MultiMap, so now we only iterate over the packages belonging to the backend in question.

src/Core/PackageKitBackend.vala
```diff
@@ -474,10 +474,17 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
 
         Pk.Exit exit_status = Pk.Exit.UNKNOWN;
         string[] packages_ids = {};
-        foreach (var pk_package in package.change_information.updatable_ids) {
+        foreach (var pk_package in package.change_information.updatable_packages[this]) {
             packages_ids += pk_package;
         }
 
+        if (packages_ids.length == 0) {
+            job.result = Value (typeof (bool));
```
donadigo: I'm not sure if that will work, but something tells me that `job.result = true` will work, because I believe Vala will automatically translate native expressions to a `Value`.

PR #991, opened 2019-04-19
Search homepage packages in parallel with newer packagekit

davidmhewitt:
This supersedes #613 and now isn't technically blocked by having a newer packagekit as the version check is done at runtime.

However, with current Juno PackageKit version it has the unfortunate effect of waiting for the state of the packages in all 3 banners to come back before displaying any of them, unlike master. So, it's not really any slower for the homepage to be fully populated, but it does feel like it's doing nothing for longer.

With PackageKit >= 1.1.10, all of the banners load within about 3 seconds of AppCenter opening on my machine, as opposed to the ~25 seconds that we have with current PackageKit in Juno.

I should also point out that this intentionally restricts the backend that the packages for the banners can come from to PackageKit as that's definitely what we want for the near future.

Since Pop!_Shop only has one banner and has the benefit of being able to run on newer PackageKit anyway, I think this will be particularly interesting to @mmstick .


cassidyjames:
@codygarver considering the incredible performance improvement on a very important piece of our stack (showing users apps they can buy in AppCenter), would it be possible to backport newer PackageKit to Juno? I'd really like to benefit from this.

src/Core/PackageKitBackend.vala
```diff
@@ -42,6 +42,26 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
 
     public bool working { get; private set; }
 
+    // The aptcc backend included in PackageKit < 1.1.10 wasn't able to support multiple packages
+    // passed to the search_names method at once. If we have a new enough version we can enable
+    // some optimisations when looking up packages
+    public bool supports_parallel_package_queries {
+        get {
+            var control = new Pk.Control ();
```
donadigo: Because a `Pk.Control` instance connects to a DBus interface, it probably should be created ever only once. If you didn't know already, this class already contains one: https://github.com/elementary/appcenter/blob/0419a054cb4449b72c51cbec4fddfc0a8dc1353a/src/Core/PackageKitBackend.vala#L121.

So instead of creating it everytime when accessing this variable, you should instead have a `private static Pk.Control?` variable that is a singleton:
```
private static Pk.Control? control;
private static unowned Pk.Control get_control () {
    if (control == null) {
        control = new Pk.Control ();
    }

    return control;
}
```

And when it's needed, access it via `get_control ()`.
`supports_parallel_package_queries` could also be static because it doesn't access anything inside the class itself.

tintou:
See https://github.com/elementary/os/issues/168 for the state of PackageKit upgrade

src/Core/PackageKitBackend.vala
```diff
@@ -42,6 +42,36 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
 
     public bool working { get; private set; }
 
+    // The aptcc backend included in PackageKit < 1.1.10 wasn't able to support multiple packages
+    // passed to the search_names method at once. If we have a new enough version we can enable
+    // some optimisations when looking up packages
+    public static bool supports_parallel_package_queries {
+        get {
+            if (control.backend_name != "aptcc") {
```
vjr: If `backend_name != "aptcc"` should it really return true or false here?

davidmhewitt: It should return true because other backends support parallel queries fine. There was a bug in aptcc that has been fixed in the newer versions.

davidmhewitt:
To test (without overwriting locally installed PackageKit):

1. Clone https://github.com/hughsie/PackageKit to any version newer than 1.1.9 (master will do).
2. Build with:
```
sudo apt build-dep packagekit
./autogen.sh --enable-aptcc --enable-local
make -j3
```
3. Run newer `packagekitd` locally with:
```
cd src
sudo killall packagekitd
sudo ./run-pk.sh aptcc
```
4. Now fire up appcenter from this branch (making sure any old instances are killed too), and observe the speed :dash:

cassidyjames:
I do think for elementary OS this needs to be blocked by the newer PackageKit. It takes a long time for anything perceptible to happen, and it looks like it's not cached when the window is closed—it has to reload everything every time you close and reopen the window.

But if we get the newer PackageKit, holy cow this is a massive improvement.

davidmhewitt:
We now have PackageKit 1.1.12 in the `elementary/staging` PPA so testing this is now as simple as:

* Add that PPA
* `sudo apt upgrade` to get PackageKit 1.1.12
* Restart PackageKit with `sudo killall packagekitd`
* Test AppCenter from this branch

davidmhewitt:
This is no longer blocked by the available PackageKit version and should no longer exhibit the duplicate package issue.

PR #992, opened 2019-04-21
De-duplicate packages from multiple sources

davidmhewitt:
This continues on from the `flatpak` branch that deals with the technical side of implementing flatpak support with some work to improve the UX of having multiple sources for the same packages.

App info pages should now have a dropdown to select between different sources for the same package and there should be no duplicated results in search or category views. The installed view should still show the packages separately so that they can be updated independently.


cassidyjames:
@davidmhewitt I think I'm fine with that location—at least to start—if it only shows on listings with multiple possible packages.

davidmhewitt:
@cassidyjames Great! Yep - that's the plan, I'm just waiting for #980 to be merged before I continue much further because I'm in merge hell between my various branches.

davidmhewitt:
I think I'm pretty much done with this being a work in progress now, I'll leave it in draft for now while I have multiple branches open so I can still be in control of manually merging between branches without people clicking the merge master button :stuck_out_tongue: 

It might still need some UX work, but I'd consider it pretty feature complete for what we'd want from initial flatpak support. So as much testing and feedback as possible would be great now.

davidmhewitt:
There seems to be a text contrast issue with the dropdown and some apps, could someone better at CSS than me have a go at fixing the application specific styles for this? @elementary/ux 

image

danrabbit:
This is getting pretty awesome!

One thing I noticed that I'm not sure is expected is that the back button remembers "navigating" between different sources. So for example:

1. Navigate to Fondo
2. Switch to the "flathub" source
3. Click "Back"

Expectation: I go back to the previous view (like search or categories)
Reality: I stay on Fondo and switch to the AppCenter source

davidmhewitt:
@danrabbit Good point. I initially did this because it would be quite difficult (and slow) to detect if there are multiple versions of the same package installed (i.e. two identical entries in the list) and then show a label to differentiate them.

I've now restricted it to now only show the extra bit of label if there are multiple versions of that package _available_, but not necessarily installed. Hopefully that's a suitable compromise.

mmstick:
Still seems to be some duplication here and there. Hard to tell which is a debian package, and which is a flatpak.

Screenshot from 2019-05-28 16-22-15

davidmhewitt:
@mmstick I've just looked at that example specifically and the flathub package has an AppStream ID of `org.wireshark.Wireshark` and the deb package has `wireshark.desktop`. There's only so much matching we can do. If a developer or packager decides to change the ID that's supposed to uniquely identify an application, it will be difficult to match.

I'll open a separate issue about it being hard to tell which is which though to prompt some discussion

PR #995, opened 2019-05-04
Reuse package objects on reload

davidmhewitt:
Instead of clearing out the list of packages in the backend and building it again when we reload the appstream data, we can simply replace the appstream component on the ones that already exist.

This helps with a few things where the package is already referenced in cached AppInfoViews or banners and holds the reference but a newer, updated package has come along.

### DIFF
--- a/src/Core/PackageKitBackend.vala
+++ b/src/Core/PackageKitBackend.vala
@@ -133,7 +133,7 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
         } catch (Error e) {
             critical (e.message);
         } finally {
-            var modified_packages = new Gee.TreeSet<string> ();
+            var new_package_list = new Gee.HashMap<string, Package> ();
             var comp_validator = ComponentValidator.get_default ();
             appstream_pool.get_components ().foreach ((comp) => {
                 if (!comp_validator.validate (comp)) {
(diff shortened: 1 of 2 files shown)


src/Core/PackageKitBackend.vala
```diff
@@ -128,24 +128,31 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
...
         } finally {
+            var new_package_list = new Gee.HashMap<string, Package> ();
             var comp_validator = ComponentValidator.get_default ();
             appstream_pool.get_components ().foreach ((comp) => {
                 if (!comp_validator.validate (comp)) {
                     return;
                 }
 
-                var package = new AppCenterCore.Package (this, comp);
                 foreach (var pkg_name in comp.get_pkgnames ()) {
```
tintou:
```suggestion
                foreach (unowned string pkg_name in comp.get_pkgnames ()) {
```

tintou:
A tiny bit (even if you're not strictly touching this line, it'd be nice to fix this) and ready to go

vjr:
> A tiny bit (even if you're not strictly touching this line, it'd be nice to fix this) and ready to go

@tintou what is the reason for changing from `var pkg_name` to `unowned string pkg_name` ? is it just a performance improvement (prevent string copy) or is there a memory leak/referencing issue?

donadigo:
@vjr this is to prevent string copies.

Issue #975, opened 2019-04-04
Inappropriate message when Installed View first shown


jeremypw:
* Open AppCenter
* Click "Installed" button
* The view opens with the message " No Results. No Apps could be found. Try changing search terms" 

This message persists while the "working" spinner is active, which may be for several seconds, before the actual list of installed apps is shown.

While the spinner is active, either no message or a more appropriate message should be shown.

cassidyjames:
I thought the installed page showed installed apps instantly and then loading just pulled in updates. Is this a regression from @davidmhewitt's recent refactoring?

jeremypw:
This did not used to happen so yes, it is a regression.

mcclurgm:
The duplicate issue was mine, but I have some proof of concept work that may be useful here so I'll copy it over.

This won't fix everything in this issue (I didn't realize the full extent of the regression), but it's a step toward working with the current state. We can implement a better placeholder within this framework by changing it in the `AbstractAppList` subclass.
This won't fix the issue of waiting to load the installed list, but at least it'll fix one of the issues with the current implementation.

In `AppListUpdateView.vala`, add the following code to `construct` (about line 53):
```
var loading_placeholder = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 12);
loading_placeholder.expand = false;
loading_placeholder.halign = Gtk.Align.CENTER;
var spinner = new Gtk.Spinner ();
spinner.start ();
var message = new Gtk.Label (_("Loading installed apps…"));
loading_placeholder.pack_start (spinner);
loading_placeholder.pack_start (message);
loading_placeholder.show_all ();
list_box.set_placeholder (loading_placeholder);
```

It'll make a loading screen that looks like this:
![](https://user-images.githubusercontent.com/29022469/78456027-bdb54b80-765e-11ea-9e14-56375ebd7cc4.png)


Obviously, this needs design work, and we'd have to manage the spinner in this case. But it shows that we can subclass AbstractAppList to present a more appropriate message.

jeremypw:
Not sure we need two spinners.

davidmhewitt:
Since it's impossible to have no installed apps, since AppCenter itself appears in this list, the "No Apps could be found" message is definitely inappropriate.

Agree that we probably don't need a second spinner though.

PR #1007, opened 2019-05-27
Cache list of Ubuntu driver packages

davidmhewitt:
This greatly improves issue #975 by keeping a cache of compatible Ubuntu drivers. It takes a while to discover the drivers and associate them to Package objects so doing this each time the installed list needs to load causes it to show the placeholder message for much longer than it needs to.

There will still be a bit of a delay loading the list on a completely cold start, but if AppCenter is running silently in the background (like it normally is), the loading of the installed list should be pretty instant again after opening it.

### DIFF
--- a/src/Core/UbuntuDriversBackend.vala
+++ b/src/Core/UbuntuDriversBackend.vala
@@ -21,6 +21,8 @@ public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
 
     public bool working { public get; protected set; }
 
+    public Gee.TreeSet<Package>? cached_packages = null;
+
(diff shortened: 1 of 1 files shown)


src/Core/UbuntuDriversBackend.vala
```diff
@@ -21,6 +21,8 @@ public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
 
     public bool working { public get; protected set; }
 
+    public Gee.TreeSet<Package>? cached_packages = null;
```
donadigo: There's no need for this to be `public`.

Issue #249, opened 2017-04-28
Second line in Bookworm banner


danirabbit:
![screenshot from 2017-04-27 19 37 26](https://cloud.githubusercontent.com/assets/7277719/25512264/40f2fd84-2b81-11e7-988e-b6124255eb0b.png)

babluboy:
@danrabbit I think the spaces in the second line is really a bookworm app.xml issue. I have fixed the same and released into AppCenter. It's currently waiting for human review. Let me know if the issue still persists in the latest version of bookworm

cassidyjames:
We should still probably handle that more gracefully in AppCenter in case an app has malformed data again.

btkostner:
Adding to this, we should make sure that there are no lists in the banner, because they look pretty bad (ignore the line in the image).

![screenshot from 2017-08-01 14 26 20](https://user-images.githubusercontent.com/3385679/28848222-ffbd9cd4-76c6-11e7-9296-f197c1859ba8.png)

danirabbit:
Gonna close this because it's really old and current banner apps seem to be working fine

babluboy:
> Gonna close this because it's really old and current banner apps seem to be working fine

Thanks Dan - trying to move Bookworm to GTK4 and GNOME44 but struggling to modify the code as so many APIs have changed...hopefully will get the changes working and deploy to Appcenter soon.

PR #1013, opened 2019-05-31
AppInfoView: strip extra spaces and line breaks

danrabbit:
It seems common that developers assume that we'll only respect line breaks from `<p>` tags and that they can maintain their own indentation in their xml files. This makes sure we strip out line breaks and extra spaces before we pass markup to appstream to be parsed. This addresses issues like #249 but for the appinfoview instead of the banner


danrabbit:
@donadigo thanks for the suggestion! Your solution is much smarter than mine, as usual ;)

PR #1016, opened 2019-06-13
Rebase Pop!_Shop unto elementary/appcenter

jackpot51:
This is only for discussion of what we may merge upstream into AppCenter in order to keep the differences in Pop!_Shop smaller.

Here are the differences between the two:
- Add debian packaging, adapted from the AppCenter deb-packaging branch
- Replace AppCenter with Pop!_Shop
- Update links in documentation
- Add featured apps, known as Pop!_Picks, to the place where the recently updated/trending carousels are
- Modify the banner with Pop styling
- Change colors with Pop styling
- Set defaults for the configuration options (curated false, icon pop-shop, name Pop!_Shop, payments false, sharing false
- Add a new define, POP_OS, that when set does the following:
  - Applies banner changes in Homepage
  - Switches recently updated/trending carousels with featured carousel
  - Changes the Houston API URL
  - Adds a configuration button to the header bar for configuring sources with `repoman`
  - Updates some hard-coded colors

There are some things we may be able to do that will make it easier to maintain our fork, as well as be able to more regularly submit patches to upstream. First, would be moving some of the things we have to configure using the POP_OS define into meson configuration. For example, the Houston API URL. Next would be further modularizing the CSS so that components that we need to edit, like the badge, only require edits in a single, isolated file. It has been good to see such a change for catagories.css. It might also be possible to template some of the files we had to edit, like the appdata.xml file, so that the name comes from the meson configuration.

I am looking for guidance from the elementary OS team on what we might do to reduce our changeset from master. Already, our use of the POP_OS define has made it easier to rebase on AppCenter more regularly, as we no longer edit the AppCenter code, only disable and rewrite sections we need to function differently. Hopefully we can come up with ways to integrate many of these changes into the upstream codebase.


meson_options.txt
```diff
@@ -1,9 +1,9 @@
 option('blacklist_file', type : 'string', value : 'appcenter.blacklist', description : 'The name of the file listing blacklisted applications')
```
davidmhewitt: I appreciate that you're adding a new option here so you have to modify this file. But if you weren't, would it not make more sense to set the option values in `debian/rules` using debhelper rather than overriding the defaults. You'd be less likely to bump into conflicts that way since you're adding all the debian files to the master branch yourselves anyway.

src/Core/Houston.vala
```diff
@@ -17,7 +17,11 @@
 */
 
 public class AppCenterCore.Houston : Object {
+#if POP_OS
+    private const string HOUSTON_API_URL = "https://api.pop-os.org/shop/v1";
```
davidmhewitt: There are a lot of instances of changing constants in files throughout this diff. I'd be tempted to suggest making a separate `Constants.vala` file here in upstream. I think we'd be open to that with the justification of it making it much easier for downstreams to pull in changes. As that file would be unlikely to change, you'd be a lot less likely to have merge conflicts when pulling in upstream changes.

src/Widgets/AppContainers/InstalledPackageRowGrid.vala
```diff
@@ -56,6 +57,11 @@ public class AppCenter.Widgets.InstalledPackageRowGrid : AbstractPackageRowGrid
             return true;
         });
 
+        release_expander_label = new Gtk.Label ("");
```
davidmhewitt: I don't know the history behind the changes in this file or the exact effect they have, but if they're intended to fix some issue that's been discovered, I'm sure the changes would be accepted upstream...

src/Widgets/PackageRow.vala
```diff
@@ -23,6 +23,9 @@ namespace AppCenter.Widgets {
         AbstractPackageRowGrid grid;
 
         public PackageRow.installed (AppCenterCore.Package package, Gtk.SizeGroup? info_size_group, Gtk.SizeGroup? action_size_group, bool show_uninstall = true) {
+#if POP_OS
+            selectable = false;
```
davidmhewitt: Does the Pop!_OS theme have the right styling so you can see what row has keyboard focus? Otherwise, this will make keyboard navigation difficult.

Conan-Kudo:
I'd really rather not see Pop!_OS' packaging incorporated into the master branch...

src/MainWindow.vala
```diff
@@ -186,11 +186,37 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
         spinner = new Gtk.Spinner ();
 
+#if POP_OS
+        var repos_button = new Gtk.Button.from_icon_name ("preferences-system-symbolic", Gtk.IconSize.SMALL_TOOLBAR);
+        repos_button.tooltip_text = _("Edit Software Sources…");
+        repos_button.valign = Gtk.Align.CENTER;
+        repos_button.clicked.connect (() => {
+            try {
+                string[] args = {
+                  "/usr/lib/repoman/repoman.pkexec"
```
Conan-Kudo: I'm not a fan of this particular change. Generally speaking, AppCenter is designed to provide a default experience. If we were going to have software sources managed in AppCenter, I'd rather it be done through PackageKit and Flatpak APIs, rather than subprocessing out to some other program.

jackpot51:
@Conan-Kudo please read the description. This is not what we expect to merge upstream, but instead shows all of the differences between upstream and downstream so we can discuss what might be done upstream to reduce the set of changes.

We do not expect our packaging or the use of repoman to be upstreamed. We do expect things like a configurable houston URL, or more modularization of the CSS.

davidmhewitt:
Going to close this as there seems to have been some progress in rebasing in Pop Shop now. Please open individual issues for any aspects of the AppCenter codebase that could be improved that you feel could help this process in the future.

PR #1037, opened 2019-07-18
Add info icon and tooltip for download size

### LINKED ISSUE #1030: Update size is no longer helpful with Flatpak
Showing the (potential maximum) size of to-be-installed packages still can make sense, but showing the size for updates is almost always wrong, and wrong by an order of magnitude or more due to the way Flatpak does diff-based updates. 

I think we should either remove download size information for updates, or better expose that this is a maximum, worst-case-scenario estimate but that the true size is likely to be much, much, much lower. We discussed this at the Metered Data hackfest and as of then there was no way to get a more accurate download estimate, because the client and server negotiate the diff on demand afaik.

cassidyjames:
Fixes #1030 

- [x] Change from "Size: X" to "Up to X" on Updates page
- [x] Add info icon with tooltip
- [x] Only show when Flatpaks are involved
- [x] Also show on app listings
- [x] Fix expansion of SizeLabel causing weird tooltip placement

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -254,6 +254,12 @@ public class AppCenterCore.Package : Object {
         }
     }
 
+    public bool is_flatpak {
+        get {
+            return (backend is FlatpakBackend) || change_information.updatable_packages.contains (FlatpakBackend.get_default ());
+        }
+    }
+
     private string? _author = null;
     public string author {
         get {
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -256,7 +256,8 @@ public class AppCenterCore.Package : Object {
 
     public bool is_flatpak {
         get {
-            return (backend is FlatpakBackend) || change_information.updatable_packages.contains (FlatpakBackend.get_default ());
+            return (backend is FlatpakBackend)
+                || change_information.updatable_packages.contains (FlatpakBackend.get_default ());
         }
     }
 
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -40,7 +40,6 @@ namespace AppCenter.Views {
         private Gtk.Stack screenshot_stack;
         private Gtk.TextView app_description;
         private Widgets.Switcher screenshot_switcher;
-        private Gtk.Stack app_download_stack;
         private Gtk.ListStore version_liststore;
         private Gtk.ComboBox version_combo;
 
@@ -266,19 +265,11 @@ namespace AppCenter.Views {
             if (!package.is_local) {
                 size_label = new Widgets.SizeLabel ();
                 size_label.halign = Gtk.Align.END;
-                size_label.margin_end = open_button.margin_end;
                 size_label.valign = Gtk.Align.START;
 
                 action_button_group.add_widget (size_label);
 
-                /* We hide the label with a stack in order to stop the size requisition changing */
-                app_download_stack = new Gtk.Stack ();
-                // app_download_stack.margin_end = 6;
-                app_download_stack.add_named (size_label, "CHILD");
-                app_download_stack.add_named (new Gtk.EventBox (), "NONE");
-                app_download_stack.hhomogeneous = false;
-                app_download_stack.set_visible_child_name ("NONE");
-                header_grid.attach (app_download_stack, 3, 1, 1, 1);
+                header_grid.attach (size_label, 3, 1, 1, 2);
             }
 
             var header_box = new Gtk.Grid ();
@@ -462,8 +453,6 @@ namespace AppCenter.Views {
                 app_version.label = package.get_version ();
             }
 
-            app_download_stack.set_visible_child_name (package.state == AppCenterCore.Package.State.NOT_INSTALLED ?
-                                                       "CHILD" : "NONE");
             size_label.update ();
             if (package.state == AppCenterCore.Package.State.NOT_INSTALLED) {
                 get_app_download_size.begin ();
@@ -489,7 +478,6 @@ namespace AppCenter.Views {
             var size = yield package.get_download_size_including_deps ();
 
             size_label.update (size);
-            app_download_stack.set_visible_child_name ("CHILD");
         }
 
         public void view_entered () {
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -33,7 +33,7 @@ namespace AppCenter.Views {
         private Gtk.Label app_screenshot_not_found;
         private Gtk.Stack app_screenshots;
         private Gtk.Label app_version;
-        private Gtk.Label app_download_size_label;
+        private Widgets.SizeLabel size_label;
         private Gtk.ListBox extension_box;
         private Gtk.Grid release_grid;
         private Widgets.ReleaseListBox release_list_box;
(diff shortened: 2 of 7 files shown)


danrabbit:
Should we make sure we're only showing this messaging if the update contains flatpaks?

cassidyjames:
@danrabbit yes because you're smart :kissing_heart:

src/Widgets/SizeLabel.vala
```diff
@@ -0,0 +1,84 @@
...
+
+        public SizeLabel (uint64 _size = 0, bool _downloading_flatpak = false) {
+            Object (
+                downloading_flatpak: _downloading_flatpak,
+                size: _size
+            );
+        }
+
+        construct {
+            tooltip_markup = "<b>%s</b>\n%s".printf (
```
cassidyjames: @danrabbit thoughts on this copy and styling?

src/Views/AppInfoView.vala
```diff
@@ -488,8 +477,8 @@ namespace AppCenter.Views {
 
             var size = yield package.get_download_size_including_deps ();
 
-            app_download_size_label.label = GLib.format_size (size);
-            app_download_stack.set_visible_child_name ("CHILD");
+            // TODO: pass in bool whether or not using Flatpak as second param
```
cassidyjames: @davidmhewitt is this something we know in this context?

cassidyjames:
Screenshots time!

## Non-Flatpak app in this branch:

Screenshot from 2019-07-23 00 41 24

## Example of what a Flatpak app should look like:

Screenshot from 2019-07-23 00 42 21

Screenshot from 2019-07-23 00 42 42

## Update all (without Flatpaks):

Screenshot from 2019-07-23 00 44 29

## Update all (with Flatpaks):

Screenshot from 2019-07-23 00 45 21

Screenshot from 2019-07-23 00 45 40

cassidyjames:
Screenshots with updated copy and code.

## App Info

### Non-Flatpak

Screenshot from 2019-07-23 13 23 33

### Flatpak

Screenshot from 2019-07-23 13 23 42

Screenshot from 2019-07-23 13 24 01

## Updates

### Non-Flatpak

Screenshot from 2019-07-23 13 26 10

### Flatpak

Screenshot from 2019-07-23 13 24 19

Screenshot from 2019-07-23 13 24 39

src/Views/AppListUpdateView.vala
```diff
@@ -157,18 +157,25 @@ namespace AppCenter.Views {
...
                 uint64 update_real_size = 0ULL;
+                bool _check_flatpak = true;
+                bool using_flatpak = false;
                 foreach (var package in get_packages ()) {
                     if (package.update_available || package.is_updating) {
                         if (package.should_nag_update) {
                             nag_numbers++;
                         }
 
+                        if (_check_flatpak && package.is_flatpak) {
```
davidmhewitt: You don't need the extra `_check_flatpak` variable here, you can just use `!using_flatpak`

src/Core/Package.vala
```diff
@@ -254,6 +254,12 @@ public class AppCenterCore.Package : Object {
         }
     }
 
+    public bool is_flatpak {
+        get {
+            return (backend is FlatpakBackend) || change_information.updatable_packages.contains (FlatpakBackend.get_default ());
```
davidmhewitt: Tempting to split into two lines based on the length?

src/Views/AppListUpdateView.vala
cassidyjames: Oh duh, you're smart.

src/Core/Package.vala
cassidyjames: @davidmhewitt okay possibly dumb question, but: how would you prefer this line be split? There aren't wrapping parens or anything so it always looks awkward no matter how I split it. Also, other lines in this same file are longer, so I'm tempted to leave it…

davidmhewitt: Yeah, not the easiest one to split. It may be better to start with an if statement that checks `backend is FlatpakBackend` and returns true at that point if so. And then have the longer statement checking `updateable_packages` on its own line. But I'm fine to leave this either way.

cassidyjames:
I've tested the functionality pretty extensively. The easiest way is to:

- Add a couple of repos like Flathub and GNOME Nightly
- Install something from the nightly one (so you will get an update relatively quickly
- View an app page that exists in both AppCenter and Flathub

Then make sure:

- The app listing shows just the size (without "Up to", the info icon, and a tooltip) when its source is AppCenter
- The app listing shows the new "Up to" and everything when its source is Flathub
- Updates when you have some flatpak updates (i.e. the nightly SDK) show the new "Up to" stuff

Finally, to test non-Flatpak updates on this branch:

- Open a Terminal and `flatpak update` to get everything up to date
- Restart AppCenter, then check the updates page. The update size should be just the size (no "Up to" stuff)

src/Widgets/SizeLabel.vala
```diff
@@ -0,0 +1,83 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace AppCenter.Widgets {
```
danrabbit: You don't need to do this. You can do `public class AppCenter.Widgets.SizeLabel` and save yourself an extra indentation level

src/Widgets/SizeLabel.vala
```diff
@@ -0,0 +1,83 @@
...
+        public SizeLabel (uint64 _size = 0, bool _using_flatpak = false) {
+            Object (
+                size: _size,
+                using_flatpak: _using_flatpak
+            );
+        }
+
+        construct {
+            tooltip_markup = "<b>%s</b>\n%s".printf (
+                _("Actual download size likely to be smaller."),
```
danrabbit: Should this be title case maybe?

src/Widgets/SizeLabel.vala
cassidyjames: 👍 was just copying convention from all the other widget classes. 😅

src/Widgets/SizeLabel.vala
```diff
@@ -0,0 +1,83 @@
...
+            size_label.show ();
+
+            var icon = new Gtk.Image.from_icon_name ("dialog-information-symbolic", Gtk.IconSize.BUTTON);
+            icon.margin_start = 6;
+            icon.show ();
+
+            icon_revealer = new Gtk.Revealer ();
+            icon_revealer.transition_type = Gtk.RevealerTransitionType.NONE;
+            icon_revealer.add (icon);
+            icon_revealer.show ();
```
danrabbit: Do we need to do individual show on each widget? `show_all` doesn't work in this grid?

src/Widgets/SizeLabel.vala
cassidyjames: Yeah could do title case and no punctuation since it's title-like

src/Widgets/SizeLabel.vala
cassidyjames: I forget why I did that. There was a reason, but it might not be relevant anyone. I'll look into it.

src/Widgets/SizeLabel.vala
```diff
@@ -0,0 +1,83 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace AppCenter.Widgets {
+    public class SizeLabel : Gtk.Grid {
+        public bool using_flatpak { get; construct set; }
```
danrabbit: If you're expecting that this widget is changed using the `update` method, you probably don't want these to be `set`

src/Widgets/SizeLabel.vala
```diff
@@ -0,0 +1,83 @@
...
+
+            update (size, using_flatpak);
+        }
+
+        public void update (uint64 size = 0, bool using_flatpak = false) {
+            string human_size = GLib.format_size (size);
+
+            if (using_flatpak) {
+                size_label.label = _("Up to %s").printf (human_size);
+                has_tooltip = true;
```
danrabbit: You can probably bind these properties to `using_flatpak` instead of repeating yourself here

src/Widgets/SizeLabel.vala
cassidyjames: Ah, because I was using `show ()` before on individual pieces instead of the icon revealer, so it was just outdated. Addressed now.

PR #1038, opened 2019-07-24
added, modified some Arabic translations

abanoub-hanna:


peteruithoven:
We handle translations through Weblate. More info: https://elementary.io/get-involved#translations

PR #1045, opened 2019-07-26
Add friendly description to OS Updates

### LINKED ISSUE #1044: OS Updates page is very wall-of-text-y
We could add some friendly copy and hide the components behind an expander.

>## Updates to system components
>
>These core updates can improve security, increase stability, fix issues, and provide new features for other apps to use.
>
>▶  95 components with updates

cassidyjames:
Fixes #1044. Not sure if I'll tackle the expander in this branch or another, but I'm currently looking at it.

### DIFF
--- a/src/Core/UpdateManager.vala
+++ b/src/Core/UpdateManager.vala
@@ -37,6 +37,7 @@ public class AppCenterCore.UpdateManager : Object {
         os_updates_component.id = AppCenterCore.Package.OS_UPDATES_ID;
         os_updates_component.name = _("Operating System Updates");
         os_updates_component.summary = _("Updates to system components");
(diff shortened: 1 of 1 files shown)


danrabbit:
It looks like the description is force overwritten with the list of included packages, so more work might be needed here

cassidyjames:
@danrabbit yeah, that's what I was working on locally. I even just tried appending and it didn't work as expected. But if we want to put the components in an expander anyway, I'm wondering if we should just consider it a "What's New" changelog and reuse that.

danrabbit:
There’s actually an OS content type from appstream so I think the correct thing to do would be to start supplying that. So I’m gonna close this

Issue #1050, opened 2019-07-26
Picture loading spinner too small


D0J0P:
It works in the header bar, because the header bar is small, but it seems pretty small in proportion to the picture size of the app's screenshot. 
![Screenshot from 2019-07-26 18 55 24](https://user-images.githubusercontent.com/19495660/61985397-0ea47880-afd7-11e9-946c-0f4a28895fdf.png)

PR #1051, opened 2019-07-27
Animate screenshot background while loading

cassidyjames:
Inspired by #1050, using the sort of shine placeholder animation that's become common on the web and in apps like Slack. If we think it's clear enough that it's loading, we could even remove the spinner altogether.

### DIFF
--- a/data/styles/loading.css
+++ b/data/styles/loading.css
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2019 elementary, Inc. (https://elementary.io)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


data/styles/loading.css
```diff
@@ -0,0 +1,37 @@
...
+        to right,
+        @bg_color 20%,
+        white,
+        @bg_color 80%
+    );
+}
+
+@keyframes loading {
+    from {
+        background-position: -600px center;
```
cassidyjames: I tried with percentages and it seems GTK didn't like that.

cassidyjames:
Video of the effect: https://youtu.be/FNpZU_eaA10

data/styles/loading.css
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.loading {
+    animation: loading 1.5s ease-in infinite;
+    background-image: linear-gradient(
+        to right,
+        @bg_color 20%,
+        white,
```
danrabbit: For dark theme compatibility 
```suggestion
        @base_color,
```

cassidyjames: Ah you're smart. Thanks!

PR #1060, opened 2019-08-16
Mark & Close Stale Issues

danrabbit:


cassidyjames:
I have mixed feelings about auto-closing stale issues. It assumes we can **always** confirm or triage **all** issues within a certain amount of time—even if the user has provided all the necessary info, will it get auto-closed if we just don't get to it in time? I'm fine with auto-closing stale issues that have been marked as incomplete by a human though; is there a way to configure it to only run on issues that have been given a specific label, instead of just excluding labels?

PR #1069, opened 2019-09-04
Only show the latest supported NVIDIA driver package

### LINKED ISSUE #1068: Hide legacy NVIDIA driver if current is available
As done downstream https://github.com/pop-os/shop/pull/131/

mmstick:
Prevents confusion when the user is given multiple installable NVIDIA drivers as an option.

Closes #1068


src/Core/UbuntuDriversBackend.vala
```diff
@@ -66,23 +69,28 @@ public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
...
+            } else if (package_name.has_prefix ("nvidia-")) {
+                nvidia_version = package_name.offset (7);
+            }
 
-            var icon = new AppStream.Icon ();
-            icon.set_name ("application-x-firmware");
-            icon.set_kind (AppStream.IconKind.STOCK);
-            driver_component.add_icon (icon);
+            if (null != nvidia_version) {
+                if (-1 == GLib.strcmp (latest_nvidia_version, nvidia_version)) {
```
jlnr: I know it's unlikely to happen anytime soon, but this will break when the driver version number grows from three to four digits. I think it would be cleaner to convert the number to an integer, and compare that.

PR #1075, opened 2019-09-18
Hide non-curated non-Flatpak

cassidyjames:
Super experimental and probably heavy-handed, but it was something @danrabbit and I had talked about trying out locally on our installs.

The goal of this branch is to hide any non-curated non-Flatpak apps, making AppCenter more of a curated + Flatpak store. This should probably be done lower-level than where I'm doing it, though?


davidmhewitt:
If we are going to do this, then this probably isn't a great approach. We can save considerable amounts of processing power and time if we never construct these package rows in the first place.

We need to be careful to not exclude any OS updates and we need some thought about whether to still include non-curated, non-flatpak apps that are installed (i.e. will AppCenter still assist with updating **all** the apps on a user's PC?)

cassidyjames:
@davidmhewitt yeah that's kind of what I figured regarding just throwing the rows away; it felt like an unnecessary hammer but I mostly wanted to test the UX of not having those apps. 😁

I think it is reasonable to update all apps, but not display them in browsing/searching contexts.

danrabbit:
I'm gonna close this since it's old and conflicts etc, but I opened https://github.com/elementary/appcenter/issues/1432 to track the idea

PR #1083, opened 2019-09-24
Add Uninstall failed dialog

arshubham:
Ready for review.

Fixes https://github.com/elementary/applications-menu/issues/292
Fixes https://github.com/elementary/appcenter/issues/1084

To be merged with https://github.com/elementary/applications-menu/pull/300


arshubham:
@donadigo Thanks for the review. My bad, should have thought of it. Will make appropriate changes to the pr.

src/Dialogs/UninstallFailDialog.vala
```diff
@@ -0,0 +1,43 @@
...
+
+public class UninstallFailDialog : Granite.MessageDialog {
+    public AppCenterCore.Package package { get; construct; }
+    public Error error { get; construct; }
+
+    public UninstallFailDialog (AppCenterCore.Package package, Error error) {
+        Object (
+            title: "",
+            primary_text: _("Uninstall failed for %s").printf (package.get_name ()),
+            secondary_text: _("This may have been caused by external, manually added software repositories or a corrupted sources file."),
```
davidmhewitt: I'm not sure that an uninstall failing would be caused by "manually added software repositories or corrupted sources files". Those type of errors occur when we're trying to update the local package cache from the repos, but shouldn't happen when trying to uninstall something.

src/Services/DBusServer.vala
```diff
@@ -66,11 +66,28 @@ public class DBusServer : Object {
...
+            throw error;
+        }
+
+        if (uninstall_confirm_dialog.run () == Gtk.ResponseType.ACCEPT) {
+            package.uninstall.begin ((obj, res) => {
+                try {
+                    package.uninstall.end (res);
+                } catch (Error e) {
+                    // Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
+                    if (e.code != 303) {
```
davidmhewitt: I don't like this magic `303` number, can you do something more like:
```
if (e is GLib.IOError.CANCELLED) {
```

That might not be the right `CANCELLED` error type, but you get the idea.

src/Widgets/AppContainers/AbstractAppContainer.vala
```diff
@@ -453,10 +453,18 @@ namespace AppCenter {
...
-                MainWindow.installed_view.remove_app.begin (package);
-            }
+            package.uninstall.begin ((obj, res) => {
+                try {
+                    if (package.uninstall.end (res)) {
+                        MainWindow.installed_view.remove_app.begin (package);
+                    }
+                } catch (Error e) {
+                    // Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
+                    if (e.code != 303) {
```
davidmhewitt: Same again here.

src/Dialogs/UninstallFailDialog.vala
cassidyjames: @davidmhewitt I think that copy just came verbatim from the other error dialog. :smile: Do we know what might _actually_ cause a failure when uninstalling?

arshubham: @cassidyjames I copied it from the other dialog. Couldn't think of anything better. :smile:  The actual error is shown in the error expander, whenever available. The sub heading was supposed to be a generic description. I am open to suggestions here.

danrabbit:
@arshubham can you follow up with fixes for this branch? I'd like to get this merged soon :)

src/Dialogs/UninstallFailDialog.vala
arshubham: I have changes the string. Is it acceptable now?

src/Services/DBusServer.vala
arshubham: I am having issues with fixes it. The packagekit backend doesn't emit a GLib error here. It emits [Error](https://valadoc.org/packagekit-glib2/Pk.Error.html) class doesn't inherit GLib.Error. I think https://valadoc.org/packagekit-glib2/Pk.ErrorEnum.html used here to get the exact error. Need some guidance here on how to proceed

cassidyjames: @arshubham can't you do `if (e is Pk.ErrorEnum.NOT_AUTHORIZED) {` or whatever the correct error is here?

arshubham: No I get a `error: `Pk.ErrorEnum.NOT_AUTHORIZED' is not a type` for the above.

src/Dialogs/UninstallFailDialog.vala
```diff
@@ -0,0 +1,43 @@
...
+
+public class UninstallFailDialog : Granite.MessageDialog {
+    public AppCenterCore.Package package { get; construct; }
+    public Error error { get; construct; }
+
+    public UninstallFailDialog (AppCenterCore.Package package, Error error) {
+        Object (
+            title: "",
+            primary_text: _("Uninstall failed for %s").printf (package.get_name ()),
+            secondary_text: _("This may have been caused by external or manually compiled softwares."),
```
danrabbit: "software" is already plural :)

```suggestion
            secondary_text: _("This may have been caused by external or manually compiled software."),
```

src/Dialogs/UninstallFailDialog.vala
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+public class UninstallFailDialog : Granite.MessageDialog {
+    public AppCenterCore.Package package { get; construct; }
+    public Error error { get; construct; }
+
+    public UninstallFailDialog (AppCenterCore.Package package, Error error) {
+        Object (
+            title: "",
+            primary_text: _("Uninstall failed for %s").printf (package.get_name ()),
```
danrabbit: I wonder if it would be more natural to say

```suggestion
            primary_text: _("Failed to uninstall “%s”").printf (package.get_name ()),
```

src/Dialogs/UninstallConfirmDialog.vala
```diff
@@ -0,0 +1,41 @@
...
+        );
+    }
+
+    construct {
+        image_icon = package.get_icon (48, get_scale_factor ());
+
+        var uninstall_button = add_button (_("Uninstall"), Gtk.ResponseType.ACCEPT);
+        uninstall_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+
+        stick ();
```
danrabbit: I think we don't need to stick this. It was only the case in the applications menu because you could get stuck in a state where the applications menu was opened and couldn't be closed until this modal (on another desktop) was closed

src/Dialogs/UninstallFailDialog.vala
```diff
@@ -0,0 +1,43 @@
...
+        );
+    }
+
+    construct {
+        image_icon = package.get_icon (48, get_scale_factor ());
+
+        response.connect (() => destroy ());
+
+        show_error_details (error.message);
+        stick ();
```
danrabbit: I think we don't need to stick this. It was only the case in the applications menu because you could get stuck in a state where the applications menu was opened and couldn't be closed until this modal (on another desktop) was closed

src/Widgets/AppContainers/AbstractAppContainer.vala
```diff
@@ -453,10 +453,18 @@ namespace AppCenter {
...
-                MainWindow.installed_view.remove_app.begin (package);
-            }
+            package.uninstall.begin ((obj, res) => {
+                try {
+                    if (package.uninstall.end (res)) {
+                        MainWindow.installed_view.remove_app.begin (package);
+                    }
+                } catch (Error e) {
+                    // Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
+                    if (e.domain != Pk.ClientError.quark ()) {
```
davidmhewitt: This will disable the error dialog for most errors, I've looked into this and you need to do something more like this:
```
// Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
// Pk ErrorEnums are mapped to the error code at an offset of 0xFF (see packagekit-glib2/pk-client.h)
if (!(e is Pk.ClientError) || e.code != Pk.ErrorEnum.NOT_AUTHORIZED + 0xFF) {
```

src/Services/DBusServer.vala
```diff
@@ -66,11 +66,28 @@ public class DBusServer : Object {
...
+            throw error;
+        }
+
+        if (uninstall_confirm_dialog.run () == Gtk.ResponseType.ACCEPT) {
+            package.uninstall.begin ((obj, res) => {
+                try {
+                    package.uninstall.end (res);
+                } catch (Error e) {
+                    // Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
+                    if (e.domain != Pk.ClientError.quark ()) {
```
davidmhewitt: This will disable the error dialog for most errors, I've looked into this and you need to do something more like this:
```
// Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
// Pk ErrorEnums are mapped to the error code at an offset of 0xFF (see packagekit-glib2/pk-client.h)
if (!(e is Pk.ClientError) || e.code != Pk.ErrorEnum.NOT_AUTHORIZED + 0xFF) {
```

src/Dialogs/UninstallConfirmDialog.vala
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    construct {
+        if (package == null) {
+            primary_text = _("Uninstall app?");
+        } else {
+            primary_text = _("Uninstall “%s”?").printf (package.get_name ());
+        }
+
+        image_icon = package.get_icon (48, get_scale_factor ());
```
davidmhewitt: You still need to handle the case of `package == null` here.

src/Services/DBusServer.vala
```diff
@@ -66,11 +66,29 @@ public class DBusServer : Object {
...
+
+        var uninstall_confirm_dialog = new UninstallConfirmDialog (package);
+
+        if (uninstall_confirm_dialog.run () == Gtk.ResponseType.ACCEPT) {
+            package.uninstall.begin ((obj, res) => {
+                try {
+                    package.uninstall.end (res);
+                } catch (Error e) {
+                    // Disable error dialog for if user clicks cancel. Reason: Failed to obtain authentication
+                    if (!(e is Pk.ClientError) || e.code != Pk.ErrorEnum.NOT_AUTHORIZED + 0xFF) {
```
davidmhewitt: Could you add a comment about the 0xFF? I wrote one in my previous comments that you could use.

PR #1093, opened 2019-10-05
flatpak: Track overall transaction progress

### LINKED ISSUE #1026: Progress bar jumps all over the place with Flatpak updates
If we can't get something that is meaningful to the end user, should we just switch to a momentary progressbar for Flatpak?

davidmhewitt:
Fixes #1026 

Take one of the recent commits from Sideload and apply it here. Instead of the progress bar going from 0-100 for each operation within a transaction, divide it up and work out how we can go from 0-100 over the full transaction.


tintou:
Maybe the `current_operation - 1` part deserves some comment on why it's required

Issue #1097, opened 2019-10-11
Add runtime depends on Flatpak


cassidyjames:
It looks like our Flatpak-using apps depend on LibFlatpak, but nothing actually pulls in Flatpak itself. The result is that the XDG data dirs are not set up properly, and users can't manage their Flatpaks from Terminal. We should probably just include Flatpak in the base install now.

PR #1098, opened 2019-10-11
deb-packaging: Recommend flatpak

davidmhewitt:
Not a dependency, but we think it might make the flatpak support work better :wink:


danrabbit:
Fixes #1097

PR #1118, opened 2019-11-02
Unescape markup for package names

### LINKED ISSUE #1153: All plugins and indicators with a "&" in their name will show with "&amp;" when listed
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug

Screenshot

Screenshot

### Expected behavior
Should be just e.g. "Date & Time Settings" /"Date & Time Indicator"...
The language setting is English. In German there is often no translation yet for the indicators and plugins, right?

Please check what applies:
-->
- [x] I'm using the latest released stable version

davidmhewitt:
Fixes #1153 

Noticed with the new metadata that we're not getting for Switchboard plugs and Wingpanel indicators that we were starting to see a lot of `Date &amp; Time Settings` etc...

Not sure if there's a reason why we don't allow markup in the title's already or if this has just been overlooked so far.


tintou:
I don't know if it might just be better to use `GLib.Uri.unescape_string` rather than allowing any markup, I don't know if an app might use this to use Pango styling in the app name (which we probably don't want)

davidmhewitt:
@tintou don't merge this yet please, `GLib.Uri.unescape_string` doesn't unescape `&amp;`.

You're right that unescaping the string is probably better than allowing any markup, but I need to find the right tool for this.

davidmhewitt:
@tintou I've written a quick utility method that replaces the common XML entities as I can't find anything in the standard libraries that does this for us. Please let me know if you know of something that I can't find.

PR #1126, opened 2019-11-05
Remove dbus service file

### LINKED ISSUE #1063: AppCenter opens every time logging in
Hi!

In latest version available in daily ppa which configuration is starting App center automatically?

Previously this command worked to stop auto start.

```
cp /etc/xdg/autostart/io.elementary.appcenter-daemon.desktop ~/.config/autostart/
echo "X-GNOME-Autostart-enabled=false" | tee -a ~/.config/autostart/io.elementary.appcenter-daemon.desktop
```

It also doesn't worked now:

```
cd /etc/xdg/autostart/
sudo mv io.elementary.appcenter-daemon.desktop io.elementary.appcenter-daemon.desktop.bak
```
Best regards.

davidmhewitt:
Fixes #1063

This prevents AppCenter from being able to be launched by DBus.

Rationale:

- The AppCenter daemon autostarts anyway with its autostart file, and the applications menu watches the bus and connects/reconnects as necessary when it starts/restarts. We might opt to reduce the autostart delay down from 60 seconds, so it's ready for the applications menu sooner.
- Having this service file makes it even more difficult to stop AppCenter autostarting. Even if you remove the autostart file, it will still start every time you open the applications menu. There are people that like to disable the autostart for various reasons (bandwidth, CPU load, etc...)
- This is the cause of the AppCenter window opening on startup that's being reported a lot at the moment. If you've opened the applications menu before the 60 second delay on the autostart, gnome session manager will start it again. Of course it's a different issue that launching it twice with the silent flag still opens a window, but I haven't yet been able to fix that.

If we choose to merge this, we'll require a follow up in the `deb-packaging` branch to remove the service from the install file.

### DIFF
--- a/data/io.elementary.appcenter.service.in
+++ b/data/io.elementary.appcenter.service.in
@@ -1,3 +1,0 @@
-[D-BUS Service]
-Name=io.elementary.appcenter
-Exec=@EXEC_PATH@ -s
--- a/data/meson.build
+++ b/data/meson.build
@@ -6,14 +6,6 @@ configure_file(
     install_dir: join_paths(get_option('datadir'), 'applications')
 )
 
-# D-Bus Service
-configure_file(
-    input: meson.project_name() + '.service.in',
-    output: meson.project_name() + '.service',
-    configuration: conf_data,
-    install_dir: dbus.get_pkgconfig_variable('session_bus_services_dir', define_variable: ['datadir', get_option('datadir')])
-)
-
 # GNOME Shell Search Provider
 configure_file(
     input: meson.project_name() + '.search-provider.ini.in',


cassidyjames:
Just remember that this is already excluded in packaging, so this issue shouldn't appear for anyone who installs from packaging. It does make it annoying when you install from source, and if we're not using this in production anyway, I agree we should probably remove it. But I don't know that it'll fix the issue users are seeing, unless that many people install from git (which I doubt).

https://github.com/elementary/appcenter/blob/d0d308277139c0b6f480e5bd793e564aefae1334/debian/rules#L16

https://github.com/elementary/appcenter/commit/9541961b041639cada1027ebeb4e443c3fadfe6a

davidmhewitt:
After gathering a bit more info in https://github.com/elementary/appcenter/issues/1063 and upgrading a stable machine that's never had AppCenter built from source on it, this file is indeed installed from packaging and is the cause of AppCenter automatically opening on people's machines.

After closer inspection, the line that @cassidyjames refers to above does not exclude the file from being installed, it just stops debhelper complaining if we don't install the file. But it is installed as per this:
https://github.com/elementary/appcenter/blob/d0d308277139c0b6f480e5bd793e564aefae1334/debian/appcenter.install#L3

PR #1130, opened 2019-11-06
Workaround PackageKit issue to fix restart notification

### LINKED ISSUE #1048: Restart required infobar no longer shows
Can't seem to get it to show with `sudo touch /var/run/reboot-required`, and it doesn't show after kernel updates anymore.

davidmhewitt:
Fixes #1048 

The cause of the restart required notifications no longer working is a resurgence of this old issue: https://github.com/hughsie/PackageKit/issues/207

After spending (way too much time) looking into this, I've found the cause of the issue within `packagekit-glib2` itself. 

A `Pk.Control` object acts as a proxy for a subset of the PackageKit DBus methods. Constructing a `Pk.Control` object does not actually make a connection to the bus or connect to any of its signals. Upon using a method upon the `Pk.Control` object for the first time, a connection is made.

A vast majority of the methods in `Pk.Client` use one or more methods in `Pk.Control`, so if we use something in `Pk.Client` first, it's that which causes `Pk.Control` to connect to the bus. If we use a synchronous method in `Pk.Client`, this is just a sync wrapper around the async method. It creates a new thread context and then calls the async method, which in turn causes the connection to the bus and signals. 

As this connection to the bus and signals has happened in a different thread context, we never receive any signals from the `Pk.Control` object in our thread. So, if we create the `Pk.Control` object and call an async method on it before anything else has chance to make a synchronous call, the signals are connected in the right thread.

I'll have a think about what the proper solution in the library should be and make a pull request, but this workaround will help in the mean-time and is similar to the other workaround for this issue we were using before I refactored everything and forgot about it: https://github.com/elementary/appcenter/pull/421


vjr:
Works for me! 👍

davidmhewitt:
I've opened a PR on PackageKit that will solve this issue upstream and remove the need for this workaround:

https://github.com/hughsie/PackageKit/pull/352/files

I'll wait to see if I get any feedback on that PR before proposing the same patch to the `os-patches` repo. I'll leave this open for now in case we still want/need the workaround.

cassidyjames:
@tintou which approach makes the most sense to you: carrying a patch for PackageKit in os-patches, or working around it here in AppCenter?

tintou:
I'm ready to accept this PR but we might then forget about it when it requires an actual fix upstream

src/Core/PackageKitBackend.vala
```diff
@@ -123,6 +123,23 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
     }
 
     static construct {
+        _control = new Pk.Control ();
+
+        // Work around an issue where if we call any synchronous method on PackageKit,
+        // we never receive signals from Pk.Control.Calling this here causes packagekit-glib2 to
+        // connect to the PackageKit DBus and fire signals on the right thread
```
cassidyjames:
```suggestion
        // connect to the PackageKit DBus and fire signals on the right thread
        // See: https://github.com/hughsie/PackageKit/issues/207
```

PR #1139, opened 2019-11-16
Include dependencies in flatpak download size calculations

davidmhewitt:
Previously it seems we were only displaying the download size for the application itself on an info page and not including dependencies. Update sizes were vaguely right as we were including updates to runtimes under the "OS Updates" package already.

This now uses the same methods used in Sideload to calculate the download size of a flatpak app including its non-installed dependencies. There's some extra logic for updates so that we calculate the download size of updates to runtimes once under the "OS Updates" package instead of multiple times under each app.


davidmhewitt:
@danrabbit I guess it's highly unlikely if not impossible, I was just trying to keep the code consistent between sideload and here but could probably simplify it if you feel it's necessary.

PR #1145, opened 2019-11-21
Delete 99disable-appcenter-daemon

danrabbit:
Don't think we want this anymore


davidmhewitt:
I think this is possibly still needed. This is only built into the appcenter-casper package which is only present on the .iso. The daemon obviously isn't disabled in a normal install.

See the thread here for the original reasoning:
https://github.com/elementary/os/issues/11

PR #1146, opened 2019-11-26
Rules: remove debhelper missing override

danrabbit:


debian/rules
```diff
@@ -12,9 +12,6 @@
 %:
 	dh $@ 
 
-override_dh_missing:
-	dh_missing --fail-missing --exclude=appcenter.blacklist
```
tintou: We still want `dh_missing --fail-missing`

danrabbit: I was hoping this might fix the build errors from Houston :p It's complaining about a missing file even though it doesn't exist :grimacing:

PR #1184, opened 2020-01-24
Replace confusing 'version' terms with 'origin' terms

### LINKED ISSUE #1183: Package "versions" terminology is confusing
Right now we're using the term "version" to refer to the app version (i.e. version 12.3.4) as well as duplicate app IDs from different sources (i.e. GNOME Builder from Flathub and GNOME Builder from the Ubuntu repos).

The former makes sense to me, but I think we should rename the latter to "sources," "remotes," or something less confusing.

cassidyjames:
Fixes #1183

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -630,7 +630,7 @@ namespace AppCenter.Views {
 
             Gtk.TreeIter iter;
             uint count = 0;
-            foreach (var origin_package in package.origin_packages) {
+            foreach (unowned Package origin_package in package.origin_packages) {
                 origin_liststore.append (out iter);
                 origin_liststore.set (iter, 0, origin_package, 1, origin_package.origin_description);
                 if (origin_package == package) {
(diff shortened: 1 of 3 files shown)


danrabbit:
I guess reading this, for our purposes does it really make it clearer to call it an `origin_package` here instead of `origin` and `origins`?

This would make our code things like `foreach (var origin in package.origins)` etc which sounds more like what it does even though it's not technically perfectly true

src/Views/AppInfoView.vala
```diff
@@ -630,16 +630,16 @@ namespace AppCenter.Views {
 
             Gtk.TreeIter iter;
             uint count = 0;
-            foreach (var version in package.versions) {
-                version_liststore.append (out iter);
-                version_liststore.set (iter, 0, version, 1, version.origin_description);
-                if (version == package) {
-                    version_combo.set_active_iter (iter);
+            foreach (var origin_package in package.origin_packages) {
```
danrabbit: While we're here should this be
```suggestion
            foreach (unowned Package origin_package in package.origin_packages) {
```

cassidyjames:
@danrabbit that's what I had originally but it was similarly confusing. I think the more verbose but more correct terminology is advantageous because you immediately know it's a package and has all the properties and methods that implies.

danrabbit:
Okay yup good point

PR #1187, opened 2020-01-24
AppInfoView: Add a non-curated icon

### LINKED ISSUE #233: Expose non-curated app remote
I've heard from a couple folks that they want a way to know which apps were made for elementary OS. While we still have apps from the Ubuntu repos, perhaps we need a way to mark those apps so that people know that they are not from our repos.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44000145-mark-apps-from-the-ubuntu-repos?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #233 

Screenshot from 2020-01-24 11 55 17@2x


cassidyjames:
I guess the question now is: should these emblems be clickable buttons, informational with tooltips, etc.? Right now it feels a little weird to mix the two.

danrabbit:
I think that it's likely all of these will be non-interactive except for the tooltip. I don't think we'll necessarily have any additional place for users to go for information about flatpak holes or OARS data that couldn't be explained inline.

The license one is kind of an oddball because there's so much additional information and I'm not sure there's a way we can really provide any kind of useful tooltip. We've probably already provided all of the information that anyone who cares about this wants in the label. But we *can* provide a link and so we do.

Originally when we added the license to the footer I don't think we had it as a link and then we went "oh that's weird everything else has a link" and so we kind of went "Well we could find a way to display the whole license" and so we did.

src/Views/AppInfoView.vala
```diff
@@ -76,8 +76,105 @@ namespace AppCenter.Views {
...
+            license_button.tooltip_text = license_url;
+
+            unowned Gtk.StyleContext license_style_context = license_button.get_style_context ();
+            license_style_context.add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+            license_style_context.add_class (Gtk.STYLE_CLASS_FLAT);
+
+            var noncurated_grid = new Gtk.Grid ();
+            noncurated_grid.margin_start = 24;
+            noncurated_grid.row_spacing = 3;
+            noncurated_grid.tooltip_markup = "<b>%s</b>\n%s\n%s".printf (
```
cassidyjames: If we're planning to keep the dialog (which I think we should), we should be able to vastly simplify this copy down. How about something like:

>Has not been reviewed by elementary for security, privacy, or system integration

danrabbit: I agree sandboxing stuff can be split out, but I'd like to revisit that in a future branch and keep the messaging the same for now

cassidyjames: My issue is it feels like a *lot* to shove into a tooltip, especially when literally the same strings are gonna be shown again when they go to install. Is there a better way to simplify it here?

danrabbit: Is there anything we can do to make it acceptable to merge into master making sure that this copy matches the dialog copy?

My general feeling is that we're probably going to iterate here a bit as we use it and I don't want to add a bunch of translation work unless we're really sure we want to diverge on these strings

cassidyjames: @danrabbit we could just use the same string as the dialog title for now, then iterate on the dialog (possibly removing it) and this copy later.

cassidyjames:
I fixed the non-curated badge showing, but we probably don't want the license thing there, either since it looks like we don't know the license information.

src/Views/AppInfoView.vala
```diff
@@ -68,8 +68,101 @@ namespace AppCenter.Views {
...
+            license_button.tooltip_text = license_url;
+
+            unowned Gtk.StyleContext license_style_context = license_button.get_style_context ();
+            license_style_context.add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+            license_style_context.add_class (Gtk.STYLE_CLASS_FLAT);
+
+            var noncurated_grid = new Gtk.Grid ();
+            noncurated_grid.margin_start = 24;
+            noncurated_grid.row_spacing = 3;
+            noncurated_grid.tooltip_text = _("“%s” has not been reviewed by elementary for security, privacy, or system integration.").printf (package.get_name ());
```
cassidyjames: Should we pull this out into a separate string since it's repeated?

PR #1188, opened 2020-01-24
AppInfoView: Remove version from header

danrabbit:
* Version numbers are largely meaningless. Packages are assumed to be the latest version, whatever that is, unless we warn otherwise.
* If someone really needs to see the version number they can scroll down to "What's New" and see it there
* We already show the version number on the updates view where it might matter to someone

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -34,7 +34,6 @@ namespace AppCenter.Views {
 
         private Gtk.Label app_screenshot_not_found;
         private Gtk.Stack app_screenshots;
-        private Gtk.Label app_version;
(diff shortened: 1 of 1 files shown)


src/Views/AppInfoView.vala
```diff
@@ -351,7 +342,6 @@ namespace AppCenter.Views {
             header_grid.attach (package_name, 1, 0);
             header_grid.attach (package_author, 1, 1, 2);
             header_grid.attach (origin_combo_revealer, 1, 2, 2);
-            header_grid.attach (app_version, 2, 0);
```
cassidyjames: While you're here, the other grid attachments should be updated; the author and origin combo don't need to be two columns wide anymore.

PR #1203, opened 2020-02-07
Clear AppInfoView caches when appstream pools are reloaded

davidmhewitt:
Currently, each time an app info page is opened, the widget is put on a stack and never removed, acting as a kind of cache.

External events like updates and appstream cache changes means that this stack of cached views may no longer be valid. So, throw away these views when we can.

### DIFF
--- a/src/Views/View.vala
+++ b/src/Views/View.vala
@@ -37,6 +37,16 @@ public abstract class AppCenter.View : Gtk.Stack {
 
         previous_package = null;
 
+        if (AppCenterCore.BackendAggregator.get_default ().cache_flush_needed) {
+            foreach (var child in get_children ()) {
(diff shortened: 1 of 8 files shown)


src/Views/View.vala
```diff
@@ -26,6 +26,17 @@ public abstract class AppCenter.View : Gtk.Stack {
     construct {
         get_style_context ().add_class (Gtk.STYLE_CLASS_VIEW);
         expand = true;
+
+        notify["transition-running"].connect (() => {
+            // Transition finished
+            if (!transition_running) {
+                foreach (var child in get_children ()) {
```
tintou:
```suggestion
                foreach (weak Gtk.Widget child in get_children ()) {
```

PR #1206, opened 2020-02-11
Banner: Strip whitespace from description

### LINKED ISSUE #1205: Whitespace in AppData breaks banner formatting
This sort of formatting should be legal:

https://github.com/marbetschar/time-limit/blob/fcd5988398406b8d6754e35cec102f3d17e464f2/data/com.github.marbetschar.time-limit.appdata.xml.in#L10-L13

However, it results in this:

image

cassidyjames:
Fixes #1205 

Similar solution to here:

https://github.com/elementary/appcenter/blob/6217e655b89c6de63f5063767979ad8bcb55ce32/src/Views/AppInfoView.vala#L718-L731


src/Widgets/Banner.vala
```diff
@@ -61,7 +61,13 @@ namespace AppCenter.Widgets {
 
                 string description;
                 if (has_package) {
-                    description = package.get_description ();
+                    string[] lines = package.get_description ().split ("\n");
+                    description = lines[0].strip ();
```
vjr: Can the description be empty in rare cases? If so, that will need to be guarded against to avoid a crash?

danrabbit: If it's featured on the homepage it should definitely have a description since it would be reviewed

PR #1208, opened 2020-02-13
Run build and install in GitHub CI

davidmhewitt:
Should be the first step towards retiring travis in this repo.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,6 +3,27 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  build:
+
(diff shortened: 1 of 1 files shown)


btkostner:
@davidmhewitt I'm a bit concerned that this is installing on Ubuntu 18.04 without any elementary packages. It might not be a big deal now, but we have no guarantee this will work (get updated) when 20.04 gets released.

davidmhewitt:
@btkostner It's using the elementary docker image

btkostner:
@davidmhewitt whoops, missed that while reading :grimacing:

PR #1213, opened 2020-02-17
Fix sorting of native apps in list views

davidmhewitt:
If the compare function was comparing two elementary native applications, it would be kind of random as to which one would come out on top.

Ensure we use lower sorting criteria (alphabetical) if both apps are native.


davidmhewitt:
For testing, take the "Accessories" category for example. In `master`, note which application is at the top of the list and then scroll down enough that a few more groups of applications load in. Then go back up to the top of the list and take note of which application is first now.

Issue #1104, opened 2019-10-17
Grab screenshot thumbnail instead of full-resolution images


cassidyjames:
It looks like AppStream on the houston side generates screenshot thumbnails at various sizes. We could be smart and grab the largest thumbnail, at least on loDPI, to save some bandwidth and time.

If we wanted to get fancy, we could load a small thumbnail first, then crossfade to the higher resolution one to get _some_ content showing more quickly.

PR #1215, opened 2020-02-17
Choose a more appropriate screenshot size to download

davidmhewitt:
Instead of always downloading the largest (original) size image, save a bit of bandwidth (especially on LoDPI) by choosing the image that is closest to the width it will be displayed at.

This is related to #1104 but doesn't necessarily completely solve it.


danrabbit:
Nice work!

PR #1230, opened 2020-03-12
Prototype Account button and Popover

### LINKED ISSUE #1229: Prototype app history/library
Previously discussed either an additional tab, or more likely a menu under an account button in the headerbar

cassidyjames:
Just opening for visibility. Will fix #1229.

image

image


cassidyjames:
To do:

- [ ] Link to Online Accounts settings
- [ ] Switch to sync installed apps

cassidyjames:
It came up that the app history dialog itself should probably actually be provided by the Authenticator as well, since it would be useful on other platforms for reinstalling purchased apps. It should probably provide a NoShowIn=Pantheon .desktop file to get to it as well.

cassidyjames:
@danrabbit if this dialog is provided by the authenticator, I wonder if the "Sync installed apps" switch should go in the popover, then? Because on other platforms, I don't know that we'd be able to tell the authenticator to install apps or anything, I think that would be AppCenter's job.

PR #1240, opened 2020-03-31
arrow.css: Don't hardcode arrow color

cassidyjames:
Instead, use `@textColorPrimary` so it remains contrasty with the `@base_color`. Fixes unreadable arrows when using a dark style.

### DIFF
--- a/data/styles/arrow.css
+++ b/data/styles/arrow.css
@@ -26,7 +26,7 @@
         0 1px 2px rgba(0,0,0,0.24);
     margin: 6px;
     padding: 6px;
-    color: @textColorPrimary;
+    color: @text_color;
 }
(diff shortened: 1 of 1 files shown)


data/styles/arrow.css
```diff
@@ -26,7 +26,7 @@
         0 1px 2px rgba(0,0,0,0.24);
     margin: 6px;
     padding: 6px;
-    color: #333;
+    color: @textColorPrimary;
```
danrabbit: You don’t want to use textcolorprimary here because it mixes with the title bar color. Instead use `@text_color`

cassidyjames: ah derp

PR #1246, opened 2020-04-03
Split progress callbacks when an update has multiple backends

### LINKED ISSUE #1197: AppCenter crashes mid-update when there are both Flatpak and regular updates
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
When AppCenter shows available updates and among them are Flatpak updates, clicking on "Update All" makes AppCenter crash before the whole update is complete.

I only noticed this when there were Flatpak runtime updates amongst "Operating System Updates". I couldn't reproduce if this happens with Flatpak apps as well.

When clicking "Update All", no password is requested, which I guess means Flatpaks are being updated first. However, when the Flatpak updates are finished, AppCenter closes instead of requesting a password to update the "normal" packages. When I open it again, I can see that there are still remaining packages to update, and running "Update All" again works as intended, since there's no mix of Flatpak/regular packages anymore.

The only Flatpak remote I have installed is Flathub, added automatically by Sideload when running a .flatpakref file.

### To Reproduce
1. Open AppCenter's "Installed" tab
2. Check if updates contain both Flatpak and regular updates
3. Click on "Update All"
4. AppCenter will not request sudo password and will close after a while

### Expected behavior
I believe password should be requested immediatelly if there is a package update that requires it, even if Flatpak updates are done first. Also, AppCenter should not crash when updating.

## Platform Information

 - OS: elementary OS
 - OS Version : Hera
 - Hardware info: Dell XPS 9570 (i7-8750H, 16GB RAM, NVidia 1050ti)

- [x] I'm using the latest released stable version

davidmhewitt:
Fixes #1197 

It seems the issue with AppCenter crashing when updating Flatpaks and Debs was when both were in the OS Updates "package".

When the first backend (PackageKit or Flatpak) finished, it emitted a "finished" state causing AppCenter to start cleaning up, but the other backend was still trying to report progress.

This fixes that by dividing the progress, so we get a progress bar that's 50/50 between the PackageKit and Flatpak backend and only emits the finished state on the last operation.


cassidyjames:
I can confirm this fixes the crash which is an improvement. It looks like there are issues where after updating, not everything reloads correctly; the OS updates was still showing the PackageKit updates as available even after they had been installed, for example. But since this fixes the crash, it's probably worth getting in and then addressing the other issue(s) separately.

davidmhewitt:
@cassidyjames in the updates list view, does the os updates "package" go below the up to date header but still has information about updates when you click it? Or, is it still under the x packages with updates header?

cassidyjames:
@davidmhewitt dang, I didn't notice exactly where it showed up. It did say it had "x updates" and had an update button, and showed that update information when clicking it.

davidmhewitt:
I've just tried with another big batch of updates from the releases over the last couple of days and with a couple of smaller batches after downgrading some packages and I've been unable to reproduce the issue that @cassidyjames found. So I would agree that it's better to merge this now and fix the crash which we have tagged as a critical issue and then look out for the other issue afterwards.

PR #1260, opened 2020-04-10
Show loading placeholder unless no results returned

### LINKED ISSUE #975: Inappropriate message when Installed View first shown

jeremypw:
Fixes #975 

Needs design input on icon and wording of the loading placeholder.


jeremypw:
Temporary loading placeholder AlertView.
Screenshot from 2020-04-10 18 28 17

danrabbit:
Looks like according to the spec that `sync-synchronizing` is probably the best icon name: https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html

Proposed a branch here to add our own icons for that: https://github.com/elementary/icons/pull/879

danrabbit:
Maybe something like:

Screenshot from 2020-04-10 11 21 33@2x

I'm not sure if "package" is considered jargon here. @cassidyjames thoughts?

cassidyjames:
@danrabbit I think “Checking for Updates” as the header would be fine, even if that's not all it's doing—it's probably the most relevant and understandable bit.

Maybe something like:

>**Checking for Updates**
>Verifying installed apps plus checking updates to the OS and apps.

Still feels a little stilted. But "packages" does seem really jargony to me.

jeremypw:
@cassidy I am not sure "Verifying installed apps" is very meaningful to the average user.  How about "Checking for updates to the OS and installed apps"?

jeremypw:
I was concerned that the loading_view might appear during searching in which case the text would not be appropriate, but in fact, the the search view only appears after the search is complete so this is not a concern.  Have amended text and icon as per suggestions from UX team.

jeremypw:
Since merging master, this no longer works as expected - the installed view is populated immediately while checking for updates is ongoing.  Do we still need a message? If so where should it appear InfoBar?

PR #1262, opened 2020-04-13
Startup and loading optimisations

davidmhewitt:
I'll leave some inline comments about the motivation for some of these changes, but this should speed up both the initial opening of AppCenter and the time it takes to load the installed view.


src/Application.vala
```diff
@@ -203,7 +203,6 @@ public class AppCenter.App : Gtk.Application {
         base.dbus_register (connection, object_path);
 
         if (silent) {
-            DBusServer.init ();
```
davidmhewitt, line 206: The only thing in this method is a synchronous loading of all the installed applications. As far as I can tell, this is just so we don't get errors when we try to uninstall something that isn't installed with the DBus server. I've moved the installed check to just before uninstalling an app to prevent this.

src/Core/Client.vala
```diff
@@ -144,8 +144,7 @@ public class AppCenterCore.Client : Object {
                 refresh_in_progress = true;
                 updating_cache = true;
                 try {
-                    success = yield PackageKitBackend.get_default ().refresh_cache (cancellable);
-                    yield FlatpakBackend.get_default ().refresh_cache (cancellable);
+                    success = yield BackendAggregator.get_default ().refresh_cache (cancellable);
```
davidmhewitt, line 147: Was previously doing this two backends separately because the Ubuntu drivers one didn't really have a concept of `refresh_cache`, but it now has caching, so may as well call `refresh_cache` on all backends now.

src/Core/FlatpakBackend.vala
```diff
@@ -72,6 +72,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
     construct {
         worker_thread = new Thread<bool> ("flatpak-worker", worker_func);
         appstream_pool = new AppStream.Pool ();
+        appstream_pool.set_flags (AppStream.PoolFlags.READ_COLLECTION);
```
davidmhewitt, line 75: The Flatpak backend's appstream pool was parsing a bunch of stuff it didn't need to (like desktop files, etc...), so we limit this with the flags.

src/Core/PackageKitBackend.vala
```diff
@@ -245,20 +245,16 @@ public class AppCenterCore.PackageKitBackend : Backend, Object {
 
             var package = package_list[pk_package.get_name ()];
             if (package != null) {
-                populate_package (package, pk_package);
+                package.mark_installed ();
+                package.latest_version = pk_package.get_version ();
+                package.update_state ();
```
davidmhewitt, line 250: Just moving a single use method inline.

src/Core/UbuntuDriversBackend.vala
```diff
@@ -47,15 +47,10 @@ public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
...
         working = true;
+
         cached_packages = new Gee.TreeSet<Package> ();
-        string? command_output;
-        var result = yield get_drivers_output (cancellable, out command_output);
-        if (!result || command_output == null || cancellable.is_cancelled ()) {
-            working = false;
-            return cached_packages;
-        }
+        var tokens = AppCenter.App.settings.get_strv ("cached-drivers");
```
davidmhewitt, line 52: The `ubuntu-drivers` command takes a long time to run, so just lookup the driver package names we used last time from a GSettings cache.

src/Core/UbuntuDriversBackend.vala
```diff
@@ -130,14 +125,28 @@ public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
     }
 
     public async bool refresh_cache (Cancellable? cancellable) throws GLib.Error {
```
davidmhewitt, line 127: When we refresh the caches of the other backends, also run the `ubuntu-drivers` command to detect any hardware changes or new drivers and store the result in GSettings.

src/MainWindow.vala
```diff
@@ -91,8 +91,6 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
 
         return_button.clicked.connect (view_return);
 
-        installed_view.get_apps.begin ();
```
davidmhewitt, line 94: This is called in the constructor of the installed view, no need to call it again in this constructor.

src/Views/InstalledView.vala
```diff
@@ -67,6 +67,8 @@ public class AppCenter.Views.InstalledView : View {
 
         yield refresh_mutex.lock ();
 
+        refresh_cancellable.reset ();
+
```
davidmhewitt, line 71: We cancel the cancellable at the start of this method to try and get the previous operation to finish earlier so we can acquire the lock. So we need to reset the cancellable before running anything that checks it, or else we cancel the operation before it even starts.

src/Core/UbuntuDriversBackend.vala
```diff
@@ -130,14 +125,28 @@ public class AppCenterCore.UbuntuDriversBackend : Backend, Object {
...
+        working = true;
+        string? command_output;
+        var result = yield get_drivers_output (cancellable, out command_output);
+        if (!result || command_output == null || cancellable.is_cancelled ()) {
+            working = false;
+            return false;
+        }
+
+        string[] tokens = command_output.split ("\n");
```
donadigo, line 136: You should filter out empty tokens from here, so that no empty strings get saved to settings. These empty strings are right now only checked when `get_installed_applications` is called.

src/Core/Package.vala
```diff
@@ -379,7 +379,6 @@ public class AppCenterCore.Package : Object {
         color_primary_text = null;
         payments_key = null;
         suggested_amount = null;
-        _latest_version = null;
```
davidmhewitt, line 382: This methods resets all of the cached data on the package when we reload the AppStream data. Since the latest version string doesn't come from AppData and instead comes from the deb package, we don't need to drop this cache.

src/Views/InstalledView.vala
```diff
@@ -67,22 +67,12 @@ public class AppCenter.Views.InstalledView : View {
...
 
         var installed_apps = yield client.get_installed_applications (refresh_cancellable);
 
-        // To gain some speedup of the installed applications list, update the state of the packages ahead
-        // of adding them to the list. Otherwise, AppCenter will query each one in turn to get the latest
-        // verison string from PackageKit. The update_multiple_package state method ignores packages that
-        // aren't PackageKit packages, so this is safe to run on the whole list, even if it includes Flatpaks
-        if (AppCenterCore.PackageKitBackend.supports_parallel_package_queries && installed_apps != null) {
-            try {
-                yield AppCenterCore.PackageKitBackend.get_default ().update_multiple_package_state (installed_apps);
```
davidmhewitt, line 80: We can drop this call to update the state of all of the packages, as the `get_installed_applications` method already sets all of the required fields on the package.

danrabbit:
Looks like the dbus server for the applications menu get initialized just fine in this branch :)

PR #1265, opened 2020-04-16
Only show plugins in installed view if they require updates

### LINKED ISSUE #929: Better distinguish plugins in list views
Right now plugins just show their name in list views, like in updates. I think it'd be nice to show the app name in a dim label next to the plugin name, or something to better associate them.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67852596-better-distinguish-plugins-in-list-views?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #929 

As discussed on the standup call yesterday, this hides plugins from all search/category lists and only shows plugins in the installed/updates list if they require an update.


src/Views/AppListView.vala
```diff
@@ -40,20 +40,30 @@ namespace AppCenter.Views {
...
-            list_store.insert_sorted (package, (GLib.CompareDataFunc<AppCenterCore.Package>) compare_packages);
+            add_row_for_package (package);
+
             if (current_visible_index < 20) {
                 show_more_apps ();
             }
         }
 
+        private void add_row_for_package (AppCenterCore.Package package) {
+            // Don't show plugins in search and category views
```
cassidyjames, line 61: This is probably fine for now, but I think @danrabbit mentioned that better logic might be:

If the plugin has the same author as the original package, don't show it (because it's probably considered part of the app). But if the author of the plugin is different than the author of the app, show it so that third-party plugins remain searchable.

PR #1268, opened 2020-04-17
Add meson option to disable packagekit autoremove

### LINKED ISSUE #860: removing applications fails if backend doesn't support autoremove
Attempting to uninstall a package results in this log message, if the backend doesn't support autoremove:

```
Package.vala:351: Operation failed for package Notejot - autoremove is not supported
```

The GUI silently fails to remove the app and simply does nothing. Maybe appcenter should fall back to attempting the removal without the `autoremove` option, or detect if the backend supports it?

davidmhewitt:
Fixes #860 

Some packagekit backends don't have autoremove implemented, but there is no way to detect this at runtime. Trying to uninstall with autoremove on a backend that doesn't support it just outright fails.

This has no effect on elementary whatsoever, it just gives packagers the option to disable this functionality on their distribution if they wish and make AppCenter a little more functional elsewhere.


tintou:
Another workaround would be to retry the deletion without autoremove when the returned error is `Pk.ErrorEnum.NOT_SUPPORTED`

tintou:
But the DNF backend shouldn't hard error on this

davidmhewitt:
Yeah, I did consider checking for the `NOT_SUPPORTED` enum but that could be thrown for other reasons or by other backends for different reasons. So I don't really want to start adding loads of special casing to things that don't affect elementary anyway.

This seems like the most efficient way to allow other distributions to package this if they wish.

jeremypw:
@davidmhewitt Converting to draft as there are conflicts and no recent activity. Please fix and re-request a review if this is still relevant.

PR #1272, opened 2020-04-18
Prevent the daemon refreshing on every startup

davidmhewitt:
Store the last refresh time in GSettings and only refresh the cache if it's been >24 hours since the last refresh. Opening the window still forces a refresh of the cache, but now at least if you start your computer multiple times in a 24 hour window, AppCenter isn't doing a cache refresh every time.

We still notify that updates are available on startup without refreshing the cache if there are some leftover from the last cache update.

I removed an unused `updating_cache` property while I was here and cleaned up some of the overly complicated setting of `refresh_in_progress`

Finally, I've ensured the next refresh gets scheduled from 24 hours after the last refresh, rather than 24 hours from when AppCenter started. Meaning if your computer starts 23 hours after the last cache refresh, AppCenter will run the next cache update in an hour and notify of any new updates.

### DIFF
--- a/data/io.elementary.appcenter.appdata.xml.in
+++ b/data/io.elementary.appcenter.appdata.xml.in
@@ -19,6 +19,7 @@
       <description>
         <p>Fixes</p>
         <ul>
+          <li>Only check for updates at computer startup if it's more than 24 hours since we last checked</li>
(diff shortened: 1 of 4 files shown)


danrabbit:
I can confirm that I get the new updates notification a full minute faster on my second restart. Nice!

PR #1286, opened 2020-04-23
Improve installation failure dialog

davidmhewitt:
Replace the dialog that appears when an app fails to install that looks like this:

With one that looks like this:

image


danrabbit:
Funnily enough I can't get the dialog to show that error! I get `** (io.elementary.appcenter:6259): WARNING **: 14:53:09.301: Package.vala:479: Operation failed for package 0install - Failed to obtain authentication.` in Terminal but no dialog

PR #1289, opened 2020-04-24
Prefer dark

danrabbit:


cassidyjames:
@danrabbit I've updated this to use the Granite util, and tweaked a few of the category styles to look less flat in the dark style. Further improvements to the categories could potentially be made, but this feels like a good start to me.

danrabbit:
If we're planning to do any more 5.x releases, this makes the checkerboard in graphics pretty washed out looking:

Screenshot from 2020-10-12 14 21 20

But everything seems to look fine to me in Greenfield

cassidyjames (approved):

PR #1296, opened 2020-04-30
Cache Houston API results to GSettings

### LINKED ISSUE #210: Cache banners [$80]

### LINKED ISSUE #476: Make API URL overridable with build flag
Right now houston's API URL is hardcoded in AppCenter. We should pull this out to let it be settable with a build flag, which would help with new feature development (more easily point at  development API server) as well as third parties who would want to use AppCenter (i.e. System76 and Pop!_OS).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49256890-make-api-url-overridable-with-build-flag?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #476
Fixes #210 

Make the API base URL configurable via meson so this can be changed in packaging by downstreams (like System76) if desired.

On a successful fetch of app lists from the API, store the result in GSettings so if we have any subsequent failures, we can use the cached list instead.

I've opted to use a GSettings dictionary rather than specific settings for e.g. most downloaded, most recent. The key used in the dictionary is the API endpoint path, so again if these are changed downstream, the caching will still work.

I've put some blank dictionary entries in the default field of the schema file so there's an example of the syntax that will help downstreams write a list of apps in here manually if they wish.

Edit: As another note, it would have been simpler in terms of code to use a `VariantDict` here, but it only supports the `a{sv}` Variant type, rather than `a{sas}`. The former is a lot harder to write a `<default>` key for by hand as you need to use `<` `>` characters in the XML.

### DIFF
--- a/src/Core/Houston.vala
+++ b/src/Core/Houston.vala
@@ -86,7 +86,7 @@ public class AppCenterCore.Houston : Object {
                         // Get the dict key of this existing item
                         existing_item.@get ("{sas}", out key, null);
 
-                        // Don't add this item to the new dict if its the one we're replacing
+                        // Copy existing items to new dict, unless this is the item we're updating
                         if (key != endpoint) {
                             dict_builder.add_value (existing_item);
                         }
--- a/src/Core/Houston.vala
+++ b/src/Core/Houston.vala
@@ -66,9 +66,13 @@ public class AppCenterCore.Houston : Object {
                 if (res.has_member ("data")) {
                     var data = res.get_array_member ("data");
 
-                    var arr_builder = new VariantBuilder (new VariantType ("as"));
-                    foreach (var id in data.get_elements ()) {
-                        var val = (string)id.get_value ();
+                    var arr_builder = new VariantBuilder (GLib.VariantType.STRING_ARRAY);
+                    foreach (unowned Json.Node id in data.get_elements ()) {
+                        unowned string? val = id.get_string ();
+                        if (val == null) {
+                            continue;
+                        }
+
                         arr_builder.add ("s", val);
                         app_ids += val;
                     }
@@ -104,7 +108,7 @@ public class AppCenterCore.Houston : Object {
 
                 var caches = caches_store.get_value ("api-caches");
                 // Get the array of app IDs corresponding to this API endpoint from the dictionary
-                var cached_ids = caches.lookup_value (endpoint, new VariantType ("as"));
+                var cached_ids = caches.lookup_value (endpoint, GLib.VariantType.STRING_ARRAY);
 
                 if (cached_ids != null) {
                     app_ids = cached_ids.get_strv ();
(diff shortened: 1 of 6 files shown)


src/Core/Houston.vala
```diff
@@ -62,12 +66,49 @@ public class AppCenterCore.Houston : Object {
                 if (res.has_member ("data")) {
                     var data = res.get_array_member ("data");
 
+                    var arr_builder = new VariantBuilder (new VariantType ("as"));
```
tintou, line 69:
```suggestion
                    var arr_builder = new VariantBuilder (GLib.VariantType.STRING_ARRAY);
```

src/Core/Houston.vala
```diff
@@ -62,12 +66,49 @@ public class AppCenterCore.Houston : Object {
...
                     foreach (var id in data.get_elements ()) {
```
tintou, line 70: While you're here :)
```suggestion
                    foreach (unowned Json.Node id in data.get_elements ()) {
```

src/Core/Houston.vala
```diff
@@ -62,12 +66,49 @@ public class AppCenterCore.Houston : Object {
...
-                        app_ids += ((string) id.get_value ());
+                        var val = (string)id.get_value ();
```
tintou, line 71:
```suggestion
                        unowned string? val = id.get_string ();
```

src/Core/Houston.vala
```diff
@@ -62,12 +66,49 @@ public class AppCenterCore.Houston : Object {
...
+                    dict_builder.add ("{sas}", endpoint, arr_builder);
+
+                    caches_store.set_value ("api-caches", dict_builder.end ());
                 }
             } catch (Error e) {
                 warning ("Houston: %s", e.message);
+
+                var caches = caches_store.get_value ("api-caches");
+                // Get the array of app IDs corresponding to this API endpoint from the dictionary
+                var cached_ids = caches.lookup_value (endpoint, new VariantType ("as"));
```
tintou, line 107:
```suggestion
                var cached_ids = caches.lookup_value (endpoint, GLib.VariantType.STRING_ARRAY);
```

tintou:
Looks good besides my tiny nitpicks

Issue #1279, opened 2020-04-22
Better search ordering/ranking


cassidyjames:
Right now it looks like we list apps alphabetically in search results, regardless of their relevance to the search. Performing the same search with `appstreamcli` seems to return better ordering, so if we can use that, we probably should—otherwise (or if we find it produces better results), we should order results by:

>- App names starting with the term
>- App names containing the term
>- Everything else

…as mentioned by @WatchMkr in https://github.com/pop-os/shop/issues/171

cassidyjames:
I would also add matches from the app's summary line, as those appear in the search view. So, if possible:

- Human-readable app names starting with the term
- Human-readable app names containing the term
- Summary containing the term
- Descriptions containing the term (i.e. everything else)

I'm not sure if AppStream also does matching on the app ID or not, but if so, we might want to figure out where that fits in. But my suspicion is that if an app ID has a search term in it, the human-readable name will as well.

davidmhewitt:
Ah, yes, I'm not sure if this is a regression from when we put Flatpak support in or if it's always been like this, but it's definitely more difficult to solve now.

AppStream supposedly gives us search results in relevance order, which is probably the order you're seeing in `appstreamcli`.

But, now that we have Flatpak support, we technically do two searches in two AppStream pools (PackageKit's and Flatpak's) and merge the result, then sort them. As far as I remember this was done because when implementing it as a single pool, AppStream would only allow unique IDs in the pool, so if you had duplicates between Flatpak/PackageKit, you'd only get one of them.

Because of this, we can't just take the two results from AppStream and append one onto the other (keeping the relevance order) because then we'd be demoting either Flatpak or PackageKit packages to the bottom of the list.

I also suspect that trying to do some/all of the string processing necessary to sort the results in the order you detail above will be too inefficient to do at any great speed. I'll try it out as an experiment, but even if it's reasonable on my machine, it'll need testing on slower hardware, because multiple substring searches over hundreds/thousands of components are likely to be crippling.

I'll open an issue in AppStream to see if it would make sense for the sort score per component to be exposed as public API so we could try and use that to sort/merge the components.

As an aside, I had a look at how gnome-software fixed this and from what I can tell, they have a custom XML library ([libxmlb](https://github.com/hughsie/libxmlb)] that's optimized for searching for strings within the huge AppStream collection XMLs and implements all of the word stemming stuff that AppStream search does too. So it looks like they don't use the search methods built into AppStream at all and have implemented their own search.

[...]

PR #1301, opened 2020-05-06
Sort search view using search term

jackpot51:
Search order is defined by the search_priority function. This takes the lowercase application name and lowercase search term, and sorts with the following order:

- Name starts with term
- Name contains term
- Name does not contain term

Fixes https://github.com/pop-os/shop/issues/171 and is a good start on https://github.com/elementary/appcenter/issues/1279

### DIFF
--- a/src/Views/AppListView.vala
+++ b/src/Views/AppListView.vala
@@ -61,7 +61,10 @@ namespace AppCenter.Views {
         private void add_row_for_package (AppCenterCore.Package package) {
             // Don't show plugins or fonts in search and category views
             if (!package.is_plugin && !package.is_font) {
-                list_store.insert_sorted (package, (GLib.CompareDataFunc<AppCenterCore.Package>) compare_packages);
+                GLib.CompareDataFunc<AppCenterCore.Package> sort_fn = (a, b) => {
+                    return compare_packages(a, b);
+                };
+                list_store.insert_sorted (package, sort_fn);
             }
         }
 
@@ -92,22 +95,6 @@ namespace AppCenter.Views {
             on_list_changed ();
         }
 
-        private static int compare_packages (AppCenterCore.Package p1, AppCenterCore.Package p2) {
-#if CURATED
-            bool p1_is_elementary_native = p1.is_native;
-
-            if (p1_is_elementary_native != p2.is_native) {
-                return p1_is_elementary_native ? -1 : 1;
-            }
-#endif
-
-            if (p1.is_plugin != p2.is_plugin) {
-                return p1.is_plugin ? 1 : -1;
-            }
-
-            return p1.get_name ().collate (p2.get_name ());
-        }
-
         private int search_priority (string name) {
             if (name != null && current_search_term != null) {
                 var name_lower = name.down ();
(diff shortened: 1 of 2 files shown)


cassidyjames:
@jackpot51 thanks for this. How does this affect performance in your experience? If it's better and minimally affects performance, I'd be happy to get the improvement merged in until we can further improve results using AppStream's ranking.

src/Views/AppListView.vala
```diff
@@ -119,10 +134,18 @@ namespace AppCenter.Views {
             if (p1_is_plugin != row2.get_package ().is_plugin) {
                 return p1_is_plugin ? 1 : -1;
             }
+#endif
+
+            int sp1 = search_priority(row1.get_name_label ());
+            int sp2 = search_priority(row2.get_name_label ());
```
cassidyjames, line 140: Lint nit pick

```suggestion
            int sp1 = search_priority (row1.get_name_label ());
            int sp2 = search_priority (row2.get_name_label ());
```

jackpot51:
I haven't seen a negative performance impact. But this is only a small part of the whole thing. If you search summaries and descriptions that might change, especially having to lowercase them.

Might be good to use a case insensitive search but I wasn't sure how to do that in vala.

jackpot51:
I think I see why that's the case. I'll update the PR soon

jackpot51:
I pushed another commit that will make the sorting of the list_store and the sorting of the list_box match. That should fix the problem with scrolling.

It might fix the non-curated header too, I just need to test this on elementary to see.

src/Views/AppListView.vala
```diff
@@ -60,13 +61,17 @@ namespace AppCenter.Views {
         private void add_row_for_package (AppCenterCore.Package package) {
             // Don't show plugins or fonts in search and category views
             if (!package.is_plugin && !package.is_font) {
-                list_store.insert_sorted (package, (GLib.CompareDataFunc<AppCenterCore.Package>) compare_packages);
+                GLib.CompareDataFunc<AppCenterCore.Package> sort_fn = (a, b) => {
```
jackpot51, line 64: The reason for this lambda is I had to make compare_packages non-static so it can see the current_search_term. Attempting to cast a class's member function to GLib.CompareDataFunc<AppCenterCore.Package> makes it crash. Wrapping it in a lambda makes it safe to call the member function.

src/Views/AppListView.vala
```diff
@@ -103,26 +121,21 @@ namespace AppCenter.Views {
...
-            if (p1_is_elementary_native != row2.get_package ().is_native) {
-                return p1_is_elementary_native ? -1 : 1;
-            }
-
-            if (p1_is_plugin != row2.get_package ().is_plugin) {
-                return p1_is_plugin ? 1 : -1;
-            }
-
-            return base.package_row_compare (row1, row2);
+            return compare_packages (row1.get_package (), row2.get_package ());
```
jackpot51, line 135: This code removal should probably have happened regardless of having the search sorting added. Without it, the list_store and list_box could be sorted differently - leading to resorting when more results are shown.

src/Views/AppListView.vala
```diff
@@ -60,13 +61,17 @@ namespace AppCenter.Views {
         private void add_row_for_package (AppCenterCore.Package package) {
             // Don't show plugins or fonts in search and category views
             if (!package.is_plugin && !package.is_font) {
-                list_store.insert_sorted (package, (GLib.CompareDataFunc<AppCenterCore.Package>) compare_packages);
+                GLib.CompareDataFunc<AppCenterCore.Package> sort_fn = (a, b) => {
+                    return compare_packages(a, b);
+                };
+                list_store.insert_sorted (package, sort_fn);
```
davidmhewitt, line 67:
```suggestion
                    return compare_packages (a, b);
                };

                list_store.insert_sorted (package, sort_fn);
```

cassidyjames:
@davidmhewitt 👍 I'll do another sanity check on Hera

PR #1308, opened 2020-05-10
Support systemwide flatpaks

### LINKED ISSUE #1029: Prefer installed version of app
If there's an app with a repo version and a Flatpak version (GIMP, for example) and the Flatpak is installed, AppCenter should display the Flatpak version by default when viewing an app's page. 

Similarly, in list views, the installed version should be preferred; so in the above example, GIMP would have an "Uninstall" button on it instead of an "Install" button.

### LINKED ISSUE #1082: Prefer Flatpak
We should probably prefer Flatpak versions of apps over uncurated repo versions of apps, i.e. when showing search results of apps and two apps have the same ID. This assumes neither are installed; if one is installed, I think #1029 still makes sense.

### LINKED ISSUE #1115: Also display applications from systemwide Flatpak remotes
Currently, due to the decision we made that AppCenter would only handle flatpaks for the user, rather than systemwide, we only display applications configured in user remotes.

The default/old instructions on the flathub website for adding the flathub remote adds it systemwide by default:
`flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo`

Therefore, anyone who has configured flathub in that way won't see flathub applications in AppCenter. However, the first time you install a flathub application with Sideload, the flathub remote would be added for the user and make it appear in AppCenter too.

I think potentially things could get quite complex if we include systemwide remotes too. I guess we'd have to de-duplicate them and prioritise any user ones. Then any remotes that are configured for the system but not the user would be included too.

I'm also not sure what happens if you instruct flatpak to install an application for the user from a systemwide remote. I imagine it works fine, but this will also need testing.

davidmhewitt:
Fixes #1115 

Supporting systemwide remotes required quite a bit of refactoring of the Flatpak backend, so this is quite a big PR, though I broke out as much as I could into separate PRs that have already been merged.

Systemwide Flatpak remotes should now show as an extra origin and it should be possible to install/update/remove from them. You don't get a password prompt for doing any of this on an admin account because the default Flatpak config automatically permits this if you're a member of the wheel group (which admins are).


src/Core/Package.vala
```diff
@@ -341,6 +336,11 @@ public class AppCenterCore.Package : Object {
                     return _("Ubuntu (non-curated)");
                 }
             } else if (backend is FlatpakBackend) {
+                var fp_package = this as FlatpakPackage;
+                if (fp_package != null && fp_package.installation == FlatpakBackend.system_installation) {
+                    return _("%s Systemwide (non-curated)").printf (origin);
```
cassidyjames, line 341: Hrm, I don't know the best way to communicate this inside a combobox. I think at least hyphenated (`system-wide`) or just `system`? And maybe since we automatically select the "right" source and have non-curated warnings anyway, we could get rid of the `(non-curated)` bit here (and below, to match) to clean things up?

```suggestion
                    return _("%s (system-wide)").printf (origin);
```

src/Core/Package.vala
```diff
@@ -341,6 +336,11 @@ public class AppCenterCore.Package : Object {
...
+                }
+
                 return _("%s (non-curated)").printf (origin);
```
cassidyjames, line 344:
```suggestion
                return origin;
```

src/Core/FlatpakBackend.vala
```diff
@@ -412,30 +523,67 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+
+        job.result = Value (typeof (bool));
+        job.result.set_boolean (true);
+        job.results_ready ();
+    }
+
+    private void preprocess_metadata (bool system, GLib.GenericArray<weak Flatpak.Remote> remotes, Cancellable? cancellable) {
+        unowned Flatpak.Installation installation;
+
+        string dest_path;
```
tintou, line 562:
```suggestion
        unowned string dest_path;
```

src/Core/FlatpakBackend.vala
```diff
@@ -528,48 +676,94 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+                    new_package_list[key] = package;
+                }
+            });
         }
+
+        package_list = new_package_list;
+    }
+
+    private string generate_package_list_key (bool system, string origin, string bundle_id) {
+        string installation = system ? "system" : "user";
```
tintou, line 748:
```suggestion
        unowned string installation = system ? "system" : "user";
```

tintou:
Two really tiny things and ready to go :slightly_smiling_face:

tintou (approved):

Issue #839, opened 2018-10-04
Uninstall process also includes install step?


peteruithoven:
When uninstalling an app I see the following steps: 
1. Waiting for authentication
2. Loading cache
3. Running
4. Removing Packages
5. **Installing**
6. Finished

Why is there an "Installing" step when uninstalling?

![peek 2018-10-05 00-34](https://user-images.githubusercontent.com/523210/46507006-b1b4b780-c836-11e8-925e-b66b6fc508a9.gif)

danirabbit:
I think this is fixed since we're no longer used packagekit

PR #1311, opened 2020-05-13
AbtractAppContainer: Pack progress below cancel

danrabbit:
This works, but the buttons are now very wide. Need to look into making the progressbar less wide


danrabbit:
I cannot for the life me figure out how to override the minimum 150px width. Gnome Software just fakes a progressbar in CSS. I'm *this close* to drawing a progressbar on pair of overlaid grids

cassidyjames:
This also partially papers over #839, as it makes that text much less prominent.

PR #1314, opened 2020-05-13
MainWindow: Use a revealer to show/hide the updates badge

danrabbit:
and we get a nice crossfade animation :)

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -39,6 +39,7 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
     private ulong task_finished_connection = 0U;
     private Gee.LinkedList<string> return_button_history;
     private Gtk.Label updates_badge;
+    private Gtk.Revealer updates_badge_revealer;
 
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -261,10 +265,10 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
 
     public void show_update_badge (uint updates_number) {
         if (updates_number == 0U) {
-            set_widget_visibility (updates_badge, false);
+            updates_badge_revealer.reveal_child = false;
         } else {
             updates_badge.label = updates_number.to_string ();
-            set_widget_visibility (updates_badge, true);
+            updates_badge_revealer.reveal_child = true;
         }
```
cassidyjames, line 272: It looks like we're only using `set_widget_visibility ()` in one other place besides this, would we want to inline that as part of this branch?

danrabbit: I didn't want to touch it yet, but yeah eventually I want to get rid of that method

PR #1319, opened 2020-05-13
Sort package origins

### LINKED ISSUE #1029: Prefer installed version of app

### LINKED ISSUE #1082: Prefer Flatpak

davidmhewitt:
Fixes #1082 
Fixes #1029 

I've split this off from #1308 to make the review easier.

It implements the sorting order as detailed [here](https://github.com/elementary/appcenter/issues/1115#issuecomment-556152487) (minus the systemwide Flatpak stuff of course)


cassidyjames:
@davidmhewitt does/should this affect which origin shows when clicking a search result? My test case here:

1. Create `/etc/apt/sources.list.d/appcenter.list`:

    ```
    # AppCenter from Juno/Hera (bionic) until we get things rolling for Odin (focal)
    deb [arch=amd64] http://packages.elementary.io/appcenter bionic main
    ```

2. Ensure Flathub was added as a user remote

3. Restart AppCenter, then search for `dippi` which is available in AppCenter and Flathub

Right now, AppCenter source shows up first. However, when I choose the Flathub version and install Dippi, the AppCenter one is always the result that shows up when clicking search results. They appear to be sorted in the "correct" order in the origin combobox, though.

davidmhewitt:
@cassidyjames So I can't reproduce that here.

I had that issue when installing the Flathub version and then using the back button to go back to the search results. Since the results are cached in the view, clicking the result still goes to the AppCenter one. Is that what you're seeing?

However, for me, after clearing the search and going again. It was correct.

I can look at adding some extra debugging to this branch if you can reproduce the issue?

cassidyjames:
As a note, I also got an "Application state not set as installed in AppCenter for package: Dippi" error when trying to uninstall the AppCenter version after a reboot. Ignoring the error and trying it again worked.

cassidyjames:
@davidmhewitt yeah I can't reproduce the AppCenter/Flathub search results ordering thing again, so I'm not sure what I was seeing.

cassidyjames:
This appears to be working as expected for me. I'll give it some more time dogfooding, but in general I'm happy with it. :)

davidmhewitt:
Going to convert this to a draft for now as I'm about to break this down further and I think there are some issues with this one.

davidmhewitt:
This now ready for review again. A separate branch to optimise the installed state of packages in the backend has now been merged so is no longer part of this branch making it easier to review and test.

This now just implements the sorting logic to ensure that the origins in the dropdown are listed in the desired order and that for the de-duplicated applications in the list view, the one displayed is the most desired one according to that order.

src/Core/BackendAggregator.vala
```diff
@@ -87,27 +87,41 @@ public class AppCenterCore.BackendAggregator : Backend, Object {
...
-        var apps = new Gee.TreeSet<Package> ((a, b) => {
-            return a.normalized_component_id.collate (b.normalized_component_id);
-        });
-
+        var apps = new Gee.HashMap<string, Package> ();
         foreach (var backend in backends) {
-            apps.add_all (backend.search_applications (query, category));
+            var results = backend.search_applications (query, category);
+
+            foreach (var result in results) {
```
tintou, line 113: It might be worth adding a:
```vala
var result_component_id = result.normalized_component_id;
```
And reusing it when necessary :slightly_smiling_face:

PR #1330, opened 2020-05-15
Have backend classes construct their own Settings objects

davidmhewitt:
Instead of having all of these core/backend classes use the `settings` instance belonging to `AppCenter.App` and hence depending on all of the GUI classes, have them construct their own.

This allows these classes to stand more freely and should in theory allow them to be compiled and tested independently from the rest of AppCenter.

### DIFF
--- a/src/Core/Client.vala
+++ b/src/Core/Client.vala
@@ -37,6 +37,12 @@ public class AppCenterCore.Client : Object {
 
     private AsyncMutex update_notification_mutex = new AsyncMutex ();
 
+    public static GLib.Settings settings;
+
+    static construct {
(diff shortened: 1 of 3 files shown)


danrabbit:
Hrm, I'm not certain about having 3 settings instances here. Would it be overkill to have a tiny settings singleton class so that we only have one instance, but it doesn't have to be initialized in the App class?

davidmhewitt:
I was doing this with the intention of adding some unit tests that didn't depend on having an X server. But I'm going down the route of using a virtual X server and doing the tests in that, so this is less of a requirement for now.

I'll re-do this in the way you suggest if it becomes necessary in the future.

PR #1346, opened 2020-05-30
AbstractAppContainer: remove threading and locking

davidmhewitt:
In theory, all of the threading and locking is now done in the backend, so none of this should be necessary anymore.

This affects the install/update action button and the cancellation of that action across all views, but my initial testing on this has gone well.


danrabbit:
Looks like with this branch that everything will work as expected and then it will switch to the progress view:

Screenshot from 2020-06-08 10 21 41@2x
Screenshot from 2020-06-08 10 22 10@2x

davidmhewitt:
Whoops, it seems that didn't happen with Flatpak packages due to a race which is what I was testing with at the time. The class listens for signals on the package and updates the view accordingly, so there's no need to manually set the view but we still had an explicit switch to the progress view that was now happening after the package update. I've removed it, so that should be fixed now.

PR #1348, opened 2020-05-31
Remove broken spinner in update view header

davidmhewitt:
I'd forgotten we had a little spinner header in the installed/updates view, but it's been broken for a while, so this PR removes the associated code. When it worked, it looked something like this:

image


davidmhewitt:
@danrabbit @cassidyjames Any thoughts on the UX of having two spinners like this? If we don't want the one in the list view, I can throw away some complexity instead of fixing it.

cassidyjames:
Yeah I think I'm fine with removing the one in the list view. It hasn't been there for quite some time and nobody really noticed or complained. 😅

davidmhewitt:
@cassidyjames Thanks, I've updated this PR to remove all the associated broken bits of code.

PR #1349, opened 2020-06-01
Installed state optimisations

davidmhewitt:
This takes some of the work from #1319 and breaks it into its own PR.

This is quite an important first step in some refactoring/optimizations that start to unblock some of the work of prioritizing different package sources and enabling systemwide flatpak support.

This should bring about no functionality changes to make testing easier.

In summary, the changes are:
* Only update the flatpak cache if changes weren't triggered by AppCenter. We were previously picking up on the `changed` signal for stuff that was being changed in AppCenter which is unnecessary.
* Refresh the installed/updatable state of applications when external changes happen (i.e. CLI tools)
* Never query the backend synchronously for the installed state of an application. Check once at startup and then try to keep it up to date.
* Only emit state changed signals on packages when the state actually changes
* Only refresh the available updates once at the end of updating a bunch of packages instead of in between each one

This will need some pretty in-depth testing to check that everything stays in sync, but I'd appreciate it if people could run this branch and test it out for a few days as it's blocking quite a bit of my other work.

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -85,8 +85,22 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
             }
 
             installation_changed_monitor.changed.connect (() => {
-                debug ("Flatpak installation changed.");
-                trigger_update_check.begin ();
+                // Only trigger a cache refresh if we're not doing anything (i.e. its an external change)
+                if (!working) {
+                    debug ("Flatpak installation changed.");
+
+                    // Clear the installed state of all packages as something may have changed we weren't
+                    // aware of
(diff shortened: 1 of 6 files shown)


davidmhewitt:
@cassidyjames @danrabbit If you have some time on this fine reviewsday Tuesday, I'd appreciate it if you could install this branch and give it a spin.

It takes the optimizations from #1319 , but none of the functional changes. Cass had previously tested that branch and was reasonably happy, but I found some issues around the behaviour when flatpaks were changed using the CLI which this aims to not break.

I'm on a bit of a spree with some more AppCenter improvements, but this is blocking quite a bit of my other stuff now, so it's pretty high priority to get some eyes on this if possible.

PR #1350, opened 2020-06-01
Rewrite ScreenshotCache

### LINKED ISSUE #1160: CPU usage too high
For some reason, CPU usage can get up to 100% of a thread as seen on this image : 
image

Of course I'm on Hera (update from Juno), Kernel 4.15.0-72-generic, Intel i5-8250U @ 3.4GHz, updates are completed, no task going, and it's constant. 

If anything I did was installing an app with Sideload in the morning from flathub, and I had 6 hours uptime, 4 of which the app center was launched.

Don't hesitate to ask for more information.

davidmhewitt:
Fixes #1160

When the screenshot cache folder goes over 100MB, there's some code in this class that gets stuck in an infinite loop, causing AppCenter to use 100% of a CPU core. And starts using another core every time you open a new app info page. Currently the only way to stop it is to kill appcenter and delete the screenshot cache folder. I've marked this as high priority as a result.

When looking into that issue I found the code here didn't really fit with our code style, so I did the following:

* Added a copyright header
* Removed yoda conditions
* Used simpler GLib methods rather than doing stuff manually with C stdlib and Soup
* Removed while loops that had the potential to go on forever if stuff went wrong
* Returned booleans to indicate success or not instead of integers
* Removed the inline thread and just used async methods instead because we already use a thread to call on this class

By the time I was finished doing that, it was basically rewritten.

It's easier to review the code in split diff view, or just read the new file.


src/Core/ScreenshotCache.vala
```diff
@@ -1,141 +1,152 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+public class AppCenterCore.ScreenshotCache : GLib.Object {
     private const int MAX_CACHE_SIZE = 100000000;
 
     public string screenshot_path;
```
danrabbit, line 22: Since it looks like we assign this in the construct block, should this be:
```suggestion
    public string screenshot_path { get; private set; }
```

src/Core/ScreenshotCache.vala
```diff
@@ -1,141 +1,152 @@
...
     private Soup.Session session;
 
-    public static ScreenshotCache? new_cache () {
-        var screenshot_path = Path.build_filename (
+    private GLib.File screenshot_folder;
+
+    private static ScreenshotCache? instance = null;
+
```
danrabbit, line 28: For a proper singleton we need the following right?
```suggestion
private  ScreenshotCache () {}
```

src/Core/ScreenshotCache.vala
```diff
@@ -1,141 +1,152 @@
...
+        init ();
     }
 
-    // Prune the cache directory if it exceeds the `MAX_CACHE_SIZE`.
-    public void maintain () {
-        var screenshot_usage = summarize_screenshot_usage ();
-        if (screenshot_usage == -1) {
-            return;
-        }
+    private void init () {
```
danrabbit, line 44: Does this need to be a separate method from `construct {}`?

cassidyjames:
@davidmhewitt to confirm, does this address https://github.com/pop-os/shop/issues/193, or is it a first step in solving it?

src/Core/ScreenshotCache.vala
tintou: no that's not required

src/Core/ScreenshotCache.vala
```diff
@@ -1,141 +1,152 @@
...
-        string? name = null;
-        Dir dir;
-        if (read_dir (out dir, screenshot_path) == -1) {
+    private async void delete_oldest_files (uint64 screenshot_usage) {
+        var file_list = new Gee.ArrayList<GLib.FileInfo> ();
+
+        FileEnumerator enumerator;
+        try {
+            enumerator = yield screenshot_folder.enumerate_children_async (
+                string.join (",", "standard::*", GLib.FileAttribute.TIME_CHANGED),
```
tintou, line 94:
```suggestion
                GLib.FileAttribute.STANDARD_NAME + "," + GLib.FileAttribute.STANDARD_TYPE + "," + GLib.FileAttribute.STANDARD_SIZE + "," + GLib.FileAttribute.TIME_CHANGED,
```

src/Core/ScreenshotCache.vala
```diff
@@ -1,141 +1,152 @@
...
+        // Sort the files by ctime (when file metadata was changed, not content)
+        file_list.sort ((a, b) => {
+            uint64 a_time = a.get_attribute_uint64 (GLib.FileAttribute.TIME_CHANGED);
+            uint64 b_time = b.get_attribute_uint64 (GLib.FileAttribute.TIME_CHANGED);
 
-        while ((name = dir.read_name ()) != null) {
-            string entry = Path.build_filename (screenshot_path, name);
-            int file_size = query_size (entry);
-            if (file_size == -1) {
+            if (a_time < b_time) {
```
tintou, line 120: This would do it:
return (int) (b_time - a_time);

src/Core/ScreenshotCache.vala
```diff
@@ -148,115 +159,62 @@ public class AppCenterCore.ScreenshotCache {
...
-                    warning ("GET request of %s failed: %s\n", url, Soup.Status.get_phrase (msg.status_code));
-                    result = 1;
-                    Idle.add ((owned)callback);
-                    return true;
-                }
+        if (msg.status_code != Soup.Status.OK || modified == null) {
+            warning ("HEAD request of %s failed: %s", url, msg.reason_phrase);
 
-                var data = msg.response_body.data;
+            if (file.query_exists ()) {
```
tintou, line 191: This is equivalent :)
```suggestion
            return file.query_exists ();
```

src/Core/ScreenshotCache.vala
davidmhewitt: I'm not so sure...

`(uint64)0 - (uint64)1 == UINT64_MAX`

Then casting UINT64_MAX to integer (int32) would lose precision and you'd get 0, which would not be the correct result.

src/Core/Client.vala
```diff
@@ -38,7 +38,7 @@ public class AppCenterCore.Client : Object {
     private AsyncMutex update_notification_mutex = new AsyncMutex ();
 
     private Client () {
-        Object (screenshot_cache: AppCenterCore.ScreenshotCache.new_cache ());
+        Object (screenshot_cache: new AppCenterCore.ScreenshotCache ());
```
tintou, line 41: The best would be to do something like this:
```
public AppCenterCore.ScreenshotCache? screenshot_cache { get; construct; default: new AppCenterCore.ScreenshotCache (); }
```

src/Core/ScreenshotCache.vala
davidmhewitt: Obviously that's an unrealistic case, but if `b < a` then `b - a` will probably be greater than INT32_MAX which would be truncated to 0 on the cast to `int`, when we'd expect the result to be < 0

davidmhewitt:
@cassidyjames It fixes that problem, yes. But there's some more code in `AppInfoView` related to screenshots that looks like it has the potential to do bad things :tm: 

I found the trigger/reproducer for your linked issue. If the screenshot cache grows >=100MB , then every time you open an app page, this `ScreenshotCache` class spawns a new thread to try and clean up and has a while loop that gets stuck forever (using 100% of a CPU thread):
```
while (screenshot_usage > MAX_CACHE_SIZE)
```

I didn't look into why it gets stuck, but this rewritten code doesn't have the same issue. You can make a 100MB file in the screenshot cache folder with the following command if you want to test before/after:

```
dd if=/dev/zero of=~/.cache/io.elementary.appcenter/screenshots/bigfile.png bs=1M count=100
```

src/Core/ScreenshotCache.vala
tintou: Yeah you're right, leaving it like this is fine then :)

src/Core/ScreenshotCache.vala
```diff
@@ -168,18 +169,40 @@ public class AppCenterCore.ScreenshotCache : GLib.Object {
...
 
-        if (msg.status_code != Soup.Status.OK || modified == null) {
-            warning ("HEAD request of %s failed: %s", url, msg.reason_phrase);
+        string? iso8601_string = null;
+        GLib.DateTime? modified_time = null;
+        // Parse the HTTP date format using Soup
+        var http_date = new Soup.Date.from_string (modified);
+        if (http_date != null) {
+            // Convert it to ISO8601
+            iso8601_string = http_date.to_string (Soup.DateFormat.ISO8601);
```
tintou, line 185: You can keep it in this scope :)
```suggestion
            string? iso8601_string = http_date.to_string (Soup.DateFormat.ISO8601);
```

PR #1366, opened 2020-06-11
AppListUpdateView: Remove "update all" complexity

davidmhewitt:
Similar to https://github.com/elementary/appcenter/pull/1346 in that it removes a lot of complexity in UI classes that's already handled in the backend. This replaces the logic for what happens when you click the "Update All" button with something much simpler.

We don't need to manually update the state of each row because they handle this internally by listening to state changes on the package.

To handle the case where if the currently running update is cancelled then the rest should be cancelled, I allowed errors from `Package.update` to bubble higher up the chain. We catch the cancelled error and break out of the loop.

The fact that other types of errors weren't handled anywhere other than a warning to the console raised the point that we don't currently have an error dialog for failed package updates. We have one for install/uninstall/refresh cache etc, but not update, so I've taken the opportunity to add one here.


danrabbit:
One problem here is that after clicking "Update All" only the currently updating component is changed to the progress view. This means a couple things:

* Components that are in the update queue still have a sensitive "Update" button
* It's not possible to cancel pending updates until the update for that component has begun

davidmhewitt:
@danrabbit So the first issue there is easily solvable (and has been).

I guess we need to question what the cancellation behaviour ought to be. If I click update all and then cancel an update, what should happen?

If I cancel the currently running update, should all of the remaining updates be cancelled too? This is what currently happens with this branch. In master, if you cancel the first update when it's running, then the rest are cancelled. Otherwise, if the first update completed, you have to cancel each of the remaining updates in turn if you want to cancel all.

If I cancel a pending update (i.e. not the currently running one, but one further down the list), do all updates get cancelled, or does just that one get removed from the pending list?

cassidyjames:
@davidmhewitt could you fix this conflict with master?

danrabbit (approved):

PR #1381, opened 2020-07-11
clicking updates badge go to installed page

### LINKED ISSUE #1344: Clicking updates badge itself should navigate to updates tab
The red badge itself in the AppCenter window (not the one in the dock) doesn't respond to clicks. We should probably pack this in an eventbox or make it a button or something and make sure it does respond to clicks

rafaelcorsi:
closes #1344

Peek 2020-07-11 10-49

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -154,11 +154,15 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         updates_badge = new Gtk.Label ("!");
         updates_badge.get_style_context ().add_class ("badge");
 
+        var eventbox_badge = new Gtk.EventBox ();
+        eventbox_badge.add (updates_badge);
+        eventbox_badge.button_press_event.connect (badge_event);
+
         updates_badge_revealer = new Gtk.Revealer ();
         updates_badge_revealer.halign = Gtk.Align.END;
         updates_badge_revealer.valign = Gtk.Align.START;
         updates_badge_revealer.transition_type = Gtk.RevealerTransitionType.CROSSFADE;
-        updates_badge_revealer.add (updates_badge);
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -272,6 +276,11 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         }
     }
 
+    bool badge_event (Gtk.Widget sender, Gdk.EventButton evt) {
```
danrabbit, line 279: We should make sure to explicitly mark this as private to follow code style

```suggestion
    private bool badge_event (Gtk.Widget sender, Gdk.EventButton evt) {
```

src/MainWindow.vala
```diff
@@ -272,6 +276,11 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
...
+        go_to_installed ();
+        return (true);
```
danrabbit, line 281: We can be more explicit here as well and use this constant:
```suggestion
        return Gdk.EVENT_STOP;
```

src/MainWindow.vala
```diff
@@ -154,11 +154,15 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         updates_badge = new Gtk.Label ("!");
         updates_badge.get_style_context ().add_class ("badge");
 
+        var eventbox_badge = new Gtk.EventBox ();
+        eventbox_badge.add (updates_badge);
+        eventbox_badge.button_press_event.connect (badge_event);
```
danrabbit, line 159: It's better to use the release event here so that we don't respond to things like drag events if the user wants to move the window for example

```suggestion
        eventbox_badge.button_release_event.connect (badge_event);
```

PR #1383, opened 2020-07-12
Include RDNN'd icon

danrabbit:
I think we should include the AppCenter icon here and reference it specifically to make sure that it's only be used in the context of AppCenter.

Currently a user could install another package manager and it would use the same icon as AppCenter. This is probably not good from a trust/security standpoint


cassidyjames:
I have raised this before, I thought, and you said it was also important for other things that are referencing a generic package installer to show the AppCenter logo for people to know that's where it will take them. But I think it does make more sense to RDNN it and then probably include a generic version of the icon in the icon set.

danrabbit:
I can't think of anything else that is referencing AppCenter off the top of my head

cassidyjames:
Off the top of my head:

Sideload shows the generic package manager icon alongside the name of the app that handles AppStream, I believe. We could update that to grab the app's icon, too.

Onboarding shows the AppCenter icon on the AppCenter view.

danrabbit:
@cassidyjames I don't believe sideload uses the package manager icon. It has its own unique icon

Onboarding checks that AppCenter is installed, so it probably makes sense to update that to use an AppCenter-specific icon

cassidyjames:
Sideload uses it here: 

https://github.com/elementary/sideload/blob/05789bae8d05f655fdfdb13ed3fc10e9e0cf3d90/src/Views/MainView.vala#L127

But we should be able to get the correct app’s icon from `GLib.Application`?

danrabbit:
Ah yeah, we should probably be getting the correct app icon there anyways since we do pull in the correct app name

cassidyjames (changes requested):
I _think_ it would it make more sense to use `Build.PROJECT_NAME` or `application_id` in all these contexts, since the icons are named by meson, too—that way there's less hardcoding of strings, and it's guaranteed to stay in sync.

cassidyjames (approved):

PR #1384, opened 2020-07-12
Set contextual icons in notifications

danrabbit:
Requires changes in the notification server. Just for testing for now


cassidyjames (approved):
Looks good! The notifications changes have landed, plus this fixes an issue with badging with the same icon.

PR #1408, opened 2020-09-15
Fix badge color in dark style

danrabbit:
Make sure we use shades of strawberry and not error color

Also breaks out the badge style into its own CSS file so we don't leak styles in the application

and does the same to categories while we're here


cassidyjames (approved):
Looks much better, and scoping the styles seems cleaner :+1:

PR #1409, opened 2020-09-17
BackendAggregator: Use Gtk.App inhibit

danrabbit:


jeremypw:
Neither this nor master prevents suspending the machine with the wingpanel session indicator (at least not in a VirtualBox vm) - is it intended that it should or does Application.inhibit () only inhibit automatic suspends due to inactivity?

danrabbit:
@jeremypw yes this only inhibits automatic suspend. The session indicator is not currently aware of apps that have requested inhibit

src/Core/BackendAggregator.vala
```diff
@@ -37,13 +39,25 @@ public class AppCenterCore.BackendAggregator : Backend, Object {
                         remove_inhibit_timeout = 0;
                     }
 
-                    SuspendControl.get_default ().inhibit ();
+                    if (inhibit_token != 0) {
+                        app.uninhibit (inhibit_token);
+                    }
```
jeremypw, line 44: This seems to defeat the object of the uninhibit timeout.  By inserting logging messages and running appcenter in a terminal you can see that about a dozen inhibit/uninhibit pairs are produced in a short time while starting up. Why not just reset the timeout and keep the token without uninhibiting?

jeremypw:
> @jeremypw yes this only inhibits automatic suspend. The session indicator is not currently aware of apps that have requested inhibit

Can bad things happen if the user suspends? If so this is a bug in the wingpanel session indicator.

Not sure how to trigger an automatic shutdown during package operations to test this PR.

danrabbit:
@jeremypw Yeah theoretically if the computer goes to sleep in the middle of a package operation that's probably dangerous. I thought there was already an issue opened in the indicator to show apps who have requested inhibit, but I can't find anything for it

jeremypw:
There is an (old) open issue about the session indicator should allow apps to close "properly" rather than forcibly terminating them when logging out, but that is not quite the same thing.

src/Core/BackendAggregator.vala
danrabbit: Yeah I think you're right. Changed it to only add an inhibit if one does not already exist

jeremypw (dismissed):
There is a question over whether `notify_property ("working")` should be emitted after timeout ends when `working` is  __`false`__ so that the spinner stops at the same time that the inhibit is released.  At the moment it gets started and stopped multiple times.  However, that issue exists in master so need not be fixed here.

jeremypw:
Seems my package system has got messed up while testing this PR with manual suspension during package update.  I'll start working on a new VM.

jeremypw (approved):
Works as expected no regressions found

PR #1411, opened 2020-09-22
Styles: Add fallbacks for downstreams

cassidyjames:
Fixes https://github.com/pop-os/shop/issues/215. Does not affect or override styles on elementary OS.

### DIFF
--- a/data/io.elementary.appcenter.gresource.xml
+++ b/data/io.elementary.appcenter.gresource.xml
@@ -4,6 +4,7 @@
     <file alias="application.css" compressed="true">styles/application.css</file>
     <file alias="arrow.css" compressed="true">styles/arrow.css</file>
     <file alias="categories.css" compressed="true">styles/categories.css</file>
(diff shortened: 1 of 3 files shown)


danrabbit:
Can we add these to the widgets they effect instead of to the whole screen?

cassidyjames:
@danrabbit sure, though I think we use palette colors for the update badge, default application header background, and categories. Does adding it for the whole screen hurt anything in this case, or would it be better to add it three times?

danrabbit:
My understanding is that this pulls in the provider for the construction of all style contexts, even ones that don't use it. So I think it's less efficient, especially as we have more widgets on screen. I could be wrong about this though and maybe it's negligible

danrabbit:
Poked someone about it and they said probably not a significant performance impact

danrabbit (approved):
I can confirm that this resolves issues when using Adwaita and I don't see any impact in Greenfield

Issue #1403, opened 2020-09-04
"runtime Flatpak runtime" updates showing under OS updates


cassidyjames:
As of today running master on my elementary OS 6 install, I am getting two updates listed under Operating System Updates, both called "runtime Flatpak runtime". Hitting "Update" doesn't appear to do anything.

Running `flatpak update` outputs:

```
 1.     org.gnome.Platform.Locale    3.36      u     flathub    < 323.0 MB (partial)
 2.     org.gnome.Platform           3.36      u     flathub    < 324.5 MB
```

Performing the Flatpak update from Terminal indeed gets rid of the OS updates listed in AppCenter.

cassidyjames:
As a note: this prevents Flatpak app updates on affected systems. We also are not getting any sort of error dialog, the install just fails silently and the updates remain showing up.

janxkoci:
I see this also in Hera, before I could see which runtimes are getting updated, now they are all listed with the same generic "name".

Also, after the same recent update, flatpak updates started to show up in AppCenter even on a netbook where I never used Sideload and installed two flatpak apps just from the terminal. Not getting automatic updates worked great for me as the netbook has just 32GB eMMC drive, but now flatpaks show up mixed with other updates. It's quite annoying, since I explicitly didn't allow AppCenter to handle flatpaks on this machine.

davidmhewitt:
> I see this also in Hera

Yes, we already know about this issue (which is why this issue report exists) and there is already a fix pending for it.

> Not getting automatic updates worked great for me 

AppCenter doesn't do automatic updates? You have to open it and click the update button.

> since I explicitly didn't allow AppCenter to handle flatpaks on this machine.

The fact that AppCenter couldn't update/manage packages that were installed systemwide before was a limitation of AppCenter, not that you had "disallowed" AppCenter to update them by installing them with the terminal. Flatpak packages are larger than traditional packages, yes, but you already have them installed. Updating them shouldn't cause much extra disk space utilisation, so I hope this isn't an issue.

However, if you want to handle your updates individually, then the terminal is probably better suited for this job. Please feel free to open any additional issue reports that you feel could improve this situation, but this is irrelevant discussion for this bug that will get lost when the bug is closed.

janxkoci:
Sorry, I only seen elementary OS 6 mentioned in the description, so I thought I will give heads up about Hera too, in case it was not known.

[...]

PR #1414, opened 2020-09-24
FlatpakBackend: Fix stuck updates issue

davidmhewitt:
Fixes part of #1403 (The names are still not displayed correctly, but they are updated now)

It turns out if you invert your booleans and don't actually tell flatpak to update any packages, things go a bit awry when you try to update flatpak packages.

I have no idea who wrote this code originally. Hint: *it definitely wasn't me* :grimacing:

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -1094,7 +1094,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
         }
 
         if (user_updates.length > 0) {
-            run_user = false;
+            run_user = true;
             transactions++;
         }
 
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Huzzah!

src/Core/FlatpakBackend.vala
```diff
@@ -1130,6 +1130,14 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
             return false;
         }
 
+        try {
+            foreach (var bundle_id in ids) {
```
tintou, line 1134: `foreach (unowned string bundle_id in ids) {` is required to avoid copying the string ;)

PR #1418, opened 2020-09-27
Replace deprecated code

meisenzahl:
- [x] ../src/Core/PackageKitBackend.vala:420.90-420.120: warning: `AppStream.ProvidedKind.MIMETYPE' is deprecated. Use MEDIATYPE
- [x] ../src/Dialogs/StripeDialog.vala:192.43-192.57: warning: `Gtk.Dialog.get_action_area' has been deprecated since 3.12
- [x] ../src/Views/AppListUpdateView.vala:323.17-323.38: warning: implicit .begin is deprecated
- [x] ../src/Dialogs/RestartDialog.vala:81.31-81.45: warning: `Gtk.Dialog.get_action_area' has been deprecated since 3.12


meisenzahl:
1. The `action_area` in `RestartDialog` gets only styled. Can we remove this code because `Gtk.Dialog.get_action_area' has been deprecated?
2. The `action_area` in `StripeDialog` gets styled and a `Gtk.LinkButton` is added. Can we replace the logic with [Gtk.Dialog.add_button](https://valadoc.org/gtk+-3.0/Gtk.Dialog.add_button.html)?

danrabbit:
I think we should be able to do something like:

```
        var privacy_policy_link = new Gtk.LinkButton.with_label ("https://stripe.com/privacy", _("Privacy Policy")) {
            halign = Gtk.Align.START,
            hexpand = true
        };

        add_action_widget (privacy_policy_link, Gtk.ResponseType.NONE);
```

But to avoid having to manually mess with margins in dialogs, we need https://github.com/elementary/stylesheet/pull/839

danrabbit (approved):

PR #1419, opened 2020-09-30
Support codec requests from gstreamer1.0-packagekit

### LINKED ISSUE #623: Codec installation
Edit: Updated with more details.

Downsteam issue https://github.com/pop-os/shop/issues/45 suggests "implementing gstreamer1.0-packagekit".

## Feature
Provide a interface to install gstreamer codecs. Currently installing codecs via `gstreamer-install-plugin-helper` results in:

```
/usr/lib/gst-install-plugins-helper "gstreamer|1.0|io.elementary.videos|Flash demuxer|decoder-video/x-flv" returns:
** Message: 18:18:51.171: PackageKit: xid = 0
** Message: 18:18:51.171: PackageKit: desktop_id = (null)
** Message: 18:18:51.171: PackageKit: Codec nice name: Flash demuxer
** Message: 18:18:51.171: PackageKit: structure: gstreamer1(decoder-video/x-flv)()(64bit)
** Message: 18:18:51.174: PackageKit: Did not install codec: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.PackageKit was not provided by any .service files
```

**Describe the solution you'd like**
Appcenter provides the interface including throwing any dialogs related to gstreamer plugins

**Existing work**
Gnome Software already implements it https://github.com/GNOME/gnome-software/blob/fea162c434fd14dd8f7d1be308ac7e1c6b1b70be/src/org.freedesktop.PackageKit.xml#L326

**Additional context**
Blocks https://github.com/elementary/videos/pull/155 and results in non function code in Music https://github.com/elementary/music/blob/master/src/Dialogs/InstallGstreamerPluginsDialog.vala

davidmhewitt:
Fixes #623 

This listens on the correct bus and uses PackageKit to find the correct gstreamer packages to install when missing plugins/codecs are requested for install by a gstreamer-based app. There's a PR open on Videos that implements that functionality and I've been testing with that.

Still a WIP, but opening for visibility and posterity. Still TODO:
- [ ] Throw a dialog that explains what's going to be installed and the potential license implications
- [ ] Actually install them

Note: gstreamer does not have fully fledged "codec" AppData hence why I'm not using a more generic Appstream way of looking these up, meaning this may not work on other distros. I found some old Github threads from 2017 ish that claim the appstream codec spec isn't fully thought out yet, and gstreamer aren't going to ship the needed metadata upstream until it is. It looks like things haven't moved on since then.


Conan-Kudo:
cc: @decathorpe

decathorpe:
@Conan-Kudo Uhhhh I have no idea how PackageKit on fedora deals with codec lookup, but it seems to work with gstreamer, since totem et al. ask for codec installation in gnome-software if they're not available to gstreamer?

davidmhewitt:
Yeah, this is more around how gnome-software looks up which plugins to install based on what GStreamer asks for. I _think_ gnome-software does it in the same way as what I'm proposing here, so Fedora should be fine, but it relies on PackageKit having the relevant code in its backend plugin (aptcc/dnf) to be able to handle these. If GStreamer, or downstreams shipped appstream metadata that said distro package x provides codec y, then we could do this in a much more package manager independent way.

But we definitely don't have that appstream data on Ubuntu/elementary, so this is the only way that works. It might just be some other distros that have less fully formed PackageKit backends for their package managers that aren't able to use this feature.

Conan-Kudo:
This feature should generally work in Fedora, since you're manually constructing the virtual names embedded in gstreamer codec packages in Fedora: https://koji.fedoraproject.org/koji/rpminfo?rpmID=25815123

PR #1421, opened 2020-10-02
Optimise search rank algorithm

davidmhewitt:
Instead of doing `string.has_prefix` and `string.contains`, we can do `string.index_of` and use the result to figure out the answer to both. 

If the needle is at the start, it's as fast as `has_prefix`, if it's not in the string, it's as slow as `contains` but this means we're doing slightly less expensive string operations in some cases. It probably doesn't save much, but for some reason I had this idea rattling around in my head as I was falling asleep, so I had to open a quick PR and clear my conscience... :sleeping:


tintou (approved):
Looks like a nice improvement

Issue #1142, opened 2019-11-20
Use Handy.Paginator for screenshots


danrabbit:
So we get nice swipe gestures

cassidyjames:
I think the only thing that's not just drop-in for Hdy.Paginator is our back/forward buttons, but that should still be able to translate well afaict.

PR #1426, opened 2020-10-07
Use libhandy for screenshots

fs3o5:
Adds #1142 and switches dependency libhandy to libhandy-1


cassidyjames (changes requested):
Yay, swiping on the trackpad feels so much better, here! But it looks like some things were lost in the port to LibHandy.

- [x] Indicator dots are not visible (could require in-app styles or stylesheet work)
- [ ] Buttons should show as insensitive when you can't go back/forward

That's what I noticed so far at a glance.

fs3o5:
The buttons should now become insensitive.
I'm not sure what's happening on the dots. For me they become visible as soon as I open any other app.

danrabbit:
I think whatever is causing the same issue with header text color is also affecting the dots here. So I would say it's not an issue with this branch

cassidyjames (approved):
UX-wise this is good to go! @danrabbit were you happy with the code?

PR #1436, opened 2020-10-19
AbstractView: Use Deck instead of Stack

danrabbit:
Does the least to replace Stack with Deck

This branch does not implement multitouch gestures
It also does not solve a bug where category views seem to get destroyed early sometimes


jeremypw (approved):
Looks sane and no new issues noticed

PR #1441, opened 2020-11-04
README.md: List libhandy dependency as libhandy-1-dev

ids1024:
It looks like a5a7f2b updated `meson.build` in this way.


danrabbit (approved):
Thanks for catching this

PR #1443, opened 2020-11-04
SharePopover: Get appinfo for mailto

### LINKED ISSUE #1442: Get default email client icon and name
https://github.com/elementary/appcenter/blob/990470b3cf96448b187f00d8696ddacf967a7b6a/src/Widgets/SharePopover.vala#L34

This is not an FD.o icon name. We should get the default mail client and use its info here

danrabbit:
fixes #1442 

If we can't get appinfo for mailto then this button won't work, so we don't pack it

### DIFF
--- a/src/Widgets/SharePopover.vala
+++ b/src/Widgets/SharePopover.vala
@@ -36,7 +36,7 @@ public class SharePopover : Gtk.Popover {
         var mail_appinfo = AppInfo.get_default_for_uri_scheme ("mailto");
         if (mail_appinfo != null) {
             email_button.image = new Gtk.Image.from_gicon (mail_appinfo.get_icon (), Gtk.IconSize.DND);
-            email_button.tooltip_text = _(mail_appinfo.get_display_name ());
+            email_button.tooltip_text = mail_appinfo.get_display_name ();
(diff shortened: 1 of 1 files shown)


src/Widgets/SharePopover.vala
```diff
@@ -31,9 +31,14 @@ public class SharePopover : Gtk.Popover {
...
     construct {
-        var email_button = new Gtk.Button.from_icon_name ("internet-mail", Gtk.IconSize.DND);
-        email_button.tooltip_text = _("Email");
-        email_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+        var email_button = new Gtk.Button ();
+
+        var mail_appinfo = AppInfo.get_default_for_uri_scheme ("mailto");
+        if (mail_appinfo != null) {
+            email_button.image = new Gtk.Image.from_gicon (mail_appinfo.get_icon (), Gtk.IconSize.DND);
+            email_button.tooltip_text = _(mail_appinfo.get_display_name ());
```
tintou, line 39:
```suggestion
            email_button.tooltip_text = mail_appinfo.get_display_name ();
```

tintou (approved):

PR #1447, opened 2020-11-11
Isolate banner CSS

danrabbit:
This is going to need at least one follow up branch because of the way widgets are set up in AppInfoView, but it makes some progress towards setting banner styles on only the relevant contexts instead of the whole screen

* Move banner CSS to its own file
* The banner widget is only used on the homepage, so it can contain all of its own styles instead of setting some classes from the homepage
* Don't add "banner" and remove "button" on every transition
* Add provider to widget where possible


tintou (approved):
looks good code-wise

Issue #1455, opened 2020-12-09
Throw toast when app is installed


cassidyjames:
## Problem

When you install an app from a list view, it installs… and then nothing happens and there's no way to open it from that view. You either have to click through to the app listing an hit "Open" there, or go to your Applications menu. When AppCenter is not focused and an app is installed, a notification is thrown, at least—but nothing happens when it is focused.

## Proposal

We throw an in-app toast notification when an app is installed, with a handy action to open the app. This could either just be in list views, or globally in the app.

## Prior Art

We throw a system notification when an app is done installing, but only when the app is unfocused. I'm not sure what GNOME, macOS App Store, and Windows Store do off the top of my head.

hanaral:
I don't think that a toast is suitable here, I thought toasts were for more formal small messages. I would not associate an app I download myself with a system function. I'd expect a notification with a colour icon and name, whether the app is focused or not, maybe even have a persistent notification with a progress bar.

cassidyjames:
@hanaral not sure what you mean by "formal small messages." Toasts are for in-app notifications, full stop. Additionally, we tend to not throw system notifications when an app is focused, as those notifications tend to be more useful when you've navigated away from the app and want to return to it. If you're browsing AppCenter and installing apps, then AppCenter throwing an in-app toast to let you know your in-app action has completed makes a lot of sense to me.

(This almost makes me wonder if there should be a way to abstract the in-app toast vs. system notification based on whether the app is focused or not, and implement that in Granite or something… but that's beside the point for now.)

hanaral:
@cassidyjames I personally associate toasts with basic yes/no sorts of functions (like as you mentioned with Camera’s video saved toast)
While I get _why_ it could be suitable for the AppCenter, I still think that something with more personality and colour would be expected of installing an app. 
If you aren’t budging on this, would it be possible to have the emblem-ok-symbolic icon to the right of the text in the toast? This would probably achieve some level of visually communicating the action without the need for a colour icon.

PR #1456, opened 2020-12-09
Throw toast when app is installed

meisenzahl:
Closes https://github.com/elementary/appcenter/issues/1455

Bildschirmfoto von 2020-12-09 19-38-34


src/Application.vala
```diff
@@ -286,6 +286,11 @@ public class AppCenter.App : Gtk.Application {
                     if (main_window != null) {
                         var win = main_window.get_window ();
                         if (win != null && (win.get_state () & Gdk.WindowState.FOCUSED) != 0) {
+                            main_window.toast.title = _("%s has been successfully installed").printf (package.get_name ());
```
cassidyjames, line 289: I know this mirrors other copy from the notification, but "successfully" is redundant. It either installed, or it didn't :smile: 

```suggestion
                            main_window.toast.title = _("%s has been installed").printf (package.get_name ());
```

src/Application.vala
```diff
@@ -286,6 +286,20 @@ public class AppCenter.App : Gtk.Application {
                     if (main_window != null) {
                         var win = main_window.get_window ();
                         if (win != null && (win.get_state () & Gdk.WindowState.FOCUSED) != 0) {
+                            var toast = main_window.toast;
+                            toast.title = _("%s has been installed").printf (package.get_name ());
```
cassidyjames, line 290: See #1458 

```suggestion
                            toast.title = _("“%s” has been installed").printf (package.get_name ());
```

cassidyjames:
Screenshot from list view
Screenshot from 2020-12-09 12-46-56

cassidyjames (approved):
UX approval from me. My only open question is if it would make sense to suppress the toast on the app's own app info view (since it's redundant with the existing "Open" button that gets revealed), but that's not a blocker for me.

meisenzahl:
I will look into hiding the toast on the app's own app info view.

src/Application.vala
```diff
@@ -282,10 +283,17 @@ public class AppCenter.App : Gtk.Application {
...
+                    last_installed_package = package;
+
                     // Check if window is focused
                     if (main_window != null) {
                         var win = main_window.get_window ();
                         if (win != null && (win.get_state () & Gdk.WindowState.FOCUSED) != 0) {
+                            var toast = main_window.toast;
+                            toast.title = _("“%s” has been installed").printf (package.get_name ());
+
+                            toast.send_notification ();
```
donadigo, line 295: It would be better to handle sending the toast and saving the package reference inside MainWindow, instead of trying to expose `last_installed_package` in Application and then accessing it back in MainWindow. 

Simply declare a `private AppCenterCore.Package? last_installed_package` member and a `public void send_installed_toast (AppCenterCore.Package package)` method in MainWindow and manage sending the toast there instead.

donadigo (changes requested):
I have one issue.

src/Application.vala
meisenzahl: I now also need `selected_package` in `AppCenter.App` to check if a `Toast` should be sent.

donadigo: I mean sure, but that didn't resolve my issue. The toast is still half managed in Application and MainWindow at the same time. And there's also no need to store `selected_package` globaly inside Application. Notice how MainWindow actually calls `show_package` and `return_clicked`  where this property could be set within MainWindow as well.

cassidyjames (commented):
Again, UX-wise this meets my approval. :smile: But I'd like to see a re-review from @elementary/desktop-developers or @donadigo.

src/Application.vala
meisenzahl: Thanks for your feedback. I have refactored that except for `selected_package`.

The `show_package` call in MainWindow is not the same as in View.

meisenzahl: I refactored it using a signal.

@donadigo thanks for being persistent 😃 The code looks much cleaner now 👍

src/MainWindow.vala
```diff
@@ -40,6 +40,11 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
     private Gee.LinkedList<string> return_button_history;
     private Gtk.Label updates_badge;
     private Gtk.Revealer updates_badge_revealer;
+    private Gtk.Overlay overlay;
+    public Granite.Widgets.Toast toast;
```
danrabbit, line 44: Does this need to be public anymore? It doesn't look to me like anything outside this class is accessing it.
```suggestion
    private Granite.Widgets.Toast toast;
```

src/MainWindow.vala
```diff
@@ -40,6 +40,11 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
     private Gee.LinkedList<string> return_button_history;
     private Gtk.Label updates_badge;
     private Gtk.Revealer updates_badge_revealer;
+    private Gtk.Overlay overlay;
```
danrabbit, line 43: It doesn't look like overlay is accessed outside of the construct block, so it can probably be scoped down to `var overlay` instead of declared here

src/MainWindow.vala
```diff
@@ -126,6 +134,19 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
 
+        toast = new Granite.Widgets.Toast ("");
+        toast.set_default_action (_("Open"));
+
+        toast.default_action.connect (() => {
+            if (last_installed_package != null) {
+                try {
+                    last_installed_package.launch ();
+                } catch (Error e) {
+                    warning ("Failed to launch %s: %s".printf (last_installed_package.get_name (), e.message));
```
danrabbit, line 145: Hm, I wonder if this should throw a dialog. If a user clicked the open button and then nothing happened, that seems bad. Thoughts @cassidyjames?

src/MainWindow.vala
```diff
@@ -306,6 +331,16 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         search_entry.text = term;
     }
 
+    public void send_installed_toast (AppCenterCore.Package package) {
+        last_installed_package = package;
+
+        if (selected_package == null || (selected_package != null && selected_package.get_name () != package.get_name ())) {
```
danrabbit, line 337: This probably deserves a comment about what this solves for. It's not immediately obvious to me what this conditional does

danrabbit (dismissed):
Just a few inline comments. Otherwise, looks pretty clean to me :)

src/MainWindow.vala
meisenzahl: For reference: This code is mostly a copy from `AbstractAppContainer.launch_package_app()`

src/MainWindow.vala
meisenzahl: Fair point

meisenzahl: Quite verbose... Open for feedback:

```
// Show a toast if
// no package is selected --> we are not on an app page
// or
// a package is selected --> we are on an app page
// but the app's name is not equal to the package's name that should be installed
if (selected_package == null || (selected_package != null && selected_package.get_name () != package.get_name ())) {
    toast.title = _("“%s” has been installed").printf (package.get_name ());

    toast.send_notification ();
}
```

src/MainWindow.vala
cassidyjames: @danrabbit I'm indifferent. It probably makes sense to do what we do in other instances in AppCenter, and then we can file an issue to throw a dialog on failure.

src/MainWindow.vala
```diff
@@ -306,6 +330,21 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
     }
 
+    public void send_installed_toast (AppCenterCore.Package package) {
+        last_installed_package = package;
+
+        // Show a toast if
+        // no package is selected --> we are not on an app page
+        // or
+        // a package is selected --> we are on an app page
+        // but the app's name is not equal to the package's name that should be installed
```
cassidyjames, line 340: I think this conveys the same information more concisely, without just repeating the code as well. :smile: 

```suggestion
        // Only show a toast when we're not on the installed app's page, i.e if
        // no package is selected (we are not on an app page), or a package is 
        // selected but it's not the app we're installing.
```

src/MainWindow.vala
cassidyjames: Suggestion added above

donadigo (approved):
Code looks good now.

cassidyjames (approved):

PR #1457, opened 2020-12-09
Application: "successfully" is redundant

cassidyjames:
Related to #1456, see https://github.com/elementary/sideload/pull/101 and https://github.com/elementary/sideload/pull/102


danrabbit (approved):
But what if the task failed successfully?!

PR #1460, opened 2020-12-14
Perform flatpak uninstalls with a transaction

### LINKED ISSUE #1406: Uninstalling a flatpak app don't uninstall they locale
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Describe the bug
when uninstalling a flatpak app via AppCenter, it's only remove the app, but the locale extension continues installed.

### To Reproduce
1. On AppCenter, Uninstall a Flatpak App
2. On Terminal, run 'flatpak list --runtime'
3. See that the app locale continues installed

### Expected behavior

that the application and locale are removed.

## Platform Information

Screenshot from 2020-09-14 22-27-38

- [ ] I'm using the latest version from git that I've manually compiled
- [X] I'm using the latest released stable version

davidmhewitt:
Fixes #1406 

This replaces `FlatpakInstallation.uninstall` with a transaction instead. `FlatpakInstallation.uninstall` is now deprecated and only uninstalls a single ref. The transaction also cleans up any locale packs and plugins related to the ref in question on uninstallation.

This only affects uninstalls of flatpak packages, so testing can center around removing both user-wide and system-wide flatpak packages.

I had to mark the `commit` parameter of `Transaction.operation_done` as nullable to get this to work as it provides a null commit on uninstallation, so I've taken the liberty to update the vapi against the version we have in the repos now and include the metadata file used to generate it while I was here. I'll push annotation fixes upstream.


vapi/Flatpak-1.0.metadata
```diff
@@ -0,0 +1,2 @@
+Transaction.operation_done.commit nullable
+TransactionErrorDetails.fatal name="non_fatal"
```
tintou, line 2: Can you also add `TransactionErrorDetails cprefix="FLATPAK_TRANSACTION_ERROR_DETAILS_"`

davidmhewitt: Oooh yes, that's nicer, thanks!

tintou (approved):
Code looks sane, I haven't tested it though

davidmhewitt:
I'm just going to merge this so we can test for regressions in the daily builds. I've tested that uninstalling flatpaks still works with this branch.

PR #1475, opened 2021-01-13
Icons: Standardize icon sizes

lainsce:
Makes app icons follow the progression of scaling that is 16, 32, 48, 64, 128.

Looks like in action:
| Homepage     | Listings | Single App View | Other Apps by Author |
| :---------: | :---------: | :---------: | :---------: |
| b | a | c | d |

### DIFF
--- a/data/styles/banner.css
+++ b/data/styles/banner.css
@@ -76,7 +76,9 @@
 .banner button arrow,
 .banner button image {
     color: @banner_bg_color;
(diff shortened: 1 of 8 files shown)


danrabbit:
It appears this only changes the size of badges, which are not app icons. Badges are used in list views for extension icons and for the updates icon.

Since this doesn't actually change anything for app developers, and 32px is so large as to obscure that icon that these are badging, I'm going to close this as "design conflict"

lainsce:
Alright! I just thought the badges were using app icons somewhere. :)

PR #1483, opened 2021-02-05
Update banner buttons for greenfield

### LINKED ISSUE #275: The 'open' button at the info page looks disabled.
When you visit an info page of an installed app, e.g. Calender, the 'Open' button looks disabled compared to the 'uninstall' button. The button does work and opens the app when clicked on it, though.

screenshot-appcenter-for-bug-report

#### Steps to reproduce:
1. Search for an installed app in AppCenter, e.g. Calender.
2. Open its info page and notice the hue difference between the 'Uninstall' and 'Open' button.

#### Expected behavior
Both buttons should have the same hue to avoid confusion.
 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/45291423-the-open-button-at-the-info-page-looks-disabled?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Will fix #275. Focused on elementary OS 6, but we should test on Pop! as well since they have the same issue downstream.

The basic idea is to just use the inverse app branding colors for all buttons, and then dim the non-suggested buttons a bit to make suggested buttons stand out better. Improvements to the exact styling are welcome, but keep in mind it should work with any sufficiently-contrasty bg and fg combinations, whether light or dark. 

- Gets rid of destructive styling for Uninstall buttons; it's not _that_ destructive, plus we were undoing the styling anyway. This is just less work than overriding it all and ensures uninstall buttons are styled as any other de-prioritized banner button.

- Always adds the suggested class to open and buy buttons; I don't think there's a practical reason to treat these differently.

- Removed `h3` styling from some buttons, which was causing mismatched font sizes for no real benefit imho.


data/styles/banner.css
```diff
@@ -48,41 +48,38 @@
...
-    background-image: none;
-    border-color: alpha(@banner_fg_color, 0.7);
-    box-shadow: none;
+    background-color: @banner_fg_color;
+    border: none;
+    box-shadow:
+        inset -1px 0 0 0 alpha(@banner_bg_color, 0.25),
+        inset 1px 0 0 0 alpha(@banner_bg_color, 0.25),
+        inset 0 -1px 0 0 alpha(@banner_bg_color, 0.25),
+        inset 0 1px 0 0 alpha(@banner_bg_color, 0.25);
```
danrabbit, line 57: I think it looks better to just not have box shadow tbh
```suggestion
    box-shadow: none;
```

data/styles/banner.css
```diff
@@ -48,41 +48,38 @@
...
-
-.banner button:focus {
-    background-color: alpha(@banner_fg_color, 0.9);
-    border-color: @banner_fg_color;
+.banner .linked button:last-child {
+    box-shadow:
+        inset -1px 0 0 0 alpha(@banner_bg_color, 0.25),
+        inset 0 -1px 0 0 alpha(@banner_bg_color, 0.25),
+        inset 0 1px 0 0 alpha(@banner_bg_color, 0.25);
 }
```
danrabbit, line 68:
```suggestion
```

data/styles/banner.css
```diff
@@ -48,41 +48,38 @@
...
 
 .banner button {
-    background-color: alpha(@banner_fg_color, 0.6);
-    background-image: none;
-    border-color: alpha(@banner_fg_color, 0.7);
-    box-shadow: none;
+    background-color: @banner_fg_color;
```
danrabbit, line 51: I would also get rid of the background-image tbh
```suggestion
    background-color: @banner_fg_color;
    background-image: none;
```

danrabbit (changes requested):
This looks good except I think it looks better to just not have the box shadow or background image

danrabbit (dismissed):
Barely even any CSS in here. Nice

cassidyjames:
@danrabbit so I think we still need _some_ separation between the button and its dropdown. Consider this:

Screenshot from 2021-02-10 17-39-04

It's hard to tell where the $5 button ends and the `v` button begins, especially compared to the combobox.

danrabbit (approved):
LFG :rocket:

PR #1485, opened 2021-02-08
AbstractView: Use Gtk.Stack instead of Hdy.Deck

### LINKED ISSUE #1482: Going back from OS Updates to Installed list prevents using mouse in list
## What Happened

If I'm checking out the details of an update by clicking on its listing

@2x

and then go back to the Installed list, the list loses focus, and I'm somehow prevented from using my mouse in the view.

Screenshot from 2021-02-04 13-37-07@2x

I can use the arrow keys to regain focus and select things, but the list doesn't respond to mouse clicks. I wonder if this is a regression from #1436, and/or could be an issue in Handy itself?

## Expected Behavior

Going back works as expected, the list stays focused, and I can use my mouse to select items in the list.

## Steps to Reproduce

1. Open AppCenter
2. Go to the Installed view
3. Click a listing (i.e. OS Updates)
4. Click the "Installed" back button at the top-left

Sometimes it takes going back and forth a few times for it to trigger, but once it triggers, it seems to be very reproducible.

### Logs

None at the moment, but can provide if it's useful.

## Platform Information

- elementary OS 6 Early Access (20210202 ISO, up-to-date as of now)
- AppCenter master

danrabbit:
Fixes #1482 

Seems safest to revert for now until we can find the source of the issue with Deck

### DIFF
--- a/src/Views/AbstractView.vala
+++ b/src/Views/AbstractView.vala
@@ -17,13 +17,14 @@
  * Authored by: Corentin Noël <corentin@elementaryos.org>
  */
 
-public abstract class AppCenter.AbstractView : Hdy.Deck {
+public abstract class AppCenter.AbstractView : Gtk.Stack {
     public signal void subview_entered (string? return_name, bool allow_search, string? custom_header = null, string? custom_search_placeholder = null);
     public signal void package_selected (AppCenterCore.Package package);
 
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
:+1:

PR #1486, opened 2021-02-10
Use Meson project name for icons

cassidyjames:
This just seems less error-prone since we're using the same string over and over again. We name the icons from the Meson project name to begin with, so let's ensure we're requesting the exact icons by using the same Meson variable.


danrabbit (approved):
Works for me

PR #1487, opened 2021-02-11
application.css: Use purple accent color

cassidyjames:
So this does the same thing we do in Music to set the orange accent, however: it looks like we might have to define all the weights for it to be used in places like suggested action buttons?

### DIFF
--- a/data/styles/application.css
+++ b/data/styles/application.css
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017–2018 elementary, Inc. (https://elementary.io)
+ * Copyright 2017–2021 elementary, Inc. (https://elementary.io)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


cassidyjames:
Yeah, so copying `grape.scss` from the stylesheet works:

```scss

/* Accent Shades */
@define-color accent_color_100 @GRAPE_100;
@define-color accent_color_300 @GRAPE_300;
@define-color accent_color_500 @GRAPE_500;
@define-color accent_color_700 @GRAPE_700;
@define-color accent_color_900 @GRAPE_900;

@define-color accent_color @GRAPE_500;
```

However, it means we actually lose the ability to do different accent colors in light and dark style. :thinking: This seems non-ideal and is an issue in any app we want to use accent colors. It makes me wonder if we should just have the app request the stylesheet variant directly, or something? Or maybe have a Granite util to hide the implementation from apps so they can just say, "gimme Grape" and we figure out what to do?

cassidyjames:
I think the proper way to do this would actually be to request the purple stylesheet itself.

Issue #1478, opened 2021-01-15
Navigating "back" to category views doesn't work


cassidyjames:
## What Happened

I was looking to install a random app to test a new Applications Menu branch, so I navigated by categories. When using the button to go back to the category's list view, the title of the page changed but the view didn't switch; I remained on the app info view.

## Expected Behavior

Navigating back from an app to a category's list view would work.

## Steps to Reproduce

1. Open AppCenter and click on a category, e.g. Music
2. Click on an app, e.g. the elementary Music app
3. Click the "← Music" back button in the headerbar

## Platform Information

up-to-date elementary OS 6 daily build

cassidyjames:
I think this may be a regression of #1436, since it deals with navigating through the stack/deck.

PR #1490, opened 2021-02-12
Fix using back button to return to category

ids1024:
Fixes https://github.com/elementary/appcenter/issues/1478.

e25c8c17b958cd96736194c057e53c723168a878 is responsible for the issue, according to a git bisect.


danrabbit (approved):
Nice one line fix! Good job tracking that down and thanks for the PR :)

PR #1491, opened 2021-02-15
Use `client.resolve_sync` in `get_package_internal`

ids1024:
`client.search_names_sync` may get a package that isn't an exact match. In testing on Pop!_OS, we noticed the `peek` package was incorrectly marked as uninstalled with the latest AppCenter.

It seems instead of `peek`, `client.search_names_sync` was getting `cardpeek` as the first result (just as `apt search peek` does).

This seems to be the correct function for this purpose.


davidmhewitt (approved):
Good spot! Thank you!

Issue #1432, opened 2020-10-13
Hide apps from the Ubuntu repos


danrabbit:
We've discussed this several times, so I wanted to make sure it's been tracked.

At this point I think it's become fairly clear that users are not getting a good experience from apps in the Ubuntu repos. We frequently here about poor experiences with Libreoffice (primarily because of the way it's packaged) and many other popular cross platform apps are several versions out of date.

To set expectations about these apps, we tried introducing a dialog that made it explicitly clear that these apps may not receive updates and aren't maintained by elementary. This did not go over well. It seems like most people don't read the contents of the dialog and they're very upset that we would throw a warning about apps they feel are well trusted

So, it seems like the best option for many of these apps is not to offer them in AppCenter at all. Especially as most of them are available as Flatpak. It would generally be a better experience for our users to get the latest version of cross platform apps from outside of AppCenter than to try to rework setting expectations for apps in the Ubuntu repos.

Bjargar:
how would that affect to newcomers?

PR #1492, opened 2021-02-20
Hide apps from the Ubuntu repos

meisenzahl:
Closes https://github.com/elementary/appcenter/issues/1432

## Before

Bildschirmfoto von 2021-02-20 08-02-19

## After

Bildschirmfoto von 2021-02-20 08-07-11

## Implementation

Most simple way to hide apps from Ubuntu repos based on https://github.com/elementary/appcenter/blob/ea7f524316302d76124a287e1e1a504e53eaeb9c/src/Core/Package.vala#L333-L343


hanaral:
This is obviously a good thing in the long term, but should it be done before releasing a new version of `io.elementary.BaseApp`?

meisenzahl:
@davidmhewitt thanks for looking into this and providing a solution. I have modified my implementation based on your suggestion and guarded it with a meson option.

cassidyjames:
It looks like this branch breaks the "More apps from…" footer at the bottom; at least in this branch, I no longer see apps from the same app developer, even when they are from Flathub.

davidmhewitt:
Yeah, I wonder why :laughing: 

https://github.com/elementary/appcenter/blob/0222a48a837eb99c5196dd80650db9142ca9f5ae/src/Core/FlatpakBackend.vala#L369-L371

meson_options.txt
```diff
@@ -5,3 +5,4 @@ option('name', type : 'string', value : 'AppCenter', description : 'The name of
 option('payments', type : 'boolean', value : true, description : 'Enable payment features and display paid apps')
 option('sharing', type : 'boolean', value : true, description : 'Display sharing features, i.e. copyable URLs to apps')
 option('api_url', type : 'string', value : 'https://developer.elementary.io/api', description : 'The base URL of the API to get app lists for the homepage banners')
+option('only_use_user_cache', type : 'boolean', value : true, description : 'Only use user cache and hide packages from system cache')
```
davidmhewitt, line 8: I think this option would be better called something like `hide_upstream_distro_apps`. Then have the description as something like "Used for hiding Ubuntu apps on elementary". Talking about system vs user cache in appstream is confusing as to what it actually does. So I think this would make it clearer to people packaging AppCenter for other distros.

davidmhewitt (approved):

Issue #1494, opened 2021-02-23
"Open" button on Installed toast doesn't open app


cassidyjames:
## What Happened

I started an install for an app (Glimpse from Flathub) within AppCenter, then kept browsing. Once Glimpse was done installing, I got the toast saying it was installed, so I clicked it. Nothing happened. I tried with a couple more apps and had the same result. The Open button on each app's info page did work as expected, though.

## Expected Behavior

The "Open" button in the installed toast should open the app, just like the "Open" button on the app info page.

## Steps to Reproduce

1. Kick off an install for an app
2. Navigate somewhere else in AppCenter
3. When it's done installing, hit the "Open" button in the installed toast

## Platform Information

elementary OS 6 Early Access, up to date with AppCenter from master

PR #1496, opened 2021-02-27
Throw a dialog on failure when trying to open an installed app

meisenzahl:
In https://github.com/elementary/appcenter/pull/1456 we added support to throw a toast when an app is installed.
Sometimes opening the application fails. For now this adds a dialog showing an error message.

Fixes https://github.com/elementary/appcenter/issues/1494

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -142,6 +142,20 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
                     last_installed_package.launch ();
                 } catch (Error e) {
                     warning ("Failed to launch %s: %s".printf (last_installed_package.get_name (), e.message));
+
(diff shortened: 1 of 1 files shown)


meisenzahl:
@cassidyjames I was not able to reproduce the error you described in https://github.com/elementary/appcenter/issues/1494:

https://user-images.githubusercontent.com/10796736/109381677-0e1f4280-78dc-11eb-9b1c-188feff69266.mp4

When I click open, Glimpse is opened in the background.

Could you please test this branch and share the error message you get.

cassidyjames:
So of course I can't reproduce this now… but it looks to me like a dialog would be good in case of future failures, as it could help us track it down.

danrabbit (approved):

PR #1502, opened 2021-03-10
AppListUpdateView: Inform instead of throwing restart dialog

### LINKED ISSUE #682: RestartDialog: get_action_area is deprecated

### LINKED ISSUE #1190: Canceling the system shutdown dialog shows a restart dialog
Currently only possible to test when #1048 is fixed.

1. Click the restart item in the infobar, and a shut down dialog opens
2. Canel the shut down dialog

Expected: nothing happens as it was canceled

Reality: a (fallback?) restart dialog opens

danrabbit:
Fixes #1190
Fixes #682 

It should generally be safe to assume the session restart interface exists. If it doesn't, rather than trying to handle restarting ourselves, we can ask the user to do it manually.

This also resolves an issue where we assumed any error we received from the restart interface was a failure. We now ignore cancellation.


jeremypw (changes requested):
This worked as expected with faked restart/error/cancel conditions.  The only thing I noticed was that the info bar changes height if there is a failure to restart (because the button is no longer visible).

danrabbit:
Hrm, I suppose we could make it insensitive instead, but is it really a bad thing that it changes height?

PR #1525, opened 2021-04-29
Package: Update APPCENTER_PACKAGE_ORIGIN

danrabbit:
Make sure we don't get those uncurated warnings for the appcenter flatpak repo


cassidyjames (approved):
dope

PR #1539, opened 2021-06-07
ScreenshotCache: don't bail if modified time is null

### LINKED ISSUE #1537: `ScreenshotCache.fetch` fails if screenshot host doesn't include Last-Modified header
Noticed this debugging why screenshots weren't showing in Pop!_Shop (an AppCenter fork) for the Popsicle package.

It seems getting the modification time for the `GFile` is returning `null`, presumably because `raw.githubusercontent.com` isn't returning a `Last-Modified` header.

I guess it shouldn't assume it's possible to get a modification time. It could just always re-download if there is no modification time. It could also take into account the `Expires` and `Cache-Control` headers, though presumably that would require fetching through `Soup` instead of `GFile`.

danrabbit:
Fixes #1537 

Same fix as merged downstream in PopShop https://github.com/pop-os/shop/pull/271

workaround for https://github.com/elementary/flatpak-platform/issues/34 as well

### DIFF
--- a/src/Core/ScreenshotCache.vala
+++ b/src/Core/ScreenshotCache.vala
@@ -192,11 +192,6 @@ public class AppCenterCore.ScreenshotCache : GLib.Object {
             warning ("Error getting modification time of remote screenshot file: %s", e.message);
         }
 
-        if (remote_mtime == null) {
-            // Used the locally cached version if it exists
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
Dope

PR #1558, opened 2021-07-22
Remap AppStream 'metadata' tag to 'custom'

### LINKED ISSUE #1555: Get custom values from appstream components
## What Happened

It doesn't appear that any custom values for an appstream component return non-null

## Expected Behavior

We should be able to get custom values like brand colors and stripe keys for flatpak apps

## Steps to Reproduce

1. Go to an app page for an app that has been published in AppCenter like Harvey
2. Notice that brand colors are not being used

You can also verify that this data is available in the local XML stored at ~`/.cache/io.elementary.appcenter/flatpak-metadata/user`

davidmhewitt:
Fixes #1555

It's a bit of a hacky workaround since it's not really AppCenter's problem that whatever is generating the collection metadata isn't being standards compliant, but it's a workaround that gets us out of a hole quickly if we want it.

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -827,8 +827,22 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
 
         root->set_prop ("origin", origin_name);
 
+        // Remap the <metadata> tag that contains the custom x-appcenter-xxx values to
+        // <custom> as expected by the AppStream parser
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
I can confirm that this works!

I wonder if we should add a `FIXME` here with a note about the upstream issue so that it's obvious we can remove this in the future?

davidmhewitt:
Sure, feel free to add that on my behalf, I'm probably away from my PC for the rest of the day now.

danrabbit (approved):

PR #1562, opened 2021-07-22
StripeDialog: Move privacy policy to secondary text

### LINKED ISSUE #1560: Privacy policy link isn't showing in payments dialog
## What Happened

The privacy policy link that we normally show in the payments dialog is missing

## Expected Behavior

This is probably something to do with action areas in dialogs. We could move this explainer to be under the email field since that is what it's most related to

danrabbit:
Fixes #1560


cassidyjames (approved):
Makes sense

PR #1563, opened 2021-07-22
Homepage: Use flatpak for recently updated packages

### LINKED ISSUE #1345: Calculate "recently updated" locally
In some of my random testing, I noticed that GNOME Software has a "Recent Releases" section that is calculated locally, and seemed to perform well. Since we have the requisite data locally, it might make sense to just do the same and avoid a remote server call here.

Screenshot from 2020-05-29 21-30-31

I know this is local because… again, in random testing that isn't generally recommended for users… I was using GNOME Software with the AppCenter remote added. AppCenter apps were showing up here as expected, correctly reflecting the recent releases in their AppData.

When we first designed AppCenter/Houston, we didn't enforce release descriptions and dates in AppData (they might not have even existed in AppStream, it was that long ago), but now we do.

danrabbit:
Fixes #1345

I'm not sure if this is the best way to do this, but it seems to work


cassidyjames:
@danrabbit hm, if I check `package.is_native` here it appears to work for me, showing GNOME Web and Harvey. :thinking: I'm not sure why Dippi isn't showing, though.

src/Core/FlatpakBackend.vala
```diff
@@ -262,6 +262,42 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+
+        user_appstream_pool.get_components ().foreach ((comp) => {
+            var packages = get_packages_for_component_id (comp.get_id ());
+            apps.add_all (packages);
+        });
+
+        system_appstream_pool.get_components ().foreach ((comp) => {
+            var packages = get_packages_for_component_id (comp.get_id ());
+            apps.add_all (packages);
+        });
```
davidmhewitt, line 276:
```suggestion
    public Gee.Collection<Package> get_native_packages_by_release_date () {
        var apps = new Gee.TreeSet<AppCenterCore.Package> (compare_packages_by_release_date);

        user_appstream_pool.get_components ().foreach ((comp) => {
            var packages = get_packages_for_component_id (comp.get_id ());
            foreach (var package in packages) {
                if (package.is_native) {
                    apps.add (package);
                }
            }
        });

        system_appstream_pool.get_components ().foreach ((comp) => {
            var packages = get_packages_for_component_id (comp.get_id ());
            foreach (var package in packages) {
                if (package.is_native) {
                    apps.add (package);
                }
            }
        });
```

Since you only really want the native packages from this method anyway, I'd be tempted to rename and adapt it to do exactly that

src/Views/Homepage.vala
```diff
@@ -184,19 +184,31 @@ namespace AppCenter {
...
         private async void load_banners () {
+            unowned FlatpakBackend fp_client = FlatpakBackend.get_default ();
+            var packages_by_release_date = fp_client.get_packages_by_release_date ();
+
+            foreach (var package in packages_by_release_date) {
+                if (recently_updated_carousel.get_children ().length () >= NUM_PACKAGES_IN_CAROUSEL) {
+                    break;
+                }
+
+                if (package.state == AppCenterCore.Package.State.NOT_INSTALLED) {
```
davidmhewitt, line 195: This is a bit more tricky since you might have a native package from a user installation and a system installation, so you need to loop over the packages in `package.origin_package` to get both and check they're _all_ not installed.

davidmhewitt (changes requested):

davidmhewitt:
Whoops, forgot to leave an explanation on my review:

Since `get_packages_for_component_id` returns _all_ of the Flatpak packages with a given component ID, so if you have Dippi from 3 remotes (AppCenter system, AppCenter user and Flathub), then you get 3 `Package` objects. One of those definitely isn't native as explained in #1564 .

Since you're only using your new method in one place, I suggest adapting it to only return native packages which solves that first issue. But then you have to solve the case of if you have multiple native packages, you need to detect if _at least one_ is installed (because of the systemwide and userwide remotes).

I think possibly, you'd also want to sort the userwide remote first so that when you click the package on the homepage, it pre-select the user one too.

You can detect if the package is from a userwide remote with this:
```
var fp_package = package as FlatpakPackage;
if (fp_package != null && fp_package.installation == FlatpakBackend.user_installation) {
// It's a user Package
}
```

src/Core/FlatpakBackend.vala
danrabbit: Makes sense to me :+1:

danrabbit:
@davidmhewitt Hey that seems to show a lot more packages now by moving the native check into the flatpak backend! Oddly, it doesn't appear that installed packages are being filtered out even when checking all of the origin packages. Packages that are installed are returning `false` for `installed`

davidmhewitt:
> Packages that are installed are returning `false` for `installed`

@danrabbit Right, the installed state of all the packages is loaded asynchronously in the background while AppCenter is starting to prevent too much slow startup, so you have to explicitly query the state on the packages you're interested in. I've pushed a fix.

cassidyjames:
Hm, one note is that I'm seeing the Flatpak Mail here even though I have the .deb Mail installed, and because we're defaulting to opening the AppCenter Flatpak version, it gets kind of confusing as to whether it's installed or not—you have to then explicitly choose the .deb version from the dropdown to see if it's installed.

It might not be a major issue, but it's kind of confusing.

davidmhewitt:
@cassidyjames Yep, forgot that case, should be resolved now

danrabbit:
@davidmhewitt thanks for all of your help here! This seems to be doing the right thing now as far as I can tell. I'm happy if everyone else is happy :)

cassidyjames (approved):

PR #1566, opened 2021-07-27
Release 3.6.1

cassidyjames:
Flatpak fixes


danrabbit:
Do we need to do anything with the flatpak branches before we release this? Should we also wait to verify that we don't need to change the location of the API endpoint for newly published apps?

cassidyjames:
@danrabbit nothing needs to change with the Flatpak branches, and the API endpoint for new apps still needs to be brought up; I _think_ it'll be the same endpoint as before but @btkostner would know. Either way I don't think it hurts to get the recently-updated stuff and Flatpak colors/pricing stuff working first.

danrabbit:
I don't think it hurts to wait to get confirmation on the API endpoint. I'd like to avoid making multiple releases in a row if possible

cassidyjames:
Re: API endpoint, it should end up being the same.

danrabbit (approved):
let's goooooo :rocket:

PR #1572, opened 2021-08-05
AppInfoView: Make screenshot carousel full page width

danrabbit:
Screenshot from 2021-08-05 12 50 24
Screenshot from 2021-08-05 12 53 00

unconstrains the screenshots a bit. On a large display you can see a couple at the same time

### DIFF
--- a/data/styles/arrow.css
+++ b/data/styles/arrow.css
@@ -24,7 +24,7 @@
         0 0 0 1px alpha (#000, 0.05),
         0 1px 3px rgba(0,0,0,0.12),
         0 1px 2px rgba(0,0,0,0.24);
-    margin: 6px;
+    margin: 12px;
     padding: 6px;
     color: @text_color;
 }
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -81,7 +81,8 @@ namespace AppCenter.Views {
 
             if (screenshots.length > 0) {
                 app_screenshots = new Hdy.Carousel () {
-                    height_request = 500
+                    height_request = 500,
+                    spacing = 128
                 };
 
                 screenshot_previous = new ArrowButton ("go-previous-symbolic") {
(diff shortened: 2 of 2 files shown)


cassidyjames:
Hmm, can we disable paging on mouse scroll wheel, then? Because as-is this makes it difficult to scroll down on the page with a traditional mouse.

cassidyjames (changes requested):
lol somehow the footer for apps from the same dev is borked in this branch

Screenshot from 2021-08-05 17 51 18

Screenshot from 2021-08-05 17 50 33

hanaral:
I'd like to point out that this makes it even more of a chore to just read the description :/
Making it full width also pushes the arrow buttons to the furthest reaches of the window, so I really don't think this is a great idea

cassidyjames (changes requested):
10ac307 is actually the opposite of what I think we need to do; now it's hard to scroll with a mouse OR touchpad unless you carefully position the mouse cursor below the screenshots ahead of time.

Instead, I think we need to disable _mouse wheel paging_ on the screenshots to allow scrolling down the page to work as expected. Besides that behavior, I'm in favor of opening up the layout and letting the screenshots be more prominent here.

danrabbit:
@cassidyjames ah okay, I see you. It looks like we're blocked until Handy 1.3 for the `allow-scroll-wheel` property

cassidyjames:
Blocked by #1589 and Handy 1.3

jeremypw:
@danrabbit Should this be closed as the issue #1578 is now fixed or do you want to make the carousel full-width anyway?

danrabbit:
@jeremypw yeah I still want to go this direction because of responsive. Related to #1744. Just gotta resolve conflicts etc

jeremypw (approved):
Code looks good, works as expected, with or without screenshots.

PR #1591, opened 2021-08-12
AbstractAppContainer: Remove is_os_updates property

danrabbit:
Only ever used once outside this class and `package` is publicly available so it's not really necessary


cassidyjames (approved):
Thank you for taking bites out of the lasagna

PR #1592, opened 2021-08-12
AbstractAppContainer: Reduce variable scope

danrabbit:
A few of these widgets are never used outside this class


cassidyjames (approved):
This makes more sense, and I'm gonna pretend I don't see my name on the git blame...

PR #1597, opened 2021-08-14
Application: Fix open button in toasts

### LINKED ISSUE #1533: Clicking "open" after downloading an app throws an error window (EOS 6 Beta)
When downloading an app from the app center then that small window, that has "open" button, pops up, the app center throws a small error window that says "failed to launch [app name] appinfo not found for package: [app name]", when I hit the "open" button in that notification. To open the app, I'd have to go to the apps menu and open it. If uninstall the app then reinstall it and do the same, it opens. It seems like it happens just when the app is downloaded for the first time. It has happened on the mail app and gnome web so far (I've removed gnome web and mail apps as a part of my testing).
Note: I, also, think "appinfo" should be corrected to "app info" instead.
Screenshot from 2021-05-15 19-33-51@2x
Screenshot from 2021-05-15 19-33-17@2x

meisenzahl:
Fixes #1533


meisenzahl:
@KarkanAlzwayed @cassidyjames it seems like this check gets the package ID. Can you confirm that this works for you?

cassidyjames (approved):
Yes, and a nice improvement!

PR #1619, opened 2021-08-21
Banner: reduce method scope

danrabbit:
* Remove single-use `set_default_brand ()`
* `next_package ()` and `set_background ()` are only used here so can be private


meisenzahl (approved):
LGTM

PR #1621, opened 2021-08-21
UbuntuDrivers: Error handling for kernel headers install

davidmhewitt:
Follow up to #1615 

There was an unhandled exception here. I've refactored it out to a single use method so it's easier to bail out with a return instead of wrapping the whole thing in a try/catch and having even more indentation.


JoseExposito (approved):
LGTM, thanks for fixing this 🚀

PR #1629, opened 2021-08-27
StripeDialog: move pay-what-you-can flow here

### LINKED ISSUE #307: Where to click to buy an app?
I opened the new App Center (Version: 0.2.1+r632+pkg53~ubuntu0.4.1.1) and clicked on the Nimbus app. I see the $1.00 price tag with the downward arrow next to it (see the screenshot), but it wasn't obvious to me that I had to click the price to purchase the app. To me it looked like a standard drop-down menu, where clicking on the text or on the arrow does the same thing.

I think it would be better to make it more obvious where is the action button. Maybe adding some text next to the price like "Purchase for $1.00" or "Install for Free"?
app center - 1

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/45395373-where-to-click-to-buy-an-app?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #307

Since we eventually want to switch to the authenticator, we have to make sure the whole pay-what-you-want flow can happen inside the payment dialog. As you can see, this also makes our lives a lot simpler in a bunch of other places

Screenshot from 2021-09-03 16 31 27


cassidyjames:
Since the developer is setting the suggested price, I think we should more heavily lean towards that design-wise, like how we used to have the button default to the suggested price and then an extra click for other prices. But I see what you were doing here with pre-filling the custom entry as well… hm.

I also wonder if this would be an opportunity to do something smarter with the suggested amounts, basing them off of the developer's suggested price, or if that's getting too clever. Maybe half the suggested price (rounded up to the next dollar), the suggested price, and double the suggested price? So for example, we could show:

```
Pay $2 for Principles

This is a one-time payment. Your 
email address is only used to send 
you a receipt. Privacy Policy

Pay What You Can
[ $1 ] [ $2 ] [ $4 ]  [ Custom… ]

[ Email                         ]
[ Card Number                   ]
[ MM / YY        ] [ CVC        ]

            [ Cancel ] [ Pay $2 ]
```

…with $2 pre-selected. For an app like Akira with a suggested price of $15, we could show $8, $15, and $30.

I like this more than just showing the default $1, $5, and $10 and then whatever the suggested price happens to be as it feels like it scales better with the developer's own valuation of their app.

There are probably edge cases here (like $1 suggested apps…) that we'd need to think through more.

danrabbit:
@cassidyjames yeah that sounds like kind of a bigger redesign than I intended to do here. I'd like to save logic changes for another branch and keep this scoped down to just adjusting the layout if needed

cassidyjames:
@danrabbit in that case, the most similar thing we can do here would be to put something like a HumbleButton on the suggested action button of this dialog, so the button itself has the developer's set price and then the popover shows the custom entries like before. Otherwise we're pretty significantly changing up how we treat the developer's suggested price.

danrabbit:
ooooooo okay that's interesting. Will do

danrabbit:
@cassidyjames alright so I started working on this and you can't pack a non-actionable (like a grid) into the action area in Gtk.Dialog, so that means I basically have to rewrite this dialog and it's a big thing that would be better done in small steps, imo. So I think having the goal of eventually having that popover in the action area is good, but it just seems like a lot for this branch

cassidyjames:
In that case I think we'll still need to do some work on how the prices are presented if we don't want to rework the dialog layout itself right now. I want to make sure we're presenting the developer's price up front as the clear default and then other options as a secondary thing _somehow_.

src/Dialogs/StripeDialog.vala
```diff
@@ -112,6 +144,17 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
         email_entry.primary_icon_name = "internet-mail-symbolic";
         email_entry.tooltip_text = _("Your email address is only used to send a receipt. You will not be subscribed to any mailing list.");
 
+        var email_label = new Gtk.Label (
+            _("Your email address is only used to send you a receipt.") +
```
cassidyjames, line 148: I think this can be more concise since it's directly associated with the email input

```suggestion
            _("Only used to send you a receipt.") +
```

src/Dialogs/StripeDialog.vala
```diff
@@ -94,16 +94,48 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
         primary_label.xalign = 0;
 
         var secondary_label = new Gtk.Label (
-            _("This is a one time payment. Your email address is only used to send you a receipt.") +
-            " <a href=\"https://stripe.com/privacy\">%s</a>".printf (_("Privacy Policy"))
+            _("This is a one time payment. You can also choose to pay a more affordable price.")
```
cassidyjames, line 97: I want to avoid accidentally implying this is unaffordable and also consider that someone might want to pay more than the sticker price. Maybe something like:

```suggestion
            _("This is a one time payment suggested by the developer. You can also choose your own price.")
```

or maybe

```suggestion
            _("This is a one time payment suggested by the developer. You can choose to pay it or select your own price.")
```

or just something along those lines

cassidyjames (changes requested):
Close as a step on the way to the authenticator. Just a couple of copy suggestions.

cassidyjames (approved):

PR #1641, opened 2021-08-31
edit-copy-symbolic -> edit-copy for visual style consistency

kdwk:
Made the copy icon in share menu not symbolic so it is visually consistent with other share icons like Reddit and Facebook


danrabbit:
Hey thanks for your branch! This is actually intentionally styled differently to try to make it more clear that this is an action and not another app

Closing as "Design conflict" since this is an intentional design choice and not a mistake :)

PR #1646, opened 2021-09-01
Banner: use accent for default colors

danrabbit:
Instead of using this kind of lame grey, use the user's accent color if none is set. Looks like:

Screenshot from 2021-09-01 16 10 20
Screenshot from 2021-09-01 16 09 31
Screenshot from 2021-09-01 16 12 53


cassidyjames (approved):
I like it! Works well in dark style as well.

PR #1648, opened 2021-09-01
Homepage: replace carousel with PackageRowGrid

danrabbit:
Screenshot from 2021-09-01 16 39 21

Gets us app descriptions and prices and gets rid of these pretty useless carousel classes


cassidyjames (approved):
:makeitso:

PR #1655, opened 2021-09-07
StripeDialog: rewrite entry validation

### LINKED ISSUE #897: Validation feedback in Stripe dialog
It seems like right now we don't have any feedback about individual fields passing client-side validation. So for example if you're filling out your details and you enter something incorrectly and the suggested action button is insensitive, it's unclear which field contains the error.

We do this in a way with the credit card field by showing an icon of the card which is nice affirmative feedback. This is missing for:
- [ ] Email
- [ ] Date
- [ ] CVC

We should show icons in the secondary position of the field when a form is client-side validated like we do when creating a new user:

screenshot from 2018-11-06 10 14 34 2x

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65447448-validation-feedback-in-stripe-dialog?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1652: StripeDialog doesn't handle $0
## What Happened

If I type $0 into the Stripe dialog in `master`, I still have to provide an email address and card number.

## Expected Behavior

It should disable and not require the irrelevant fields.

## Steps to Reproduce

1. Open AppCenter from `master`
2. Click the payment button on a paid app
3. Enter $0
4. Try to install

## Platform Information

OS 6 stable with AppCenter from `master`

danrabbit:
Fixes #1652 
Fixes #897

### DIFF
--- a/src/Dialogs/StripeDialog.vala
+++ b/src/Dialogs/StripeDialog.vala
@@ -125,8 +125,7 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
         };
 
         var selection_list = new Gtk.Grid () {
-            column_spacing = 6,
-            margin_bottom = 24
+            column_spacing = 6
         };
         selection_list.add (custom_amount);
         selection_list.add (or_label);
@@ -169,7 +168,8 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
 
         card_number_entry = new AppCenter.Widgets.CardNumberEntry () {
             activates_default = true,
-            hexpand = true
+            hexpand = true,
+            margin_bottom = 6
         };
         card_number_entry.bind_property ("has-focus", card_number_entry, "visibility");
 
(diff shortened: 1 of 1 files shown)


cassidyjames:
@danrabbit yeah I think revealer makes sense here. One weird thing is when typing `0`, you have to lose focus to get the revealer to hide which feels kind of awkward. When using the stepper buttons it seems to update instantly. Not sure if that's some default GTK thing, but I wonder if it's worth working around so when you type `0` it instantly hides the revealer.

danrabbit:
@cassidyjames So I was messing around with it and it seems like the input just isn't committed into the spinbutton until you press enter or lose focus. There is a method to manually update when text is inserted, but I could only either get it to segfault or it would be one keypress behind

cassidyjames (approved):
@danrabbit alright, in that case let's merge this as it makes the dialog usable, and we can revisit that if/when we redesign this for the authenticator.

PR #1661, opened 2021-09-08
FlatpakBackend: implement get_packages_by_author

danrabbit:


src/Core/FlatpakBackend.vala
```diff
@@ -405,7 +405,38 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+        var packages = new Gee.ArrayList<AppCenterCore.Package> ();
+        var package_ids = new Gee.ArrayList<string> ();
+
+        foreach (var package in package_list.values) {
+            if (packages.size > max) {
+                break;
+            }
+
+            if (package.component.id in package_ids) {
+                break;
```
davidmhewitt, line 417: This feels like it shouldn't be a `break` but a `continue`? Or else you'll end up bailing out if the same package id comes up twice, rather than when you've found as many as you need?

davidmhewitt (changes requested):

src/Core/FlatpakBackend.vala
danrabbit: Ah yes, thanks for catching that!!

davidmhewitt (approved):

PR #1662, opened 2021-09-08
Homepage: Use Hdy.Carousel for banner

### LINKED ISSUE #1657: Switch homepage banner to HdyCarousel
For swipey swipes

cassidyjames:
Fixes #1657. Minimal visual change for now, but adds multi-touch swiping and lets us do more interesting things later.

Before | After
---|---
Screenshot from 2021-09-10 15-20-35@2x | Screenshot from 2021-09-10 15-19-18@2x

- [x] Add actual app descriptions
- [x] Add icons
- [x] Re-do custom colors
- [ ] Re-use screenshot pager arrow buttons? Or clicking a non-centered banner switches to that banner.
- [ ] Figure out width/general styling/clamp
- [x] Clean up unused Banner/styling code
- [x] Ensure we're still handling offline/error edge cases (e.g. packing in a revealer and not showing if there's no data)


src/Views/Homepage.vala
```diff
@@ -301,4 +290,96 @@ public class AppCenter.Homepage : AbstractView {
...
+                margin = 64,
+                valign = Gtk.Align.CENTER
+            };
+
+            package_grid.attach (icon_image, 0, 0, 1, 3);
+            package_grid.attach (name_label, 1, 0);
+            package_grid.attach (summary_label, 1, 1);
+            package_grid.attach (description_label, 1, 2);
+
+            var css_class = package.normalized_component_id.replace (".", "-");
```
danrabbit, line 361: Since you're styling the widget instead of the screen it's probably not necessary to create a unique class per widget

src/Views/Homepage.vala
```diff
@@ -301,4 +290,96 @@ public class AppCenter.Homepage : AbstractView {
...
+
+            var style_context = get_style_context ();
```
danrabbit, line 363:
```suggestion
            unowned var style_context = get_style_context ();
```

data/styles/banner.css
```diff
@@ -28,25 +28,6 @@
...
-
-.banner.card:hover {
-    box-shadow:
-        inset 0 0 0 1px alpha(shade(@banner_bg_color, 1.7), 0.05),
-        inset 0 1px 0 0 alpha(shade(@banner_bg_color, 1.7), 0.45),
-        inset 0 -1px 0 0 alpha(shade(@banner_bg_color, 1.7), 0.15),
-        0 10px 8px -11px alpha(shade(@banner_bg_color, 0.6), 0.8),
-        0 8px 12px alpha(shade(@banner_bg_color, 0.8), 0.6);
-}
-
```
danrabbit, line 49: Why did we get rid of this stuff? It was a pretty cool hover effect imo

src/Views/Homepage.vala
```diff
@@ -301,4 +290,96 @@ public class AppCenter.Homepage : AbstractView {
...
+            style_context.add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
```
danrabbit, line 364: Why do these have suggested action?

src/Views/Homepage.vala
```diff
@@ -301,4 +290,96 @@ public class AppCenter.Homepage : AbstractView {
...
+
+            if (background_color != null && foreground_color != null) {
+                var css = CSS.printf (
+                    css_class,
+                    background_color,
+                    foreground_color
+                );
+                provider.load_from_data (css, css.length);
+
+                get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
danrabbit, line 381: Since you already cache the style context above you could re-use it :)

danrabbit (dismissed):
Since you're not using `banner` class here you've lost a lot of the styles for the background gradient and the overrides for default button styles

data/styles/banner.css
cassidyjames: Yeah fair. I was seeing if I could get away with using default button styles (and much simpler CSS without using variables) but I can bring the glow back.

src/Views/Homepage.vala
cassidyjames: It was for the fallback/default instead of using custom styles and overriding things, but to get the glow (and proper outlines) it looks like I'll need to use Gtk.CSS variables anyway.

cassidyjames:
Okay, this could be ready to go, or at least ready to review. Couple more notes:

- I had to remove the margin from the whole grid since I wanted banner_carousel to go edge-to-edge (otherwise it's awkwardly clipped 12px in from the sides). As a result, I had to tweak margins all over the place to get things lined back up. An alternate approach would be to pack each main section into its own grid, set the margins on the appropriate child grids, and then lean more on the column_spacing for the vertical margins. I didn't know if that was better or worse.

- I think we still want adjust the width of each banner card, maybe using Hdy.Clamp. But it might not be strictly necessary and could be a follow-up, so I thought I'd stop here for now and get feedback.

- Same with introducing back/forward arrows or possibly clicking an offscreen banner card to switch to that card. Might not be strictly necessary, so I thought I'd get feedback first.

danrabbit (changes requested):
I think at a minimum we should retain the auto-cycling behavior.

A simple way to deal with the banner width would be to hexpand. You won't be able to see any banners off screen, but at least it would fit the container without any magic

It seems like one fairly large regression here is clicking the dots doesn't navigate, which makes it kind of awkward to navigate with a mouse. I guess if it was auto-rotating that at least you could wait for the banner you wanted to pop up again

cassidyjames:
@danrabbit you can use a scroll wheel (sideways scroll or vertical) to page, but I agree the lack of paging by clicking the dots is a little annoying. My intent was to make clicking on an obscured button switch to that button, which I could still do, but obviously doesn't work if we just hexpand them.

I've added a timer. Still need to cancel the timer when hovering or at least when paging manually, otherwise that would get even more annoying.

danrabbit:
Yeah pausing the timer on mouse enter event sounds smart

cassidyjames:
@danrabbit okay, hexpanded and hooked up timer stop/start based on hover!

src/Views/Homepage.vala
```diff
@@ -25,44 +25,45 @@ public class AppCenter.Homepage : AbstractView {
...
+    private Gtk.Revealer banner_revealer;
     private Gtk.FlowBox recently_updated_carousel;
     private Gtk.Revealer recently_updated_revealer;
 
+
     construct {
-        switcher = new Widgets.Switcher () {
-            halign = Gtk.Align.CENTER
+        banner_carousel = new Hdy.Carousel () {
+            allow_long_swipes = true
```
cassidyjames, line 44: Thoughts on this @danrabbit? I'm kind of indifferent but I think you liked enabling long swipes in other contexts.

danrabbit:
This works really well, nice job! Don't forget to add a line to appdata, otherwise I say :shipit:

danrabbit (approved):

PR #1663, opened 2021-09-09
StripeDialog: Put the money icon in the right place depending on the locale

tintou:
In French we would write "5 $" and never "$ 5".


danrabbit:
I'm conflicted on this because we always put icons which describe the purpose of the entry in the primary position and it really messes up the layout to move it. It's more of a modifier to the entry itself than to number inside of the entry

tintou:
I feel like here the exception is that the icon is actually a label (using the same weight and font than a normal label) so it really looks weird for me as a french speaker.

danrabbit:
It's a different weight and font and spacing also :p 

Screenshot from 2021-09-15 10 52 16

cassidyjames (changes requested):
I'm kind of indifferent on the design; if the French person expects it on the right, I'm inclined to listen even if it looks funky to us. But I'm unable to confirm this actually uses the locale setting; I've set my locale to use French formats and rebooted, but the dialog in this branch still has it on the left.

tintou:
@cassidyjames we're not using LC_MONETARY and I really don't know any software that actually use that, we just rely on the language

PR #926, opened 2018-12-05
Content ratings

cassidyjames:
WIP (just a nonfunctional prototype for now). Will fix #34. We should get some nice icons as mentioned in #34. Design input _very_ welcome. :D

image 17

PR #1664, opened 2021-09-09
AppInfoView: Add content warnings

### LINKED ISSUE #34: Expose content rating
Currently GNOME Software is using https://odrs.gnome.org/oars
It seems to provide a wide range of granularity but more is better than less here

It would be nice to expose it to the user.

~A nice to have would be to be able to restrict certain categories in the Parental Control Plug~ See #629 

Launchpad Details: [#LP1665076](https://bugs.launchpad.net/bugs/1665076) Corentin Noël - 2017-02-15 18:55:59 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123507-expose-content-rating?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #34
Supersedes #926 

Screenshot from 2021-09-09 16 23 58

Screenshot from 2021-09-09 16 27 09

It's not super detailed or comprehensive, but it gets some info on the page.

Something to consider is if we want to remove the content warning dialog now that we show info on the page


cassidyjames (commented):
I don't _love_ this placement and layout, especially when many apps seem to just have one warning here. If we want it "above the fold" I think I'd like it in the header, but I imagine the header is really cramped and complicated with the loading bar and whatnot… so I'm unsure of the solution. I would be fine either shipping and iterating on this, or taking some time to think about the AppInfoView layout.

Regarding the explicit warning dialog, I think it's worth having unless/until we can provide some more context in these warnings; right now we would treat some extreme cases the same as mild which doesn't feel precise enough to convey the actual level of warning.

I also wonder if "Content Warnings" sounds a little _too_ scary. Just "Content" is probably too vague, but I wonder if we can go a different route than "warning."

danrabbit:
Yeah it's a bit weird in the header

Screenshot from 2021-09-10 14 54 14@2x

I would like to get _something_ in there and then iterate on it, personally. Most apps won't show content warnings at all, but it would be good to have a starting point. I'm sure we could do a lot more with styling and providing more accurate information, but something is better than nothing at all imo. Just adding in an halign seems to improve things quite a bit:

Screenshot from 2021-09-10 15 00 34@2x

"Content warning" is the language I've seen used on Mastodon for example :shrug: I don't think "warning" is an inherently scary word. Could maybe use something in the area of "considerations"?

cassidyjames:
Content Advisory is another one I've seen e.g. on Disney+ but that sounds super stiff. We can ship this and iterate. 🤷

cassidyjames (approved):

PR #1668, opened 2021-09-14
AppInfoView: fix content warning wrap

danrabbit:
Flowbox will still wrap if it needs to

Before
Screenshot from 2021-09-14 14 37 59

After:
Screenshot from 2021-09-14 14 35 58


src/Views/AppInfoView.vala
```diff
@@ -115,10 +115,11 @@ namespace AppCenter.Views {
...
 
             var oars_flowbox = new Gtk.FlowBox () {
-                halign = Gtk.Align.START,
-                column_spacing = 6,
+                column_spacing = 24,
                 row_spacing = 12,
-                margin_bottom = 24
+                margin_bottom = 24,
+                // Workaround for Flowbox preserving space around for items that don't exist
+                max_children_per_line = 1000
```
tintou, line 122:
```suggestion
                orientation = Gtk.Orientation.VERTICAL
```

tintou (changes requested):
Gtk.FlowBox implements Gtk.Orientable

src/Views/AppInfoView.vala
danrabbit: I want it to be horizontal though :sweat_smile:

tintou: That's how it is :man_shrugging: It depends if you're talking of the widget wrapping which will be vertical or the widget agglomeration with will be horizontal

danrabbit: The orientation isn't an issue here. The problem is that FlowBox is reserving `view-width`/`max-children` pixels for each child so if there are less children than `max-children`, it won't `halign` all the way left

cassidyjames:
Maybe I'm missing something, but: why not just use a grid instead of a FlowBox here?

danrabbit:
@cassidyjames We're already at 6 items and I have branches that will make it like 8 and plus translations, It will very very likely be possible to wrap in German :p

tintou (approved):

cassidyjames:
@danrabbit ah fair, I was commenting after seeing the max-per-line thing which I thought would also break wrapping. But that makes sense and makes more sense with your latest commits. :+1:

PR #1669, opened 2021-09-14
Oars socialinfo

danrabbit:
* Show descriptions for OARS ratings in a label rather than a tooltip
* Break out `social-info` into its own item with specific descriptions for each level

Screenshot from 2021-09-15 10 40 17

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -1068,25 +1068,31 @@ namespace AppCenter.Views {
 
     class ContentType : Gtk.Grid {
         public ContentType (string title, string description, string icon_name) {
+            orientation = Gtk.Orientation.VERTICAL;
             row_spacing = 3;
 
             var icon = new Gtk.Image.from_icon_name (icon_name, Gtk.IconSize.DND) {
+                halign = Gtk.Align.START,
                 margin_bottom = 6
             };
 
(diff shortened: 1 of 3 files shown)


cassidyjames:
Self-assigning because I want to poke at the layout here a bit

src/Views/AppInfoView.vala
```diff
@@ -1048,15 +1068,25 @@ namespace AppCenter.Views {
...
 
             var label = new Gtk.Label (title);
 
+            var description_label = new Gtk.Label (description) {
+                wrap = true,
+                max_width_chars = 25,
+                justify = Gtk.Justification.CENTER
+            };
+            description_label.get_style_context ().add_class (Granite.STYLE_CLASS_SMALL_LABEL);
+            description_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
```
cassidyjames, line 1085:
```suggestion

            unowned var description_label_context = description_label.get_style_context ();
            description_label_context.add_class (Granite.STYLE_CLASS_SMALL_LABEL);
            description_label_context.add_class (Gtk.STYLE_CLASS_DIM_LABEL);
```

cassidyjames (changes requested):
See inline nit

cassidyjames (approved):

PR #1670, opened 2021-09-14
Oars conflict violence

danrabbit:
Probably better after #1669 

Trying to separate a bit between like actual violence and just like situations with conflict of some kind. The OARS definitions here are a little fucky https://hughsie.github.io/oars/generate.html


cassidyjames (approved):
Makes sense :+1:

PR #1672, opened 2021-09-15
AppInfoView: more specific info for oars chat

danrabbit:
`internet-chat` isn't an fdo icon name, so we probably shouldn't rely on it existing

Screenshot from 2021-09-15 11 38 27


cassidyjames (approved):

src/Views/AppInfoView.vala
```diff
@@ -181,15 +175,41 @@ namespace AppCenter.Views {
                     oars_flowbox.add (gambling);
                 }
 
+                var social_chat_value = rating.get_value ("social-chat");
+                var social_audio_value = rating.get_value ("social-chat");
```
cassidyjames, line 179: Oops, did you mean

```suggestion
                var social_audio_value = rating.get_value ("social-audio");
```

danrabbit: Oops, thanks for catching that!

cassidyjames (approved):
Alright, should be good now :sweat_smile:

PR #1673, opened 2021-09-15
AppInfoView: Indicate multiplayer

danrabbit:
The presence of chat kind of implies multiplayer so maybe this should only show if this is exactly `MILD`? But I'm not sure. Is it important to explicitly call out when there is multiplayer but not chat?

Screenshot from 2021-09-15 11 50 19


cassidyjames:
The use case for calling out multiplayer minus chat is that other players can still do things you'd find mean/objectionable so I think it has merit. I do think I agree that it might make sense to only show this when == `MILD`.

cassidyjames (approved):
Eh we can figure out the exact ranking if/when we add support for values greater than mild. But for now this works.

PR #1679, opened 2021-09-15
Workflows: Test for Fedora and Pop configurations

danrabbit:
Add builds for downstream configurations to make sure we don't break stuff


danrabbit:
@decathorpe do you use other options besides `homepage=false`?

decathorpe:
Yes, I've been additionally using `-Dpayments=false -Dcurated=false`, though I'm thinking about changing at least some of those flags for Fedora builds in the near future.

.github/workflows/main.yml
```diff
@@ -21,9 +26,22 @@ jobs:
...
         meson build
-        ninja -C build
         ninja -C build install
-              
+
+    - name: Build (Fedora)
+      env:
+        DESTDIR: out
+      run: |
+        meson configure -Dcurated=false -Dhomepage=false -Dpayments=false  build
```
cassidyjames, line 35: Since the `homepage` build flag was dropped:

```suggestion
        meson configure -Dcurated=false -Dpayments=false build
```

.github/workflows/main.yml
```diff
@@ -21,9 +26,22 @@ jobs:
...
+        ninja -C build install
+
+    - name: Build (Pop!_Shop)
+      env:
+        DESTDIR: out
+      run: |
+        meson configure -Dcurated=false -Dpayments=false -Dsharing=false build
```
cassidyjames, line 42: Should also test with `name` build flag, even though it _shouldn't_ affect building.

```suggestion
        meson configure -Dcurated=false -Dpayments=false -Dsharing=false -Dname=Pop\!_OS build
```

cassidyjames (changes requested):

cassidyjames (approved):

PR #1683, opened 2021-09-15
AppInfoView: Show an icon for uncurated apps

### LINKED ISSUE #233: Expose non-curated app remote

danrabbit:
Fixes #233

Pretty wordy. I also wonder if we could come up with a better title than "Not Reviewed" but more obviously meaningful than just "Non-curated" which I feel like doesn't mean anything to most people. I've seen people complain that maybe the app is reviewed by whoever maintains the remote that it's from, so it's not necessarily true to say it hasn't been reviewed. But something indicating that it's there because of Sideload I dunno. Opening for discussion :)

Screenshot from 2021-09-15 14 01 15

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -106,6 +106,16 @@ namespace AppCenter.Views {
                 margin_bottom = 24
             };
 
+            if (!package.is_native) {
+                var uncurated = new ContentType (
+                    _("Non-Curated"),
+                    _("Not reviewed by elementary for security, privacy, or system integration"),
+                    "security-low-symbolic"
+                );
+
+                oars_flowbox.add (uncurated);
+            }
+
             var ratings = package_component.get_content_ratings ();
             for (int i = 0; i < ratings.length; i++) {
                 var rating = ratings[i];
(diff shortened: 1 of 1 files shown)


cassidyjames:
I am leaning toward the "Non-Curated" since that's the language we use elsewhere; otherwise I would want to change all of that language at the same time. And I think dropping the first sentence would be enough to make it not wordy but convey the same implications; I think trying to explain where it _did_ come from is probably outside the scope of this badge?

cassidyjames:
Also as a random thought, I'd probably stick this one first because it also affects the following data; if we didn't curate it, we can't guarantee the rest of that data, either—so it's the most important badge.

danrabbit:
@cassidyjames pushed

cassidyjames (commented):
Dope

cassidyjames (approved):

PR #1684, opened 2021-09-15
Change OARS icons namespace

danrabbit:
I'm not sure why we chose this super long-ass namespace, but this way the namespace isn't more than half the icon name, but mainly it stops us conflicting with old versions of icons in the system icon set (Gambling and Violence)


cassidyjames (approved):
👍

Issue #1682, opened 2021-09-15
don't show paid applications when compiling with `-Dpayments=false`


decathorpe:
It looks like the compile-time option `-Dpayments=false` does not hide paid applications, instead it only makes the "Install for $$$" buttons insensitive with an "Requires payments, which are not enabled" tooltip.

Since users usually cannot recompile AppCenter to enable payments, this message is completely non-actionable, and it would probably make more sense to just not show paid applications when compiled with `-Dpayments=false`.

decathorpe:
Side note: It appears that hiding paid applications is actually the intention behinf the `-Dpayments=false` option, at least according to its description in `meson_options.txt` (`option('payments', type : 'boolean', value : true, description : 'Enable payment features and display paid apps')`), so either paid applications appearing in prominent places on the homepage (carousel, recently updated, etc.) is a bug, or the description of the meson flag is outdated.

decathorpe:
Hm. Looks like this was either a caching issue, or it has been resolved in the meantime - because I cannot reproduce this issue any longer. :100:

PR #1686, opened 2021-09-16
Remove `homepage` flag in favor of `curated`

### LINKED ISSUE #1680: homepage carousel and "recently updated" broken unless "appcenter" flatpak remote is present
I tried re-enabling the homepage feature for Fedora builds of AppCenter, but I encountered a weird issue. If I only dropped `-Dhomepage=false`, then nothing changed, except that appcenter now logged a lot of error messages:

```
(io.elementary.appcenter:3731): Handy-CRITICAL **: 22:42:59.170: hdy_carousel_box_get_nth_child: assertion 'n < hdy_carousel_box_get_n_pages (self)' failed
```

And every few seconds, this one:

```
(io.elementary.appcenter:3731): Handy-CRITICAL **: 22:50:30.149: hdy_carousel_box_scroll_to: assertion 'HDY_IS_CAROUSEL_BOX (self)' failed
```

Additionally, no carousel or "recently updated" section is installed on the homepage unless I add the `appcenter` flatpak remote on the system. Then, restarting appcenter, those sections appear, and the first error message disappears (which makes me think that the code doesn't like it when the carousel has zero pages, or something like that). Removing the flatpak remote again and restarting AppCenter makes both sections disappear from the homepage again.

On Fedora, I cannot install third-party repositories for users, so it would be great if AppCenter would keep basic functionality even if the flatpak remote was not set up.

---

This happens in AppCenter from git master, compiled with `-Dcurated=false -Dpayments=false`, on Fedora 34 and 35.

cassidyjames:
Fixes #1680. Since we no longer hit an external API on the homepage, this PR:

- Removes the `homepage` compile flag and conditionals on the homepage
- Renames `get_native_packages_by_release_date ()` to `get_featured_packages_by_release_date ()`
- Uses the `curated` flag directly in the `get_featured_packages_by_release_date ()` method
- Updates the descriptions of some Meson options while we're here to be more informative

As a result, on Fedora you'll now get a homepage with banners of the most recently updated apps as well as the Recently Updated carousel with 12 more apps. This makes the homepage much more useful for downstreams while also reducing the maintenance burden for us by no longer doing conditional layout things.

If we do end up hitting an external API for featured apps, we could still use the `curated` flag to determine that behavior.


danrabbit:
This looks good to me. Nice simplification. If this makes @decathorpe happy, I'm more than +1 here :)

decathorpe:
I'll try this branch on Fedora later today. If everything works as advertised in the PR description, I'll be very happy indeed 😁

danrabbit:
I can confirm that this works as expected with default options and uses all the flathub stuff with curated=false

Screenshot from 2021-09-16 10 21 44

danrabbit (approved):

decathorpe:
There's some rough edges I see with the current state:

- compiling with `-Dcurated=false` lists any packages from the official Fedora repositories for things that are also available from any flatpak remote as "Unknown Origin (non-curated)", which is misleading at best. additionally, the carousel and "recently updated sections" still only feature applications from the "appcenter" flathub remote if it's present on the system, which kinda defeats the whole `curated=false` option

Screenshot from 2021-09-16 21-55-12
(here, any "Unknown Origin (non-curated) applications are actually official Fedora packages that are also alternatively available as flatpaks)

Screenshot from 2021-09-16 21-55-16
(here, you can see that in the case of Evince, three sources are listed, where the "Unknown origin" one is actually the "most officially supported" version of the application - the package from the official Fedora repositories - while the "fedora" and "appcenter" versions are the flatpak alternatives)

- even though I compiled with `-Dpayments=false`, paid applications are still shown if the "appcenter" flatpak remote is enabled (contrary to what the meson_option(s.txt) option claims), not sure why no paid applications are shown in the screenshot from elementary OS

Screenshot from 2021-09-16 21-54-52
(this is with the appcenter flatpak remote enabled, not sure why the carousel and "recently updated" section show only "curated" applications, even though that option is disabled)

Screenshot from 2021-09-16 22-00-18
(this is with the appcenter flatpak remote disabled, this looks like everything works as expected when it's not present)

Though this behaviour might even be a good thing, since people would need to enable the "appcenter" flatpak remote manually, and if they don't, appcenter will feature only system applications.

cassidyjames:
@decathorpe hm, most of those points should not be related to this branch, as I didn't touch any of those bits. Though you seeing only AppCenter apps when `curated` is false doesn't make sense to me. Is it possible those AppCenter apps all have more recent updates than other packages that would show here?

cassidyjames:
I think even given @decathorpe's points, this is a better result for Fedora and downstreams than `master`, isn't it? The homepage is more useful and engaging.

decathorpe:
> @decathorpe hm, most of those points (e.g. about the origins, app lists, and AppInfoViews) should not be related to this branch, as I didn't touch any of those bits. Those should be filed separately (like you did with #1682).

Right, I just came across those issues because I only recently re-enabled the "homepage", some might not be related to the branch. Ignore that which is not related to this PR :) I'll file issues separately, where applicable.

> All this branch touches is the homepage and explicitly not performing an `is_native` check when `curated` is false. So only AppCenter apps when `curated` is false doesn't make sense to me. Is it possible those AppCenter apps all have more recent updates than other packages that would show here?

It is possible, but rather unlikely - unless the "Recently updated" section only lists flatpak applications (which might not have been recently updated) and no native applications at all (which definitely were recently updated in the Fedora 35 development branch, at least all the GNOME 41 updates are all very recent).

decathorpe:
> this is a better result for Fedora and downstreams than master, isn't it?

I agree, it's defintely an improvement over the status quo. 👍🏼 💯

cassidyjames:
>unless the "Recently updated" section only lists flatpak applications (which might not have been recently updated) and no native applications at all

Ah, indeed, this is implemented in the Flatpak backend. AppCenter is largely moving to be focused on Flatpak, but I can see the confusion if you're not using the `hide_upstream_distro_apps` build flag. :stuck_out_tongue: Since this is still a better state than `master`, I'm in favor of merging now and filing issues for anything that still seems off.

PR #1690, opened 2021-09-17
meson: require libhandy-1 >= 1.1.90

decathorpe:
The allow-long-swipes property of HdySwipeTracker that has been used since PR#1662 was added to libhandy with version 1.1.90:

https://gitlab.gnome.org/GNOME/libhandy/-/blob/master/NEWS


davidmhewitt (approved):
Thank you!

PR #1692, opened 2021-09-17
AppInfoView: Set banner provider for context, not screen

danrabbit:
This removes the styles for the combobox because technically the button is a child of the combobox. I'm sure there's some get_child fuckery we could do to get those styles back if we really wanted to, but it kind of looks fine to me?


cassidyjames (changes requested):
Hm, looks like we're missing the styling on other buttons in the banner which looks either weird or broken depending on which one it is:

Screenshot from 2021-09-17 15-01-50@2x

Screenshot from 2021-09-17 15-02-02@2x

Screenshot from 2021-09-17 15-02-07@2x

danrabbit:
Oops you're right. My bad

cassidyjames:
(I am fine with the combobox style, though, and even the cancel button style honestly)

danrabbit:
@cassidyjames sorry for taking so long to fix this! All the action buttons should be styled now

zeebok (approved):

PR #1693, opened 2021-09-17
AppInfoView: load_more_content on realize

danrabbit:
Sometimes I've experienced an issue where the method is trying to pack widgets into the carousel before it's completely constructed so we get warnings about how the toplevel container can't hold more than one widget

I'm not sure why this is a timeout waiting for the end of the transition and why we're passing in the cache from `AbstractView`. Is there a significant difference between this and connected to `realize`?


davidmhewitt (approved):
Yeah, this seems way better! I can't find any obvious regressions.

Issue #1665, opened 2021-09-10
Homepage scrolls horizontally at small sizes


cassidyjames:
## What Happened

Noticed when working on #1662 but is present in master. When you resize the window to a small size (e.g. the minimum), it looks like the Recently Updated flowbox is making the page scroll horizontally.

## Expected Behavior

The contents adapt to smaller widths and don't overflow

## Steps to Reproduce

1. Open AppCenter
2. Resize the width down horizontally
3. See the scrollbar

## Platform Information

OS 6 with AppCenter from master

danrabbit:
Fixed in 1694

PR #1694, opened 2021-09-17
Homepage: don't horizontally scroll

cassidyjames:
One way of fixing #1665. This simply stops the page from scrolling horizontally, which will force the window to fit its contents. It fixes the visual weirdness so is a first step, but the root cause is that three-wide in the recently updated flowbox is too wide at smaller sizes.

However, setting the flowbox's minimum to 2 makes the children expand to a weirdly wide size, so I don't know if that's what we want to do or if there's something smarter to be done. Regardless, this is probably a good first step.

Renamed the `category_scrolled` while I was here because it's not just for the categories, it's for the whole homepage and that was confusingly named when I was first investigating it. :) Alternatively could name it `homepage_scrolled` if that's more clear.

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -23,7 +23,7 @@ public class AppCenter.Homepage : AbstractView {
     private const int MAX_PACKAGES_IN_CAROUSEL = 12;
 
     private Gtk.FlowBox category_flow;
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Horizontal scrolling should just be disabled by default in GTK lol

PR #1701, opened 2021-09-30
Package: centralize description string handling

### LINKED ISSUE #1700: Crash when an app doesn't have a description
### What Happened?

With `curated` set to `false`, apparently apps from other remotes can have no description. So we get a crash:

```
(io.elementary.appcenter:7793): GLib-CRITICAL **: 09:55:49.009: g_strsplit: assertion 'string != NULL' failed
```

### Steps to Reproduce

1. Have the Flathub remote
2. Compile with `-Dcurated=false`
3. Try to open AppCenter

It can be hit or miss depending on what was most recently updated in the Flathub remote, but it's possible to get a segfault

### Expected Behavior

No segfault lol

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

```shell
(io.elementary.appcenter:7793): GLib-CRITICAL **: 09:55:49.009: g_strsplit: assertion 'string != NULL' failed
```

### Hardware Info

_No response_

cassidyjames:
Fixes #1700


src/Widgets/Banner.vala
```diff
@@ -57,11 +57,14 @@ public class AppCenter.Widgets.Banner : Gtk.Button {
...
+        if (package.get_description () != null) {
+            string[] lines = package.get_description ().split ("\n");
+            description = lines[0].strip ();
+
+            for (int i = 1; i < lines.length; i++) {
+                description += " " + lines[i].strip ();
+            }
         }
 
         int close_paragraph_index = description.index_of ("</p>", 0);
```
davidmhewitt, line 70: Not directly related to your PR, which seems logical, but I'm not a fan of this `index_of` a thing that might not be there, and the subsequent hardcoded `slice` either.

I think I'd like to have this PR as an opportunity to at least comment why half of this stuff is happening or fix it up

davidmhewitt (changes requested):

cassidyjames:
@davidmhewitt it looks like `AppStream.markup_convert_simple ()` does a lot of what we were trying to do manually, so I simplified a bit to use that and added a comment about grabbing only the first line for the banner. Hopefully this makes more sense?

cassidyjames:
Actually, this doesn't quite do all the same things; if someone has

```html
<p>
    This is line             one of the description!
    And this is another line but should be part of the same paragraph.
</p>
```

This doesn't strip the internal whitespace (like HTML would), and it still treats the two lines as separate lines. Since we always want to follow the same cleaned-up formatting rules here, I think it makes more sense to implement whatever transforming of the string we want directly in `Package.get_description ()`, and then we can more clearly comment and step through each part so it's much simpler to just grab and display that description wherever we want.

If that approach makes more sense, I'll pursue that instead.

src/Core/Package.vala
```diff
@@ -626,13 +625,28 @@ public class AppCenterCore.Package : Object {
...
             description = component.get_description ();
+
             if (description == null) {
                 if (backend_details == null) {
                     populate_backend_details_sync ();
                 }
 
                 description = backend_details.description;
             }
+
```
davidmhewitt, line 636: Add a `description == null` check here and return `null` if so. It's possible that a package still won't have a description even after trying two different ways to get it.

davidmhewitt (changes requested):

src/Core/Package.vala
davidmhewitt: Worth noting that this change would resolve the segfault that @danrabbit was seeing when opening the "OS Updates" component while AppCenter was still loading.

danrabbit (approved):
@davidmhewitt do you still have changes you'd like to see here?

I can confirm that the crash I was experiencing has been resolved

davidmhewitt (approved):

PR #1708, opened 2021-10-06
AbstractAppContainer: Use ProgressButton

### LINKED ISSUE #1653: Loading bar in PackageRow breaks compact layout
## What Happened

If I install an app from the homepage, the loading bar expands the PackageRow widget and breaks the layout

While installing | After installed
---|---
Screenshot from 2021-09-07 10-11-28 | Screenshot from 2021-09-07 10-11-30

## Expected Behavior

The layout shouldn't break when installing apps. :) We've previously discussed replacing the progressbar somehow, and moving progress to a separate place (like a popover on the loading spinner in the headerbar, or at the top of the "installed" page).

## Steps to Reproduce

1. Open AppCenter and resize to a skinny width
2. Install an app from the homepage

## Platform Information

OS 6 stable with AppCenter from `master`

### LINKED ISSUE #1698: The title should be fully legible
Screenshot
That is for me, without ellipsis: `Operating System Updates`.
I use a up-to-date elementary OS 6.

cassidyjames:
This adds a progress bar to the button instead of packing a whole Gtk.ProgressBar. As a result, the UI doesn't resize. Fixes #1653, fixes #1698.

I initially investigated using a [Dazzle.ProgressButton](https://valadoc.org/libdazzle-1.0/Dazzle.ProgressButton.html), but in testing, it wasn't styled correctly (and it seemed overkill to add Dazzle as a dependency for something relatively simple). Instead, this branch makes a new simple ProgressButton using a similar approach to that of Dazzle.ProgressButton.

- [x] Are we okay putting the technical details in a tooltip?
- [x] Check/add RTL support (Dazzle looks to have this in its custom styling)
- [x] Decide if we should use the system `@accent_color` or something else

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -266,13 +266,15 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
         var apps = new Gee.TreeSet<AppCenterCore.Package> (compare_packages_by_release_date);
 
         foreach (var package in package_list.values) {
+            if (!package.is_explicit && !package.is_plugin) {
 #if CURATED
-            if (package.is_native) {
-                apps.add (package);
-            }
+                if (package.is_native) {
+                    apps.add (package);
+                }
 #else
-            apps.add (package);
+                apps.add (package);
 #endif
+            }
         }
 
         return apps;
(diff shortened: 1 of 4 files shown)


cassidyjames:
@danrabbit alright, RTL is handled and I pulled the non-changing CSS out into a GResource so we're only changing the one line. Looks and behaves the same as before (with the addition of RTL support).

PR #1722, opened 2021-10-18
AppInfoView: use contrasting color instead of primary_text

cassidyjames:
An experiment in reducing the metadata needed, in interest of https://github.com/ximion/appstream/issues/187

This changes AppInfoView to only care about the primary color, automatically selecting an appropriate [`Granite.contrasting_foreground_color ()`](https://valadoc.org/granite/Granite.contrasting_foreground_color.html) for the text drawn over the background.

In looking through apps on AppCenter, this seems to cover the majority of use cases. Some apps that used a neutral background with a saturated foreground would be encouraged to set the primary color to the saturated color, but otherwise this doesn't cause any overt breakage.

### DIFF
--- a/data/styles/screenshots.css
+++ b/data/styles/screenshots.css
@@ -1,19 +1,19 @@
 /*
- * Copyright (c) 2019 elementary, Inc. (https://elementary.io)
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


danrabbit:
Closing since there are conflicts etc. Submitted a new pull with the same idea as https://github.com/elementary/appcenter/pull/1880

PR #1723, opened 2021-10-22
AppInfoView: Change app description to Gtk.Label

isantop:
This was originally a Gtk.Label in the very beginning, but early on was changed to be a Gtk.TextView in e37fc748b4693d64d48a119f532e181eb3dfd3c7

While working on changes to make AppCenter function at narrow window widths, I noticed that some applications (Notably Veloren in Flathub) had description widgets that prevented the window from shrinking below a minimum width which was very wide and unsuitable for narrow windows. Changing the App description to a label seems to prevent this, however, I was unsure of the original reasons for the change to a TextView


cassidyjames (approved):
@danrabbit pointed out it was due to manual parsing of markup (before we were able to use AppStream to do the same and generate Pango markup); see: https://github.com/elementary/appcenter/commit/c8261a889be91561ab045df3777e2862c7444123. So this should be good, and works nicely for me. Thanks!

cassidyjames:
@isantop ah can you merge master here? It looks like you didn't give repo owners permissions to write to your fork.

isantop:
@cassidyjames I updated the branch based on the current master branch here (f6b4a32cbc401bc658eb13b2cff03b28890bbecf)

cassidyjames (approved):

PR #1729, opened 2021-10-28
Changes to support narrower window Widths

isantop:
This PR enables AppCenter to shrink to much narrower window widths (down to 500px) without compromising usability and without affecting the design and layout at larger widths. It does this by:
  * Shrinking the window's size_request to allow the window to shrink to narrower widths.
  * Changing the minimum number of items in the Category and Recently Updated FlowBoxes 
  * Changes to the AppView to help ensure that Recently Updated Apps don't try and take more space than is required (ensuring they can reflow within the FlowBox correctly)
  * Makes the screenshot Carousel full-width, enlarging the screenshots and making them more usable at narrower sizes.
  * Sets up a custom AppScreenshot widget which allows screenshots to scale live in realtime and clamps them to a reasonable maximum width.
  * Arranges the AppInfoView header into a FlowBox to allow the Action buttons to flow underneath the App metadata at narrow widths
  * Shrinks the Search entry at narrow window widths to make more room for labels and the view switcher.

These changes greatly improve usability on very small screens and in tiling window managers such as with Pop! COSMIC and Pop Shell.

Requires #1723 Because of fixes to text wrapping in that PR


danrabbit:
Closing since we were breaking this up into smaller branches

Issue #1076, opened 2019-09-18
Flag for Flatpak-only


cassidyjames:
@sysrich had expressed interest at GUADEC in using a Flatpak-only AppCenter for a project he was working on. It could be useful to provide a flag to disable the PackageKit backend and only talk to Flatpak.

Conan-Kudo:
If the AppCenter backends (PackageKit and Flatpak) were structured as plugins, it'd be very easy to just partition it out and ship variations that have it and don't have it. This is how GNOME Software and Plasma Discover do it, and that makes it a lot easier to deal with the variation without building it differently each time.

tetebueno:
In order to mitigate this in the meantime, is there a way te tell if an application is being offered by Flatpak in plain sight?

cassidyjames:
@tetebueno currently if the app is only offered via Flatpak (or the non-Flatpak source has a non-matching ID, meaning we can't associating them), you can see an `(i)` icon next to the install size because Flatpak sizes are estimates that don't fully take into account the diff-based nature of their downloads. It's not ideal (or intended to be the way to identify Flatpaks), but it works at the moment.

Fixing #233 might also help better expose the source when only one exists.

PR #1739, opened 2021-11-01
build: add packagekit_backend meson option

bobby285271:
Sorry for this weird merge request, but sadly appcenter doesn't work on NixOS currently because of the current status of the packagekit on NixOS is not that matured. Excluding PackageKitBackend and UbuntuDriversBackend resolves most issues and at least we can search / install / update / remove some flatpak applications now after adding some flatpak remotes.

|  |  |  |  |
|:----:|:----:|:----:|:----:|

This is probably not a solution for https://github.com/elementary/appcenter/issues/1076, however it will be great if appcenter can work on NixOS.

And very sorry I am adding 8d75bb9a6edcc7fb6fe9791df00310fce103adc6 as some (but not all) app icons failed to load on both banner and recently updated section and I have no idea for this yet (and seems the issue does not happen elsewhere, see above screenshots, and I noticed that when I keep the cache in `~/.cache/io.elementary.appcenter` then launch the original appcenter with all backends, all icons load). If the issue can be reproduced, will be appreciated if anyone can hack on it, otherwise I am also happy to drop that.

Not sure if this is related:

```
Gtk-WARNING **: 19:13:31.983: Could not load a pixbuf from icon theme.
This may indicate that pixbuf loaders or the mime database could not be found.
```

|  |
|:---:|

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -107,13 +107,9 @@ public class AppCenter.Homepage : AbstractView {
             column_spacing = 24,
             orientation = Gtk.Orientation.VERTICAL
         };
-#if PACKAGEKIT_BACKEND
         grid.add (banner_revealer);
         grid.add (recently_updated_revealer);
         grid.add (categories_label);
-#else
-        category_flow.margin_top = 12;
-#endif
         grid.add (category_flow);
 
         scrolled_window = new Gtk.ScrolledWindow (null, null) {
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -107,9 +107,13 @@ public class AppCenter.Homepage : AbstractView {
             column_spacing = 24,
             orientation = Gtk.Orientation.VERTICAL
         };
+#if PACKAGEKIT_BACKEND
         grid.add (banner_revealer);
         grid.add (recently_updated_revealer);
         grid.add (categories_label);
+#else
+        category_flow.margin_top = 12;
+#endif
         grid.add (category_flow);
 
         scrolled_window = new Gtk.ScrolledWindow (null, null) {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,6 +42,13 @@ jobs:
         meson configure -Dcurated=false -Dpayments=false -Dsharing=false -Dname=Pop\!_Shop build
         ninja -C build install
 
+    - name: Build (NixOS)
+      env:
+        DESTDIR: out
+      run: |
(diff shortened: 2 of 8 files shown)


src/Views/Homepage.vala
```diff
@@ -99,9 +99,13 @@ public class AppCenter.Homepage : AbstractView {
             column_spacing = 24,
             orientation = Gtk.Orientation.VERTICAL
         };
+#if PACKAGEKIT_BACKEND
```
davidmhewitt, line 102: I don't understand these changes to `Homepage`. The banners/carousels shouldn't have any dependency on PackageKit as they display Flatpak packages anyway. So a PackageKit related build flag shouldn't be affecting things that don't relate to PackageKit. What's the issue you're seeing while these lines are still enabled?

davidmhewitt: Sorry, I've just seen your description in the PR description. I think this probably needs some debugging rather than just disabling this code, because this is unrelated to PackageKit.

bobby285271: Actually I also think this is weird and maybe I need someone to reproduce the issue I mentioned above (as shown in the last screenshot)...🤯 @davidak Can you try?

bobby285271: I guess I will just revert this for now...🤦

bobby285271: Is it possible not to remove `~/.cache/io.elementary.appcenter/flatpak-metadata/*` on every launch?

https://github.com/elementary/appcenter/blob/aa29793377be481c02a7b88cf3aa1e398a4d4f16/src/Core/FlatpakBackend.vala#L665

bobby285271:
No longer needed after https://github.com/NixOS/nixpkgs/commit/e717c594abb80c90bfbebe61c1d47387906e2d1a and https://github.com/NixOS/nixpkgs/commit/59dc15a3d8678332e70086d568695aea0fe24cfe.

P.s.: We are now installing AppCenter by default if Flatpak support is enabled (https://github.com/NixOS/nixpkgs/commit/bb357d8203cc901bf1e729ec544c8c8b52cc329d). Packagekit with Nix backend should work after https://github.com/PackageKit/PackageKit/commit/e8140563ec7f78dd723068396dcad40e47580379 and https://github.com/NixOS/nixpkgs/commit/92a6ad8626eb08dbf76fa389d595c7ac7accd70e, but as mentioned in https://github.com/NixOS/nixpkgs/commit/ea611d2e171f3f1722faa7d5cd156c0ff6bef854 I believe appstream metadata is still needed to use Packagekit backend of AppCenter.

PR #1742, opened 2021-11-03
Homepage: Allow AppCenter Homepage to shrink to narrow sizes

isantop:
Updates the layout for ListPackageRowGrid to better wrap at narrow sizes and to reflow at larger sizes. This allows the Homepage flowboxes to each be set at 1 min children per row, allowing the Homepage to shrink to much narrower sizes.

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -71,7 +71,7 @@ public class AppCenter.Homepage : AbstractView {
             row_spacing = 12,
             homogeneous = true,
             max_children_per_line = 5,
-            min_children_per_line = 3
+            min_children_per_line = 1
         };
 
         var recently_updated_grid = new Gtk.Grid () {
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -70,8 +70,7 @@ public class AppCenter.Homepage : AbstractView {
             column_spacing = 12,
             row_spacing = 12,
             homogeneous = true,
-            max_children_per_line = 5,
-            min_children_per_line = 1
+            max_children_per_line = 5
         };
 
         var recently_updated_grid = new Gtk.Grid () {
--- a/src/Widgets/AppContainers/AbstractAppContainer.vala
+++ b/src/Widgets/AppContainers/AbstractAppContainer.vala
@@ -111,7 +111,7 @@ namespace AppCenter {
             action_button_group.add_widget (open_button);
 
             action_stack = new Gtk.Stack ();
-            action_stack.hexpand = true;
+            action_stack.hexpand = false;
             action_stack.hhomogeneous = false;
             action_stack.transition_type = Gtk.StackTransitionType.CROSSFADE;
             action_stack.add_named (button_grid, "buttons");
--- a/src/Widgets/AppContainers/AbstractAppContainer.vala
+++ b/src/Widgets/AppContainers/AbstractAppContainer.vala
@@ -111,7 +111,6 @@ namespace AppCenter {
             action_button_group.add_widget (open_button);
 
             action_stack = new Gtk.Stack ();
-            action_stack.hexpand = false;
             action_stack.hhomogeneous = false;
             action_stack.transition_type = Gtk.StackTransitionType.CROSSFADE;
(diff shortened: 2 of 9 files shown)


danrabbit (changes requested):
Hey thanks for working on this! With this branch I'm getting some really early wrapping for labels:

Screenshot from 2021-11-04 10 56 49

It looks like we need to take the hexpand off the stack to prevent that since it's claiming space that could be used by the label instead

danrabbit:
@isantop ffr, we use squash merge, so you don't need to rebase changes in pull requests. It actually makes it easier to review not to have to start fresh every time if we can just look at the commit history here between reviews

isantop:
> Hey thanks for working on this! With this branch I'm getting some really early wrapping for labels:
> 
> Screenshot from 2021-11-04 10 56 49
> 
> It looks like we need to take the hexpand off the stack to prevent that since it's claiming space that could be used by the label instead

Seems it needs to move to the package summary label, whoops! :sweat_smile: 

image

src/Widgets/AppContainers/ListPackageRowGrid.vala
```diff
@@ -28,8 +28,13 @@ public class AppCenter.Widgets.ListPackageRowGrid : AbstractPackageRowGrid {
     construct {
         package_summary = new Gtk.Label (null) {
             ellipsize = Pango.EllipsizeMode.END,
+            hexpand = true,
+            lines = 3,
```
danrabbit, line 32: Can we do 2 lines here instead of 3? that seems like just a ton of wrapping. We end up in some cases with only 1 or 2 words per line

src/Widgets/AppContainers/AbstractAppContainer.vala
```diff
@@ -111,7 +111,7 @@ namespace AppCenter {
             action_button_group.add_widget (open_button);
 
             action_stack = new Gtk.Stack ();
-            action_stack.hexpand = true;
+            action_stack.hexpand = false;
```
danrabbit, line 114: expand is false by default, so this line can be removed

src/Views/Homepage.vala
```diff
@@ -71,7 +71,7 @@ public class AppCenter.Homepage : AbstractView {
             row_spacing = 12,
             homogeneous = true,
             max_children_per_line = 5,
-            min_children_per_line = 3
+            min_children_per_line = 1
```
danrabbit, line 74: This is probably 1 by default so we probably don't need to specify this and can remove this line

danrabbit (changes requested):
Seems closer, but this totally breaks the updates view :sweat_smile: 

Screenshot from 2021-11-08 14 00 40

I wonder if we need to use Gtk.Box here so we can set pack_start, pack_end, and set_center_widget

src/Views/Homepage.vala
isantop: It sets to 0 by default; that seems to have the same behavior as setting to 1, but figured I'd highlight the difference in case it was relevant.

isantop:
> Seems closer, but this totally breaks the updates view sweat_smile
> 
> I wonder if we need to use Gtk.Box here so we can set pack_start, pack_end, and set_center_widget

Setting the `info_grid` hexpand prop (in `InstalledPackageRowGrid`) seems to fix this. I can look at at changing to Gtk.Box, if that would be preferable.

danrabbit:
@isantop I'm not too picky about it. Whatever ends up being easier :)

isantop:
Should be much better now.

danrabbit:
Updates view alignment is still pretty jacked

Screenshot from 2021-11-09 10 54 28

isantop:
Hmm, I think I know how that can be fixed. If not, I'll work on converting to a Gtk.Box

danrabbit:
Looks good, just one more minor thing! I think we need to add maybe a static horizontal size group for the description stack so that these items can be left aligned instead of of center aligned. Compare:

This branch:
Screenshot from 2021-11-10 14 20 39


main:
Screenshot from 2021-11-10 14 22 18

danrabbit:
Closing since we merged a competing branch

PR #1743, opened 2021-11-03
MainWindow: Allow search bar to shrink when window is narrow

isantop:
This reduces the required width for components in the headerbar, allowing the window to shrink to narrower sizes. Also changes the placeholder text when narrow to avoid ellipsizing the placeholder, which might reduce its usefulness.


danrabbit:
It doesn't seem like the search entry is really a limiting factor to window width here. Even when hiding release notes on the updates view, there's a lot of room in the headerbar:

Screenshot from 2021-11-12 09 50 16

PR #1745, opened 2021-11-03
AppInfoView: Make App Info header wrap at narrow screen sizes

isantop:
Reduces the minimum width for the window based on the width of the header items by allowing the action stack to wrap to be below the other app metadata.

Requires #1743


jeremypw:
Closing as dependency PR was closed and the header has been substantially redesigned with responsiveness in mind resulting in conflicts with this PR.

jeremypw:
Thanks for the contribution anyway!

PR #1747, opened 2021-11-07
Create CategoryView

### LINKED ISSUE #29: Grid view in AppCenter
I think that list view is not very convenient, because it is involves a lot of scrolling. By adding Grid view option in every category, AppCenter will have more compact look which make the search for new apps much easier and much more apps will be visible to the user at once, so more new apps will be discovered.
The grid will be resizable (In the sense of number of tiles per row) so every user will find the most convenient option for him.
Every tile in the grid will have the following information:
Icon
Name
Short summery about the app
Install\Uninstall option

The tile will behave as list item. Click will open the app dialog (Same as before), click on install or uninstall will bring authentication popup, after authentication, a progress bar will appear under the summery of the app and a description of the current step will be next to it.
The install\uninstall button will change to Cancel (Same as before).

Launchpad Details: [#LP1633400](https://bugs.launchpad.net/bugs/1633400) Ariel Bereslavsky - 2016-10-14 09:57:00 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43123490-grid-view-in-appcenter?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #29

Use flowboxes instead of a listbox to make better use of space

Screenshot from 2021-11-10 14 48 04

### DIFF
--- a/data/io.elementary.appcenter.appdata.xml.in
+++ b/data/io.elementary.appcenter.appdata.xml.in
@@ -11,6 +11,15 @@
     <p>The open source, pay-what-you-want app store from elementary. Reviewed and curated by elementary to ensure a native, privacy-respecting, and secure experience. Browse by categories or search and discover new apps. AppCenter is also used for updating your system to the latest and greatest version for new features and fixes.</p>
   </description>
   <releases>
+    <release version="3.9.0" date="2021-11-10" urgency="medium">
+      <description>
+        <p>Improvements:</p>
+        <ul>
+          <li>Categories show as a grid instead of a list</li>
+          <li>Updated translations</li>
+        </ul>
+      </description>
+    </release>
     <release version="3.8.2" date="2021-10-29" urgency="medium">
       <description>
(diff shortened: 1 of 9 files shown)


cassidyjames:
Mentioned in Slack but sharing here for the record and screenshots: it seems some categories (I would guess due to bad Flathub data) end up feeling _really_ spaced out with this flowbox. It's not worse than it being on one line, but it might be worth looking at and maybe truncating earlier:

Screenshot from 2021-11-15 13-53-16

Screenshot from 2021-11-15 13-53-05

cassidyjames (approved):
I like this a lot! Nice work.

Issue #1751, opened 2021-11-11
Back button requires double click


13r0ck:
### What Happened?

https://user-images.githubusercontent.com/58987761/141362243-8a965a5f-93b6-45c5-89e7-c3529cb2e56c.mp4

Related to #1715 & pop-os/shop#300

### Steps to Reproduce

1. Open a category
2. Search
3. Going back home requires clicking back twice

### Expected Behavior

After searching the back button should still go home

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

PR #1752, opened 2021-11-11
Make back button behaviour more consistant

### LINKED ISSUE #1750: Back button disappears after navigating back to search results
### What Happened?

It looks like the logic for hiding the back button is bad, as reported downstream: https://github.com/pop-os/shop/issues/300

### Steps to Reproduce

1. Search for something in AppCenter
2. Click a result
3. Go back to the results
4. Notice the back button is gone

### Expected Behavior

The back button should always be visible on the search results view

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

13r0ck:
Fixes #1750 and #1751


cassidyjames (changes requested):
@13r0ck thanks for submitting this! However, I'm still seeing some weird behavior here. When navigating from Home → Category (e.g. Accessories) → Search, the back button here still says "Home," and takes you home. However, I'd expect it to back out of the search and bring you back to the category.

If that's a much more invasive change I'd consider merging this as-is to fix the listed issues, but it still feels off, so I wonder if we can take the time to make it right.

13r0ck:
Ya it will take me a bit to get around to that, but I will make it happen. :smile:

13r0ck:
Nevermind I fixed it today

src/MainWindow.vala
```diff
@@ -104,6 +104,7 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         search_view.package_selected.connect (package_selected);
         search_view.subview_entered.connect (view_opened);
         search_view.home_return_clicked.connect (show_homepage);
+        search_view.category_return_clicked.connect(show_category);
```
cassidyjames, line 107: Lint: space before paren

```suggestion
        search_view.category_return_clicked.connect (show_category);
```

src/Views/SearchView.vala
```diff
@@ -51,17 +52,15 @@ public class AppCenter.Views.SearchView : AbstractView {
                 viewing_package = false;
 
                 if (current_category != null) {
-                    subview_entered (_("Search Apps"), true, current_category.name);
+                    subview_entered (_(current_category.name), true, current_category.name);
```
cassidyjames, line 55: If I'm reading this correctly, this should not be marked as translatable 

```suggestion
                    subview_entered (current_category.name, true, current_category.name);
```

src/Views/SearchView.vala
```diff
@@ -84,7 +83,7 @@ public class AppCenter.Views.SearchView : AbstractView {
         }
 
         if (current_category != null) {
-            subview_entered (_("Search Apps"), true, current_category.name);
+            subview_entered (_(current_category.name), true, current_category.name);
```
cassidyjames, line 86: Same as above

```suggestion
            subview_entered (current_category.name, true, current_category.name);
```

cassidyjames (changes requested):
Thanks again for this! I've taken an initial look and noticed some small fixes. I'll give it some testing but I think I'd like a more thorough review from @elementary/desktop-developers.

cassidyjames:
I can confirm that this seems to fix the back button labeling and behavior, though!

13r0ck:
Ah I didn't know that `_()` meant the string was translatable. Thank you for the review and teaching me!

zeebok (commented):
I think the code here looks good. I do wonder why some kind of stack isn't used to track previously visited views but that is something to explore another time. @cassidyjames if your tests are showing green I think this is good to go!

cassidyjames (approved):

PR #1753, opened 2021-11-11
Allow narrower window sizes

danrabbit:
Supersedes #1742

* MainWindow: change size request to default size
* AppInfoView, Homepage, CategoryFlowbox: Remove minimum flowbox child number
* AbstractAppContainer.vala: Remove actionstack hexpand to allow name and summary labels to fill width

* InstalledPackageRowGrid:
  * create infosizegroup here instead of passing it through multiple constructors
  * Add hexpand back to actionstack here since we need it to left align expanders which don't preserve width for hidden children

ListPackageRowGrid: expand labels and let them wrap


isantop (approved):
More or less what I had, with a couple of improvements and a cleaner history. Should work well to base the rest of the PRs on.

PR #1755, opened 2021-11-12
PackageRow: remove methods specific to updates view

danrabbit:
Turns out a bunch of this stuff is only used in the updates view so there's no real need to abstract it this much

Everywhere AppRowInterface is used, it needs to be a ListBoxRow. So there's really no point trying to separate PackageRow since that's the only implementation and it is a ListBoxRow

Apparently we also just didn't guard against package being null a lot?


jeremypw:
@danrabbit Please fix conflicts

jeremypw (approved):
Makes sense - no regressions found

Issue #1656, opened 2021-09-07
Crash when updating Flatpak


danirabbit:
## What Happened

When running updates, I get a crash from the flatpak backend

## Steps to Reproduce

1. Have some flatpak updates
2. Run Update All

### Logs

```
Starting program: /home/daniel/Projects/appcenter/build/src/io.elementary.appcenter 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff31d4700 (LWP 755101)]
[New Thread 0x7ffff29d3700 (LWP 755102)]
[New Thread 0x7ffff2158700 (LWP 755103)]
[New Thread 0x7ffff1957700 (LWP 755104)]
[New Thread 0x7ffff0bfc700 (LWP 755105)]
[New Thread 0x7fffe928c700 (LWP 755106)]
[New Thread 0x7fffe8a8b700 (LWP 755107)]
[Detaching after vfork from child process 755108]
[Thread 0x7ffff1957700 (LWP 755104) exited]
[Detaching after fork from child process 755528]

(io.elementary.appcenter:755097): Gtk-WARNING **: 11:22:20.139: Could not load a pixbuf from icon theme.
This may indicate that pixbuf loaders or the mime database could not be found.
[New Thread 0x7ffff1957700 (LWP 756492)]
[Thread 0x7ffff1957700 (LWP 756492) exited]
[New Thread 0x7ffff1957700 (LWP 756493)]
[Detaching after fork from child process 756494]
[Detaching after fork from child process 756496]
[Detaching after fork from child process 756498]
[New Thread 0x7fff971e3700 (LWP 756499)]
[Detaching after fork from child process 756501]
[Detaching after fork from child process 756503]
[Detaching after fork from child process 756505]
[Detaching after fork from child process 756507]
[New Thread 0x7fff969e2700 (LWP 756509)]
[Thread 0x7ffff1957700 (LWP 756493) exited]
[Thread 0x7fff971e3700 (LWP 756499) exited]
[Detaching after fork from child process 756523]
[Detaching after fork from child process 756527]
[Detaching after fork from child process 756533]
[Thread 0x7fffe928c700 (LWP 755106) exited]
[Detaching after fork from child process 756947]
[Detaching after fork from child process 756951]
[Detaching after fork from child process 756955]
[Detaching after fork from child process 756957]
[Thread 0x7fff969e2700 (LWP 756509) exited]
[...]

PR #1756, opened 2021-11-12
AbstractAppList: Remove unnecessary abstractions

danrabbit:
* `construct_row_for_package` is never called in the abstract class and it does a different one-liner in each implementation
* Replace `add_row` with connecting to `ListBox.add` so we make sure these things actually happen any time a widget is added to the list and not just when this special function is called
* `on_list_changed ()` only ever calls `invalidate_sort ()`, so just do that directly instead


jeremypw (approved):
Makes sense. I got a crash when updating with this PR but probably unrelated to this (see #1656)

PR #1763, opened 2021-11-15
CategoryFlowBox: create Fun & Games

### LINKED ISSUE #1761: Include "Amusement" in Games
### Problem

There are currently no games out of the box, but there are a couple of "Amusement" apps that currently only show in Accessories

### Proposal

Showing "Amusement" apps in Games would make it not totally empty and give these apps a place to go

### Prior Art (Optional)

_No response_

danrabbit:
Fixes #1761 

Screenshot from 2021-11-15 15 30 54

This ended up being a bigger branch because I wanted to do something more interest with the design here and the way we were creating the category items was kinda jank. This makes it easier for each category card to have a more unique design in its content area, while staying consistent with the card itself.

I only created a new card for Games here and made a `LegacyCard` class that maintains the behavior from CategoryItem

### DIFF
--- a/data/styles/categories.css
+++ b/data/styles/categories.css
@@ -273,24 +273,34 @@
 }
 
 .category.games {
-    background-image:
-        linear-gradient(
-            to bottom,
-            mix(@SLATE_500, @BLUEBERRY_500, 0.5),
-            @SLATE_900
-        );
-    border-color: alpha(@SLATE_900, 0.8);
+    border: none;
     box-shadow:
-        inset 0 0 0 1px alpha(white, 0.02),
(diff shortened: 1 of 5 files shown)


jeremypw:
@danrabbit Please fix conflict.

tintou:
I'm trying to translate it in French in my head and `Fun & Games` would be translated as `Jeux & Divertissements` but that makes me wonder what would distinguish `Fun` from `Games`?

danrabbit:
@tintou theres a category "amusements" that is included here that aren't quit games but more like toys? Like a magic 8 ball app. Not really quite a game, but its for play

tintou (approved):

PR #1765, opened 2021-11-16
CategoryFlowBox: Add Privacy & Security

### LINKED ISSUE #1760: Add Security Category
### Problem

There are a small handful of security focused apps in AppCenter now, but they're currently only showing in "Accessories"

### Proposal

Maybe we should add a "Security" category to show these off

### Prior Art (Optional)

_No response_

danrabbit:
Does the least to fix #1760 

Screenshot from 2021-11-16 14 37 45
Screenshot from 2021-11-16 14 37 25

I didn't see anything else here that really struck me as fitting in: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

I also have PRs against [Oblivate](https://github.com/elfenware/obliviate/pull/21/files) and [Hasher](https://github.com/JeysonFlores/hasher/pull/23/files) so there would be a couple more apps here

### DIFF
--- a/src/Widgets/CategoryFlowBox.vala
+++ b/src/Widgets/CategoryFlowBox.vala
@@ -133,6 +133,9 @@ public class AppCenter.Widgets.CategoryFlowBox : Gtk.FlowBox {
             "Translation",
             "WordProcessor"
         }, "writing-language"));
+        add (get_category (_("Privacy & Security"), "preferences-system-privacy", {
+            "Security",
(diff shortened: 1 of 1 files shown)


cassidyjames:
Yeah I don't see any other main or additional categories that fit. I'm fine with this and then we can follow up with a nicer design when we pull it into its own class.

cassidyjames (approved):
👍

PR #1768, opened 2021-11-17
Core: Reduce the memory footprint and avoid unnecessary reference increase

tintou:
Also always allow null Cancellable.
Also avoid to recreate the same strings in the loops.


vjr:
Hello... @tintou  do you have any data/measure of how much memory footprint improvement is achieved by this branch?

tintou:
@vjr I haven't but if you look at the code, you'll see that I'm reducing the number of string copies here and there (the most important is in loops where we were recreating the same string over and over)

davidmhewitt (changes requested):
These look like really good changes! I'll test it out later.

It looks like you need to mark a few more cancellable objects as nullable in `Job.vala` as well. I only did the ones where the associated method had a nullable cancellable parameter.

tintou:
@davidmhewitt Indeed, done, also set the Error to be nullable.

davidmhewitt (approved):
Sorry it took so long to come back to this. I've tested this and can't find any regressions, and the code looks good, so lets go!

PR #1769, opened 2021-11-18
AbstractView: Replace Stack with Deck

### LINKED ISSUE #1141: Add Navigation Gestures
We should have swiping navigation gestures, likely using Handy.Deck

### LINKED ISSUE #1517: Sliding animations not correct
## What Happened

In category view, I can click on apps to view its information. However, sometimes the it slides from left to right instead of right to left. This issue is observed in 'maths, science and engineering' only at the moment.

https://user-images.githubusercontent.com/39268094/114126586-4a0de480-992b-11eb-80b8-2d0d8eb98b89.mp4

## Expected Behavior

The sliding animation should be from right to left

https://user-images.githubusercontent.com/39268094/114126695-893c3580-992b-11eb-8b7f-99e05d1206d5.mp4

## Steps to Reproduce

1. Open AppCenter
2. Click on the 'math, science and engineering' category
3. Try clicking on apps

## Platform Information

Screenshot from 2021-04-08 13-53-56

### LINKED ISSUE #1628: Could not return to home page from app page
I went to an app page, Mixer, and got a notification to install updates. I click on it, was sent to the updates tab, installed the updates, went back to the home tab, only to be stuck on the app page from before and couldn't go back home. 
Screenshot from 2021-08-25 23 33 28

### LINKED ISSUE #1751: Back button requires double click

danrabbit:
Gotta watch out for https://github.com/elementary/appcenter/issues/1482

fixes #1141
fixes #1517 
fixes #1751
fixes #1628

* Replace Stack with Hdy.Deck and use the deck itself for navigation history
* Remove SearchView and make it a subview of HomePage like CategoryView. Since you can’t search the installed page this makes navigation way simpler

Cases to test:
* Navigating back from AppInfoViews on HomePage
* Navigating back from AppsInfoViews on InstalledView
* Navigating back from CategoryViews
* Navigating back from Search
* Search from homepage
* Search from category view
* Search is disabled anywhere from installed view
* Launching a search from the Applications Menu
* Navigating back from external search/launch
* Navigating back from external AppInfo/launch


danrabbit:
I think this should all work now. Definitely needs thorough testing since it's pretty invasive

jeremypw (changes requested):
Navigation seems to work as expected. However, there are some issues with search that are not present in master:

* Searching from HomePage for the term "file" results in multiple entries (up to 3) for some apps
(Searching from Slingshot for "file" does not produce multiple entries)
* If a search is made while the cache update is happening (e.g. immediately after opening) and the search view is left open until an "Update found" notification occurs, then the view switches to "No Apps found".

danrabbit:
@jeremypw fixed the issue with search views being cleared. Looks like some old code that was meant for when category views were lists was have unintended effects here.

I can't reproduce the issue with duplicate entries when searching for "file" can you post a screenshot?

jeremypw:
Multiple search results
Screenshot from 2022-05-05 18 22 49

jeremypw:
I had more difficulty triggering the multiple item bug this time but managed it with a different search term.  I think it is more to do with the way the search term is entered/edited than the actual term.  Perhaps multiple similar searches are contributing to the same results list?

jeremypw:
It seems to be causes by adding/subtracring to the initial search term and (maybe) scrolling down while search is ongoing.  I have obtained upto 6x duplication of some items this way.

danrabbit:
@jeremypw ah thank you! I was able to reproduce and fix the issue

jeremypw (changes requested):
I found another bug with this PR, not sure if it exists in master.  If you double click on a category item on the Home page, then two category pages are produced. If you double click on an app item on the Home page or on a category page then two app pages are produced.  As it can be slow for a category page to display (with no feedback) multiple clicks are quite likely to occur.

Edit: This does not occur in master.

danrabbit:
Ah yeah I see that. I'm not sure exactly how to prevent multiple clicks from being parsed here, but I suppose one thing that needs to be done regardless is make loading these views much faster.

This branch doesn't improve that too much yet, but first steps: https://github.com/elementary/appcenter/pull/1851

danrabbit:
@jeremypw okay so after the async loading branch is merged, it's quite difficult to double click here. You can do it if you really try, but it's probably not something that would happen in normal usage?

jeremypw:
It is still quite easy for me to get duplicate views in the deck. It does not actually require a double-click - two single clicks does it provided you click on e.g. the banner again before the new view slides into place. As I am, atm, running from a relatively slow spinning disk (having bricked my SSD) there is an appreciable delay before the new view appears.  I'll see whether there is a simple way of preventing this issue.

danrabbit:
Ah interesting. Yeah for me it's like maybe 100-150ms. You have to really be trying to double click as fast as you can 😅

danrabbit:
@jeremypw I'm happy with that :)

jeremypw:
It mainly affects the banner, where the transition takes 200ms (+) to complete so there is a fair amount of time to click again on the left hand edge of the banner. There is an even longer delay if you click on the banner immediately after launching AppCenter for the first time.  Anyway, I am happy to approve this now.

jeremypw (approved):

danrabbit:
Thanks for testing thoroughly and working with me on this!!

PR #1770, opened 2021-11-18
Homepage: get currently_viewed_category accurately

danrabbit:
Seems kinda weird to try to set this value manually when we can reliably get the actual value on demand


jeremypw (approved):
Makes sense, no regressions found.

PR #1778, opened 2021-11-30
Avoid multiple circular references

tintou:
The use of notify on external objects and of timeout actually increases the reference of the object, try to release it on-demand.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -15,15 +15,7 @@
 */
 
 public class AppCenter.MainWindow : Hdy.ApplicationWindow {
-    public bool working {
-        set {
-            if (value) {
-                spinner.start ();
-            } else {
-                spinner.stop ();
-            }
-        }
-    }
+    public bool working { get; set; }
 
     private Gtk.Revealer view_mode_revealer;
     private Gtk.Stack custom_title_stack;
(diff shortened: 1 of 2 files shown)


src/MainWindow.vala
```diff
@@ -114,10 +105,12 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         search_view.subview_entered.connect (view_opened);
         search_view.home_return_clicked.connect (show_homepage);
 
-        unowned AppCenterCore.BackendAggregator client = AppCenterCore.BackendAggregator.get_default ();
-        client.notify["working"].connect (() => {
+        unowned AppCenterCore.BackendAggregator aggregator = AppCenterCore.BackendAggregator.get_default ();
```
jeremypw, line 108: Can't we use `unowned var` here?

jeremypw:
The code makes sense and uninstalling/ installing works as before.  However, even without this PR, I am experiencing crashes when running "Update All" including "System Updates" and I haven't got to the bottom of that yet. A crash also occurred when doing the same thing with this PR.  Not sure I can properly test this when my existing AppCenter installation is not stable :disappointed:

jeremypw (approved):
LGTM. No regressions found.

PR #1783, opened 2021-12-03
Core: Pass the ChangeInformation object around instead of its callback

tintou:
Allows to make sure that we are safe regarding the lifetime of the ChangeInformation
and allows to pass it safely to multiple closures.


danrabbit:
Not sure if this is related to this branch since I don't see a hashmap added here, but I just ran an "Update All" wish this branch and got a crash with:

```
ERROR:hashmap.c:3710:gee_hash_map_node_iterator_next: assertion failed: (_stamp == _map._stamp)
Bail out! ERROR:hashmap.c:3710:gee_hash_map_node_iterator_next: assertion failed: (_stamp == _map._stamp)
Aborted
```

jeremypw (approved):
I confirm that this PR allowed an "Update All" process to complete without crashing for me.

Issue #1776, opened 2021-11-24
Pull apt updates less frequently?


danrabbit:
### Problem

Fetching updates from apt is slow and takes a long time. Most of the time when we're opening AppCenter, we really only want to see new apps, which are all from Flatpak

### Proposal

Perhaps we should make sure to only fetch Flatpak updates when we open AppCenter and keep apt updates just to the daily timer

cassidyjames:
This makes sense to me. Only caveat I'd have is if we'd want to add a manual refresh button to trigger an apt update, e.g. in the case of a critical security update (it would be hard to tell users, "Just wait for tomorrow and then it'll prompt you" when the status quo is "just open AppCenter and update.").

AI-ien:
Isn't that the task for Appcenter daemon to fetch updates in background so when you open Appcenter you don't need to wait for updates?

Another option would be to make slight design changes by adding third tab 'updates' where on select updates are rechecked.

SubhadeepJasu:
> This makes sense to me. Only caveat I'd have is if we'd want to add a manual refresh button to trigger an apt update, e.g. in the case of a critical security update (it would be hard to tell users, "Just wait for tomorrow and then it'll prompt you" when the status quo is "just open AppCenter and update.").

I am cool with that. A manual refresh button or change the update button to "Check for updates" when the update information is outdated.

PR #1791, opened 2021-12-12
Add a manual refresh button to update all caches

meisenzahl:
Accompanies #1790 and implements @cassidyjames' suggestion from #1776.

https://user-images.githubusercontent.com/10796736/145714034-23b95f9b-2eb9-479e-8a87-9b02916e6a14.mp4


cassidyjames:
Another UX/security question I have: can we restrict automatic updates to curated apps? Automatically updating non-curated apps means apps could update with more sandbox holes that haven't been human-reviewed, which sounds like a bad idea. Otherwise I think we'd need to implement some sort off manual approval process if the Flatpak sandbox changes, which sounds even more complicated.

hanaral:
This should probably be a little more discreet, as having the button on the main page and binding it to ctrl+R gives it a browser like feel, which is really not a good idea when refreshing can potentially lock up the appcenter on slower systems.
How about showing a banner (on the 'Updates' tab) asking if the user wants to manual refresh if an automatic apt one hasn't occurred in x amount of time?

meisenzahl:
@elementary/ux do you have an opinion on @hanaral's comments?

cassidyjames:
@meisenzahl I think adding it to the menu from #1793 makes the most sense, actually.

meisenzahl:
https://user-images.githubusercontent.com/10796736/148538211-0867c664-8bdd-4482-9e93-8b2d27fc53fa.mp4

src/MainWindow.vala
```diff
@@ -227,6 +229,17 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
             }
         });
 
+        var refresh_accellabel = new Granite.AccelLabel.from_action_name (
+            _("Refresh"),
```
cassidyjames, line 233: It might not be clear what this is refreshing, since someone might think it's reloading the current view? So slightly more explicit is probably better:

```suggestion
            _("Check for Updates"),
```

src/MainWindow.vala
```diff
@@ -112,6 +113,7 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         notify["working"].connect (() => {
             Idle.add (() => {
                 spinner.active = working;
+                refresh_menuitem.sensitive = !working;
```
danrabbit, line 116: Instead of setting the menuitem sensitive, this should set the action as disabled so that we also disabled the keyboard shortcut

danrabbit (changes requested):

danrabbit (approved):

Issue #1471, opened 2021-01-04
Make loading updates async from main AppCenter window


hanaral:
## What Happened
Trying to do a search for example, will make the AppCenter lock up for a while if the AppCenter is loading a large amount of available upgrades in the background

## Expected Behavior
The updating process is seperate from the fronted so that the user does not face lockups just trying use the app normally

## Steps to Reproduce

1. Open AppCenter and immediately try to search
2. It should chug (or even lock up in the case of my install HDD on my desktop)

This seems to be less obvious with an SSD, but you can still see it freeze for a second even with a small number of available upgrades

## Platform Information
elementary OS 6 Daily
Tested on Lenovo ThinkPad X240 and desktop with AMD Ryzen 3 1200 & RX 570

ids1024:
I don't know if there's any good tooling to check what is blocking an application's main thread, but running in gdb and hitting ctrl+c when the UI is blocked is one way to get a trace of what seems to be blocking.

One thing I'm seeing is the call to `Flatpak.Transaction.run` in `FlatpakBackend.get_download_size_by_id`. That function is `async` but the documentation for `flatpak_transaction_run` [explicitly states](https://docs.flatpak.org/en/latest/libflatpak-api-reference.html#flatpak-transaction-run) "Note that this call blocks until the transaction is done.".

I think there are various other issues like that (in both the flatpak and packagekit backends). Given there doesn't seem to by an asynchronous API for this, presumably all such calls into `libflatpak` should be in a separate thread or process.

davidmhewitt:
@ids1024 Most of the long running operations in the Backend classes are already run in a separate thread. There's a thread pool and a pattern of a public async method queueing up work via a private internal worker method on that pool.

You'll definitely see that for the install/update/remove methods. It's possible that this specific one was just missed for some reason.

ids1024:
Yes, most of those operations happen in a background thread, but this issue seems to exist due to a few that do not.

It seems like there should be a good way to profile what functions are blocking the UI thread. That would help for getting a more complete picture of what issues there are here.

davidmhewitt:
I believe sysprof's "speedtrack" should be able to identify blocking I/O happening on the main thread.

aral:
I’ve noticed that checking for app updates blocks the main thread rather considerably (to the point where I thought the app had hung the first few times it happens). I can reproduce every time I launch AppCenter.

danieleiervo:
A full year is gone and I can still reproduce: are there any updates on this?

PR #1793, opened 2021-12-12
Automatically install curated Flatpak updates

### LINKED ISSUE #1446: Automatically install Flatpak updates
See #288 for some context, but we should consider automatically installing Flatpak updates on non-metered networks.

UX to be determined, i.e. whether or not there is a preference for it, and if so, where that would be.

meisenzahl:
Closes #1446

Bildschirmfoto von 2021-12-12 18 04 51


meisenzahl:
I tested this in a VM and unfortunately found that the `update_available` property is not `true` for packages that have updates available. I'll have another look at this.

meisenzahl:
Seems to work now:

https://user-images.githubusercontent.com/10796736/145728054-a9a96aa8-fa1a-4d57-81e1-a1340149afdc.mp4

meisenzahl:
AppCenter is already running in the background as `io.elementary.appcenter -s`. If the package caches update is older than 24 hours, the Flatpak updates would be installed automatically in the background.

I am open to feedback from @elementary/ux on whether the `automatically-install-flatpak-updates` setting should be enabled by default.

hanaral:
@meisenzahl I mean that appcenter should manage updates along with installs as a seperate process (not subprocess) as proposed in #1471 (if anyone is willing to implement it)
Good to see a preexisting gsetting for metred connections, it would be even better if there was a allowlist gsetting for networks so it never does autoupdates besides user-set networks

cassidyjames:
UX wise I agree this should be off by default until we can handle metered connections. At that point I think it would be safe to turn it on by default and/or prompt users to enable it.

If we can distinguish between "automatic" and explicitly-set non-metered (e.g. with `NM_METERED_NO`) connections, we could choose to do that, but I don't know if that's necessary.

Screenshot from 2021-12-13 11 02 01

cassidyjames:
Another UX/security question I have: can we restrict automatic updates to curated apps? Automatically updating non-curated apps means apps could update with more sandbox holes that haven't been human-reviewed, which sounds like a bad idea. Otherwise I think we'd need to implement some sort off manual approval process if the Flatpak sandbox changes, which sounds even more complicated.

data/io.elementary.appcenter.gschema.xml
```diff
@@ -41,5 +41,10 @@
       <summary>Unix UTC time of last cache refresh</summary>
       <description>Used to determine when AppCenter last refreshed its caches and checked for package updates</description>
     </key>
+    <key type="b" name="automatically-install-flatpak-updates">
+      <default>false</default>
+      <summary>Automatically install Flatpak updates</summary>
+      <description>Whether to automatically install Flatpak updates</description>
```
cassidyjames, line 47:
```suggestion
      <description>Whether to automatically install updates to curated Flatpaks</description>
```

meisenzahl:
@cassidyjames yes we could query if the connection was automatically or explicitly configured as metered.

The last push restricts automatic updates to curated apps.

src/MainWindow.vala
```diff
@@ -214,11 +214,40 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
         spinner = new Gtk.Spinner ();
 
+        var automatically_install_flatpak_updates_button = new Granite.SwitchModelButton (_("Automatic Updates")) {
+            description = _("Automatically install updates to curated apps")
+        };
+
+        var menu_popover_grid = new Gtk.Grid () {
+            column_spacing = 6,
+            margin_bottom = 6,
+            margin_top = 12,
```
cassidyjames, line 224: Should be even margins here

```suggestion
            margin_top = 6,
```

src/MainWindow.vala
```diff
@@ -214,11 +214,40 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
+        menu_popover_grid.add (automatically_install_flatpak_updates_button);
+
+        menu_popover_grid.show_all ();
+
+        var menu_popover = new Gtk.Popover (null);
+        menu_popover.add (menu_popover_grid);
+
+        var menu_button = new Gtk.MenuButton () {
+            can_focus = false,
+            image = new Gtk.Image.from_icon_name ("open-menu", Gtk.IconSize.SMALL_TOOLBAR),
```
cassidyjames, line 238: This should be the large toolbar size, but I know that's gonna require some styling fixes as well since we do weird things with margins/padding in CSS here (I think due to the updates badge overlay). I may have started addressing it in #1230 but I also know that PR is pretty outdated.

mcclurgm:
If I may add a UX thought, it may make sense to place the switch at the top of the "Installed" tab, instead of being the only entry under a gear menu.

Since it's about the updates that appear in that tab, to me it's a natural fit there, and not in the home tab where it's also currently visible. That would also make it more visible, so it would kind of implicitly prompt people to enable the setting without adding an explicit dialog. 
Having the switch always present may be unnecessary clutter, though.

src/MainWindow.vala
meisenzahl: Any suggestions on how we are going to implement this at this point?

cassidyjames:
@mcclurgm that might be fair considering it's the only item in that menu, but we've been talking about adding a menu for some time, e.g. for the items in #1230 or even for the refresh/fetch action from #1791. I would be okay moving to a menu in anticipation of those actions.

src/MainWindow.vala
cassidyjames: @meisenzahl actually in testing, using `LARGE_TOOLBAR` looks really close; we might just have to tweak the styling here to match other apps; it looks like 3px matches Files, at least, but I'm also not sure how we would get it to scale with the text-size like [our `rem()` sass function](https://github.com/elementary/stylesheet/blob/9b7122b07b3a1878501d13d842c54e83ef9061c6/src/gtk-3.0/_index.scss#L162-L178) in the stylesheet… :thinking: That might be over-thinking it, though, and 3px here might be fine.

https://github.com/elementary/appcenter/blob/c9d9f16e4617160d511ba099b0fd8aa29dbea98e/data/styles/application.css#L18-L21

meisenzahl: Using `LARGE_TOOLBAR` and 0.188rem for padding:

Bildschirmfoto von 2021-12-13 20 35 43

data/styles/application.css
```diff
@@ -16,6 +16,6 @@
  */
 
 .titlebar {
-    padding-bottom: 0;
-    padding-top: 0;
+    padding-bottom: 0.188rem;
+    padding-top: 0.188rem;
```
cassidyjames, line 20: Oh, duh! Smart. In this case I think we should leave a comment about where this is coming from for future spelunkers, though. Like "3px at 9pt font, see: https://github.com/elementary/stylesheet/blob/9b7122b07b3a1878501d13d842c54e83ef9061c6/src/gtk-3.0/_index.scss#L162" or something.

meisenzahl:
The last commit filters all automatically installed Flatpak apps in `UpdateManager` so they are not shown in Installed tab. Next I will investigate how to show them but not count them in the notification.

## Before

image

## After

image

zeebok:
> Sorry for all the random review comments instead of aggregating into one review! I think this is close, but I think one problem to solve UX-wise is that we should avoid throwing the notification for updates that are going to be automatically installed, as there's not really a user action to be performed here (besides waiting for them to be installed).

Another to consider is the expected behavior when a user opens the app with available updates already listed and then turns on the setting. I would think it should automatically trigger a refresh on the list and run updates on curated flatpak apps.

Otherwise the code that is here seems good and Cassidy already commented on language so I don't need to ask about that.

meisenzahl:
@zeebok thanks for the suggestion! I have implemented that when changing the setting the updates are installed, or canceled.

zeebok (approved):

data/io.elementary.appcenter.gschema.xml
```diff
@@ -41,5 +41,10 @@
       <summary>Unix UTC time of last cache refresh</summary>
       <description>Used to determine when AppCenter last refreshed its caches and checked for package updates</description>
     </key>
+    <key type="b" name="automatically-install-flatpak-updates">
+      <default>false</default>
+      <summary>Automatically install Flatpak updates</summary>
+      <description>Whether to automatically install updates to curated Flatpaks</description>
+    </key>
```
cassidyjames, line 48: Series of nits. I know your name is more precise, but it's pretty wordy, when a more concise key would do—especially as I don't see us offering multiple different "automatic updates" mechanisms inside of AppCenter in the future. This would also more closely match the copy used in-app, which is convenient.

```suggestion
    <key type="b" name="automatic-updates">
      <default>false</default>
      <summary>Automatically install updates</summary>
      <description>Whether to automatically install updates to curated Flatpak apps</description>
    </key>
```

src/Core/Client.vala
```diff
@@ -168,6 +165,26 @@ public class AppCenterCore.Client : Object {
 
             return GLib.Source.REMOVE;
         });
+
+        if (nm.get_network_available ()) {
+            if (last_cache_update_is_old && AppCenter.App.settings.get_boolean ("automatically-install-flatpak-updates")) {
```
cassidyjames, line 170: Matching above suggestion

```suggestion
            if (last_cache_update_is_old && AppCenter.App.settings.get_boolean ("automatic-updates")) {
```

src/Core/UpdateManager.vala
```diff
@@ -100,6 +101,11 @@ public class AppCenterCore.UpdateManager : Object {
             if (appcenter_package != null) {
                 debug ("Added %s to app updates", flatpak_update);
                 apps_with_updates.add (appcenter_package);
+
+                if (!(AppCenter.App.settings.get_boolean ("automatically-install-flatpak-updates") && appcenter_package.is_native)) {
```
cassidyjames, line 105:
```suggestion
                if (!(AppCenter.App.settings.get_boolean ("automatic-updates") && appcenter_package.is_native)) {
```

src/Core/UpdateManager.vala
```diff
@@ -122,7 +128,10 @@ public class AppCenterCore.UpdateManager : Object {
                     continue;
                 }
 
-                os_count++;
+                if (!AppCenter.App.settings.get_boolean ("automatically-install-flatpak-updates")) {
```
cassidyjames, line 131:
```suggestion
                if (!AppCenter.App.settings.get_boolean ("automatic-updates")) {
```

src/MainWindow.vala
```diff
@@ -214,11 +214,48 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
 
         spinner = new Gtk.Spinner ();
 
+        var automatically_install_flatpak_updates_button = new Granite.SwitchModelButton (_("Automatic Updates")) {
```
cassidyjames, line 217: Similar to the other suggestion; this is super wordy :)

```suggestion
        var automatic_updates_button = new Granite.SwitchModelButton (_("Automatic Updates")) {
```

src/MainWindow.vala
```diff
@@ -214,11 +214,48 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
+            description = _("Automatically install updates to curated apps")
+        };
+
+        automatically_install_flatpak_updates_button.notify["active"].connect (() => {
+            if (automatically_install_flatpak_updates_button.active) {
```
cassidyjames, line 222:
```suggestion
        automatic_updates_button.notify["active"].connect (() => {
            if (automatic_updates_button.active) {
```

src/MainWindow.vala
```diff
@@ -214,11 +214,48 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
+
+        var menu_popover_grid = new Gtk.Grid () {
+            column_spacing = 6,
+            margin_bottom = 6,
+            margin_top = 6,
+            orientation = Gtk.Orientation.VERTICAL,
+            row_spacing = 6
+        };
+
+        menu_popover_grid.add (automatically_install_flatpak_updates_button);
```
cassidyjames, line 237:
```suggestion
        menu_popover_grid.add (automatic_updates_button);
```

src/MainWindow.vala
```diff
@@ -252,6 +289,12 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         int window_width, window_height;
         App.settings.get ("window-position", "(ii)", out window_x, out window_y);
         App.settings.get ("window-size", "(ii)", out window_width, out window_height);
+        App.settings.bind (
+            "automatically-install-flatpak-updates",
+            automatically_install_flatpak_updates_button,
```
cassidyjames, line 294:
```suggestion
            "automatic-updates",
            automatic_updates_button,
```

cassidyjames (changes requested):
Some naming nits, but this looks pretty god to me. My only other question is: since we're reading the GSetting multiple times in Client, does it make sense to read it once and cache its value, or is it fine to read it multiple times?

cassidyjames (changes requested):
I can't confirm this is actually automatically installing curated Flatpak updates. I toggled the setting on, then manually downgraded Dippi (e.g. with `flatpak update --commit=b826d7d99eb010f63c60405404363386e66ffcf83b1ce55f0e3fbed8634589e4 com.github.cassidyjames.dippi`). AppCenter still shows the update as available, but as far as I can tell, won't install it automatically.

cassidyjames:
Ah @meisenzahl I see now that this is using the last refresh time to avoid auto updating all the time. That's smart, and by resetting the refresh and restarting AppCenter, I can confirm the update was automatically installed.

cassidyjames:
Alright, I think I just have one last @elementary/ux question: how do we treat paid apps that were installed for free? Traditionally, we treated these a bit differently to re-prompt the user to pay. Should these be excluded from automatic updates like they're excluded from the "Update All" logic? My inclination is yes, but that makes it harder to explain in the app.

src/MainWindow.vala
```diff
@@ -214,11 +214,48 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
 
         spinner = new Gtk.Spinner ();
 
+        var automatic_updates_button = new Granite.SwitchModelButton (_("Automatic Updates")) {
+            description = _("Automatically install updates to curated apps")
```
cassidyjames, line 218: If we exclude non-paid-for paid app, this might work:

```suggestion
            description = _("Automatically update free and paid-for curated apps in the background")
```

src/Core/Client.vala
```diff
@@ -168,6 +165,26 @@ public class AppCenterCore.Client : Object {
...
             return GLib.Source.REMOVE;
         });
+
+        if (nm.get_network_available ()) {
+            if (last_cache_update_is_old && AppCenter.App.settings.get_boolean ("automatic-updates")) {
+                yield refresh_updates ();
+                debug ("Update Flatpaks");
+                var installed_apps = yield FlatpakBackend.get_default ().get_installed_applications (cancellable);
+                foreach (var app in installed_apps) {
+                    if (app.is_native && app.update_available) {
```
cassidyjames, line 175: I think this should work excluding paid apps that were installed for free:

```suggestion
                    if (app.is_native && app.update_available && !app.should_pay) {
```

cassidyjames (approved):

PR #1801, opened 2022-01-03
Expose screenshot captions

### LINKED ISSUE #1027: Expose screenshot captions
Screenshots can optionally have a translatable caption. We should expose this somewhere, perhaps below the screenshot or as a tooltip.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

Suggested at https://twitter.com/phase1g_o/status/1146395526319005697

Antolius:
closes #1027 

Feature can be tested with either of Internet → Starfish or Games → Sage apps.

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -833,7 +833,7 @@ namespace AppCenter.Views {
                     return null;
                 }
 
-                List<string> urls = new List<string> ();
+                List<CaptionedUrl> captioned_urls = new List<CaptionedUrl> ();
 
                 var scale = get_scale_factor ();
                 var min_screenshot_width = MAX_WIDTH * scale;
@@ -857,20 +857,25 @@ namespace AppCenter.Views {
                         }
                     });
 
+                    var captioned_url = new CaptionedUrl (
+                        screenshot.get_caption (),
+                        best_image.get_url ()
+                    );
+
                     if (screenshot.get_kind () == AppStream.ScreenshotKind.DEFAULT && best_image != null) {
-                        urls.prepend (best_image.get_url ());
+                        captioned_urls.prepend (captioned_url);
                     } else if (best_image != null) {
-                        urls.append (best_image.get_url ());
+                        captioned_urls.append (captioned_url);
                     }
                 });
 
-                string?[] screenshot_files = new string?[urls.length ()];
-                bool[] results = new bool[urls.length ()];
+                string?[] screenshot_files = new string?[captioned_urls.length ()];
+                bool[] results = new bool[captioned_urls.length ()];
                 int completed = 0;
 
                 // Fetch each screenshot in parallel.
-                for (int i = 0; i < urls.length (); i++) {
-                    string url = urls.nth_data (i);
+                for (int i = 0; i < captioned_urls.length (); i++) {
+                    string url = captioned_urls.nth_data (i).url;
                     string? file = null;
                     int index = i;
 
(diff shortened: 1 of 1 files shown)


Antolius:
With a decision in  #1027 to go with tooltips for exposing captions I think this pull request is ready for review.


Caption shown over a "busy" app screenshot.
busy-screenshot


Caption shown over a "sparse" app screenshot.
sparse-screenshot


Overly long caption is [shortened](https://docs.elementary.io/hig/reference/text#shortened-text).
long-caption


According to [Appstream Metadata specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots) captions should be:

> ... a short (ideally not more than 100 characters) description of what the user can see...

We can check caption length for curated apps during their review. But since captions will be shown for non-curated apps too I implemented a limit of 500 characters (5 times the "ideal maximum"). If you'd like I can take it down to some smaller number or enlarge it to allow for even longer captions. 500 seemed like a good compromise between suggested "ideal max" and what looked good to me, but in the end it is an arbitrary magic number.

src/Views/AppInfoView.vala
```diff
@@ -929,6 +935,11 @@ namespace AppCenter.Views {
                 image.icon_name = "image-x-generic";
                 image.halign = Gtk.Align.CENTER;
                 image.gicon = pixbuf;
+                if (caption != null) {
+                    // AppStream spec says "ideally not more than 100 characters"
+                    int max_caption_len = 500;
+                    image.tooltip_text = ellipsize (caption, max_caption_len);
+                }
```
cassidyjames, line 942: Actually, does it make sense use GTK's [custom tooltip](https://valadoc.org/gtk+-3.0/Gtk.Tooltip.set_custom.html) functionality to add a label, and use GTK and Pango's [built-in ellipsization](https://valadoc.org/gtk+-3.0/Gtk.Label.ellipsize.html) here?

```suggestion

                if (caption != null) {
                    image.has_tooltip = true;
                    image.query_tooltip.connect ((x, y, keyboard_tooltip, tooltip) => {
                        tooltip.set_custom (new Gtk.Label (caption) {
                            ellipsize = Pango.EllipsizeMode.END,
                            // AppStream spec says "ideally not more than 100 characters"
                            max_width_chars = 200
                        });

                        return true;
                    });
                }
```

src/Views/AppInfoView.vala
```diff
@@ -940,6 +951,17 @@ namespace AppCenter.Views {
...
+            if (long_text.length > max_length) {
+                StringBuilder sb = new StringBuilder (long_text);
+                sb.truncate (max_length);
+                sb.append ("\u2026");
+                return sb.str;
+            }
+
+            return long_text;
+        }
+
```
cassidyjames, line 964: If using the GTK label's ellipsize, we don't need this.

```suggestion
```

cassidyjames (changes requested):
I could be wrong, but I think it makes more sense to use GTK's built-in ellipsization than handling that ourselves.

src/Views/AppInfoView.vala
Antolius: I tried this out, but I think Label applies ellipsis to each line independently, i.e. it will cut off each line at 200 chars. On the one hand that is a bit long for a single line. On the other hand it doesn't prevent entire caption from being longer than 200 chars (if it has multiple lines). This is what it looks like to use the Label:

label-ellipsize

I tried applies line wrapping on top of it, but it doesn't work in combination with ellipsizing. In the end I think manually cutting the caption at 200 chars total and using the default tooltip implementation (which limits the width by wrapping) gives the nicest looking result:

tooltip-200

src/Views/AppInfoView.vala
```diff
@@ -929,6 +935,11 @@ namespace AppCenter.Views {
                 image.icon_name = "image-x-generic";
                 image.halign = Gtk.Align.CENTER;
                 image.gicon = pixbuf;
+                if (caption != null) {
+                    // AppStream spec says "ideally not more than 100 characters"
+                    int max_caption_len = 500;
```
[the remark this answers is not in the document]
Antolius: Max length of 200 looks good!

cassidyjames (approved):
Thank you for your work on this!

Antolius:
No problem! And thank you for engaging with the PR so quickly and for guiding me along the way!

PR #1804, opened 2022-01-07
Warn about apps built against EOL runtimes

### LINKED ISSUE #1803: Warn about apps built against old runtimes
### Problem

Apps built against older runtimes may not be 100% API compatible or take advantage of new system features

### Proposal

Similar to the non-curated warning, we should let users know if an app was built against an elementary runtime version that is older than the OS version

### Prior Art (Optional)

_No response_

meisenzahl:
Re: #1803

https://user-images.githubusercontent.com/10796736/148535232-e277f6a1-5f29-46e0-9f99-b24bd8d9e080.mp4

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -498,7 +498,6 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
 
         var added_remotes = new Gee.ArrayList<string> ();
 
-        bool runtime_eol = false;
         try {
             var transaction = new Flatpak.Transaction.for_installation (installation, cancellable);
             transaction.add_default_dependency_sources ();
@@ -537,7 +536,9 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
                         var remote_ref = installation.fetch_remote_ref_sync (remote_name, kind, name, arch, branch, cancellable);
 
                         if (remote_ref.get_eol () != null || remote_ref.get_eol_rebase () != null) {
-                            runtime_eol = true;
+                            if (package != null) {
+                                package.runtime_eol = true;
+                            }
                         }
                     } catch (Error e) {
                         warning ("Error while fetching remote ref: %s", e.message);
@@ -573,10 +574,6 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
             if (!(e is Flatpak.Error.ABORTED)) {
                 throw e;
             }
-        }
-
-        if (package != null) {
-            package.runtime_eol = runtime_eol;
         }
 
         return download_size;
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -498,7 +498,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
 
         var added_remotes = new Gee.ArrayList<string> ();
 
-        uint deprecated_dependencies_number = 0;
+        bool runtime_eol = false;
         try {
             var transaction = new Flatpak.Transaction.for_installation (installation, cancellable);
             transaction.add_default_dependency_sources ();
@@ -537,7 +537,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
                         var remote_ref = installation.fetch_remote_ref_sync (remote_name, kind, name, arch, branch, cancellable);
 
                         if (remote_ref.get_eol () != null || remote_ref.get_eol_rebase () != null) {
-                            deprecated_dependencies_number++;
+                            runtime_eol = true;
                         }
                     } catch (Error e) {
                         warning ("Error while fetching remote ref: %s", e.message);
@@ -576,7 +576,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
         }
 
         if (package != null) {
-            package.deprecated_dependencies_number = deprecated_dependencies_number;
+            package.runtime_eol = runtime_eol;
         }
 
         return download_size;
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -494,6 +494,33 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
         var flatpak_ref = Flatpak.Ref.parse (bundle_id);
         bool is_app = flatpak_ref.kind == Flatpak.RefKind.APP;
 
+        //  var remote_ref = installation.fetch_remote_ref_sync (string remote_name, Flatpak.RefKind kind, string name, string? arch, string? branch, cancellable);
+
+        installation.list_remotes ().foreach ((remote) => {
+            var elements = bundle_id.split ("/", 4);
+            
+            var remote_name = remote.get_name ();
+            //  var name = installation.get_id ();
+            var name = elements[1];
+            var arch = elements[2];
+            var branch = elements[3];
+
+            try {
+                var remote_ref = installation.fetch_remote_ref_sync (remote_name, flatpak_ref.kind, name, arch, branch, cancellable);
+                print ("is_app: %s\n", is_app ? "true" : "false");
+                print ("remote_name: %s\n", remote_name);
+                print ("name: %s\n", name);
+                print ("arch: %s\n", arch);
+                print ("branch: %s\n", branch);
+
+                print ("eol: %s\n", remote_ref.get_eol ());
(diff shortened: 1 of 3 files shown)


src/Views/AppInfoView.vala
```diff
@@ -743,6 +751,22 @@ namespace AppCenter.Views {
...
+                    _("Uses a runtime that is no longer supported and does not receive security updates"),
+                    "software-update-urgent-symbolic"
+                );
+
+                oars_flowbox.insert (end_of_life, 0);
+                oars_flowbox.show_all ();
+
+                if (oars_flowbox.get_children ().length () > 0) {
+                    oars_flowbox_revealer.reveal_child = true;
+                }
```
meisenzahl, line 768:
```suggestion
                oars_flowbox_revealer.reveal_child = true;
```

src/Core/FlatpakBackend.vala
```diff
@@ -498,6 +498,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
 
         var added_remotes = new Gee.ArrayList<string> ();
 
+        uint deprecated_dependencies_number = 0;
```
cassidyjames, line 501:
```suggestion
        bool runtime_eol = false;
```

src/Core/FlatpakBackend.vala
```diff
@@ -527,6 +528,21 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+                    try {
+                        var remote_name = entry.get_remote ();
+                        var kind = entry_ref.kind;
+                        var name = entry_ref.name;
+                        var arch = entry_ref.arch;
+                        var branch = entry_ref.branch;
+                        var remote_ref = installation.fetch_remote_ref_sync (remote_name, kind, name, arch, branch, cancellable);
+
+                        if (remote_ref.get_eol () != null || remote_ref.get_eol_rebase () != null) {
+                            deprecated_dependencies_number++;
```
cassidyjames, line 540:
```suggestion
                            runtime_eol = true;
```

src/Core/FlatpakBackend.vala
```diff
@@ -559,6 +575,10 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
             }
         }
 
+        if (package != null) {
+            package.deprecated_dependencies_number = deprecated_dependencies_number;
```
cassidyjames, line 579:
```suggestion
            package.runtime_eol = runtime_eol;
```

src/Core/Package.vala
```diff
@@ -37,6 +37,8 @@ public class AppCenterCore.Package : Object {
     public const string APPCENTER_PACKAGE_ORIGIN = "appcenter";
     private const string ELEMENTARY_STABLE_PACKAGE_ORIGIN = "elementary-stable-focal-main";
 
+    public uint deprecated_dependencies_number { get; set; default = 0U; }
```
cassidyjames, line 40:
```suggestion
    public bool runtime_eol { get; set; default = false; }
```

src/Views/AppInfoView.vala
```diff
@@ -743,6 +751,23 @@ namespace AppCenter.Views {
 
             var size = yield package.get_download_size_including_deps ();
             size_label.update (size, package.is_flatpak);
+
+            if (package.deprecated_dependencies_number > 0 && !is_eol_shown) {
```
cassidyjames, line 755:
```suggestion
            if (package.runtime_eol && !is_eol_shown) {
```

src/Views/AppInfoView.vala
```diff
@@ -743,6 +751,23 @@ namespace AppCenter.Views {
...
+                is_eol_shown = true;
+                var end_of_life = new ContentType (
+                    _("End of Life"),
```
cassidyjames, line 758: Ironically, even though I'm aligning on this terminology in code, I think it won't mean much to users. We should probably do something more straightforward, like:

```suggestion
                    _("Outdated"),
```

src/Views/AppInfoView.vala
```diff
@@ -743,6 +751,23 @@ namespace AppCenter.Views {
...
+                    ngettext (
+                        "Uses a runtime that is no longer supported and does not receive security updates",
+                        "Uses runtimes that are no longer supported and do not receive security updates",
+                        package.deprecated_dependencies_number
+                    ),
```
cassidyjames, line 763: Trying to make this more meaningful to users; also, if we change this to "technologies" instead of runtimes, we don't need to care about the exact number of runtimes. :)

```suggestion
                    _("Built with older technologies that may not work as expected or receive security updates"),
```

cassidyjames (changes requested):
An initial pass; feel free to push back on any changes! But here's what I'm thinking:

- Flatpak seems to use `eol` as the end of life terminology consistently rather than `deprecated`, so I think it makes sense to align our code to that for clarity.
- However, End of Life might not really mean what users expect, especially when we're talking about the runtimes and not necessarily the app itself (unless I'm reading this wrong, in which case, oops!).
- So, I made the user-facing copy more generic; this also reflects somewhat how iOS and Android talk about apps made for an older version of the OS, but we can't really say "for an older version of elementary OS" because that's not exactly how runtimes work.
- Lastly, I don't see us needing to care about the number of EOL runtimes, so I refactored that into a bool. **Edit:** I see you started with an `is_eol` bool so this was basically just reverting that; feel free to change this further if it can be simplified.

cassidyjames:
I also want to note that this doesn't precisely close #1803; that issue is about a mismatch between the OS version and the runtime version for curated apps, while this PR focuses on EOL runtimes. I think both are valuable, but what @danrabbit is asking for is a little different; if I install an app build against the not-EOL OS 6 runtime on OS 7, it might be useful to still point out that the app was built for OS 6 and thus might be missing out on features/integration.

In that situation, I think EOL warning would take precedence, but we'd need to do some version matching and only look at curated apps, etc. I think that is a separate PR, though.

danrabbit:
Yeah for EOL I'd be inclined to use a much larger warning like a floating infobar

cassidyjames:
@danrabbit I think I'm still _fine_ with using our content warning space for EOL runtimes, as it keeps the security/updates/expectations information in one place, but I could see that argument, too. :thinking:

src/Core/FlatpakBackend.vala
```diff
@@ -498,7 +498,7 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
 
         var added_remotes = new Gee.ArrayList<string> ();
 
-        uint deprecated_dependencies_number = 0;
+        bool runtime_eol = false;
```
cassidyjames, line 501: Okay, I see why you pulled this out into its own uint before, but I think we can revert further and just use `package.runtime_eol` directly again?

```suggestion
```

src/Core/FlatpakBackend.vala
```diff
@@ -527,6 +528,21 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+                    try {
+                        var remote_name = entry.get_remote ();
+                        var kind = entry_ref.kind;
+                        var name = entry_ref.name;
+                        var arch = entry_ref.arch;
+                        var branch = entry_ref.branch;
+                        var remote_ref = installation.fetch_remote_ref_sync (remote_name, kind, name, arch, branch, cancellable);
+
+                        if (remote_ref.get_eol () != null || remote_ref.get_eol_rebase () != null) {
+                            runtime_eol = true;
```
cassidyjames, line 540:
```suggestion
                            if (package != null) {
                                package.runtime_eol = true;
                            }
```

src/Core/FlatpakBackend.vala
```diff
@@ -559,6 +575,10 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
             }
         }
 
+        if (package != null) {
+            package.runtime_eol = runtime_eol;
+        }
+
```
cassidyjames, line 581:
```suggestion
```

cassidyjames (commented):
Alright, this reverts back further to basically where your `is_eol` bool was.

cassidyjames (commented):
I'm happy with this as an initial pass UX wise. It adds value and uses existing mechanisms to present an important bit of information. I think I'd still like a pass from @elementary/desktop-developers in case I missed anything code-wise, though!

Screenshot from 2022-01-07 11-11-16

cassidyjames (commented):
_oops ignore this accidental double review_

PR #1806, opened 2022-01-07
Check for firmware updates

### LINKED ISSUE #1805: Check for firmware updates
### Problem

I found out yesterday by accidentally opening the firmware page in the system settings that there are updates for my hardware. Since firmware updates sometimes close critical security gaps, I think it is important that the user is informed about possible updates.

Bildschirmfoto von 2022-01-07 15 23 40

### Proposal

Check once a day if firmware updates are available and notify the user.

### Prior Art (Optional)

_No response_

meisenzahl:
Closes #1805

https://user-images.githubusercontent.com/10796736/148558985-a5c0fc94-6d3e-416c-80a6-fb4c5c55b92b.mp4

I think the functionality is very useful and important. I am open to moving that to another component. 🙂


cassidyjames:
I think @davidmhewitt had thought about this in the past? Does this make sense to live in AppCenter even if it's actually pointing to System Settings for the updates themselves?

danrabbit:
Yeah this might make more sense as part of Settings Daemon

meisenzahl:
> Yeah this might make more sense as part of Settings Daemon

@danrabbit if you want, I can prepare a PR.

danrabbit:
@meisenzahl that would be great!

cassidyjames:
I _think_ @davidmhewitt's thoughts were that pulling in all the AppStream stuff into the settings daemon and having yet-another-thing checking for updates didn't feel nice when we already have AppCenter sitting here with the check-for-updates logic and everything, so I _think_ he said it did make sense in AppCenter. I just wanted to confirm that before we get too into the weeds here or implementing it in the settings daemon, though.

cassidyjames:
@davidmhewitt from Slack:

>If it's just a notification that points to system settings, then either AppCenter or settings Daemon is probably the same
>But we have better infrastructure in AppCenter for displaying richer details if that's something we ever wanna do
>And it already checks for updates on a schedule

I'll stop pinging here and let you all sort that out because you know the codebases better than me, just trying to prevent wasted work. :smile:

danrabbit:
Yeah I think something to consider is that if the notification comes from AppCenter it would have the AppCenter icon etc and we would have to add an action in AppCenter to launch system settings

If it's in settings daemon it could use the system settings icon and it seems less of a hack to me to have settings daemon launch system settings.

Settings Daemon also already has some scheduling code for things like dark mode

danrabbit:
Gonna close this since there's a branch for doing this in Settings Daemon

PR #1807, opened 2022-01-07
Warn about OS runtime mismatch

meisenzahl:
Supersedes #1804


meisenzahl:
Okay, that was a bit more work than I thought 😅

I now show a warning in the `AppInfoView` based on the runtime version. The logic to generate the warning text is implemented in `AppInfoView`.

It works pretty well, but doesn`t look very pretty...

@elementary/ux I am definitely grateful for UX suggestions 😁

Bildschirmfoto von 2022-01-08 08 48 58

meisenzahl:
This pull request targets `master` and is based on #1804. Let me know if you want to handle #1804 first and this separately.

cassidyjames (changes requested):
@meisenzahl sorry for any misunderstanding, but I think this case is less critical than an EOL runtime; @danrabbit was saying she thought _EOL runtimes_ might necessitate an inline infobar (but I'm happy with the `ContentType` implementation for now). I think version mismatches would also make sense as a `ContentType` implementation, perhaps something like this:

>:arrow_down: (maybe `software-update-available-symbolic`?)
>**Outdated**
>Built for an older version of elementary OS; might not support the latest features

meisenzahl:
@cassidyjames okay I will address the points.

Do we want to show warning only for major version differences? So for example on OS 7 for an OS 6 app?

What about an OS 6 app on OS 6.1?

What about an app that uses a next or daily runtime?

cassidyjames:
@meisenzahl 

>Do we want to show warning only for major version differences? So for example on OS 7 for an OS 6 app? What about an OS 6 app on OS 6.1?

I think major version makes sense; minor version bumps is more of a concern for us when reviewing apps but shouldn't have potentially breaking changes like a major mismatch could.

>What about an app that uses a next or daily runtime?

Good question; I imagine we could either ignore/not warn for now, or if we want, we could badge it as "unstable" or something, but I think that's outside the scope of this PR.

meisenzahl:
I refactored the code and introduced `RuntimeStatus`:

```csharp
public enum RuntimeStatus {
    UP_TO_DATE,
    END_OF_LIFE,
    MAJOR_OUTDATED,
    MINOR_OUTDATED,
    UNSTABLE;
}
```

This simplifies the display of appropriate warnings:

## End of Life

Bildschirmfoto von 2022-01-10 23 14 08

## Built for older version

Bildschirmfoto von 2022-01-10 23 13 07

## Unstable

Bildschirmfoto von 2022-01-10 23 13 46

If you don't like this, we can revert https://github.com/elementary/appcenter/pull/1807/commits/25da400a2374821fe74980635a5dd91411610a65.

Marukesu:
One extra case is when the SDK is the Runtime, for example, in code.

danrabbit (approved):
These are probably not the right icons to use since these icons are for when an update is available, and for these apps an update is not available. But, we can fix that before the next release. Nice work on this feature :)

PR #1816, opened 2022-01-16
Show only curated apps in "Other Apps" footer

### LINKED ISSUE #1814: Mix of curated and non-curated apps in "Other Apps" footer
### What Happened?

If I go to a curated app like Dippi, I can see a mix of curated and non-curated apps in the "Other Apps by Cassidy James Blaede" footer, like my Plausible app on Flathub. While this seems convenient, we're just string-matching the developer so it's possible we'd get false positives here, possibly lending more credibility than is due to an app that's not even maintained by its developer on Flathub.

### Steps to Reproduce

1. Ensure you have the Flathub remote added
2. Browse to my Dippi app
3. Look at the "Other Apps" footer

Also easily reproducible/testable by looking at a GNOME app like GNOME Web

### Expected Behavior

I think we should only show the "Other Apps" footer on curated apps; it's the only case where we've actually reviewed the apps and confirmed the developer is who they say. 

At the very least, we should not show non-curated apps in the footer of curated apps.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

meisenzahl:
Closes #1814

## Before

Bildschirmfoto von 2022-01-16 11 34 17

## After

Bildschirmfoto von 2022-01-16 11 36 53


cassidyjames:
So my question here is: should we always only show curated apps? Or should we treat the footer differently on non-cursted app info pages?

Right now if you go to Plausible, you see my other apps because they share the same name. But they're from a different remote which seems weird. Should we restrict by remote? Should we only show the footer on curated apps?

Thoughts @elementary/ux ?

danrabbit:
@meisenzahl I'd like to get away from making alt stores second class. I think if there's an issue with making sure that authors are actually the same, it would make more sense to show other apps from the same remote than to restrict this feature to only apps from the AppCenter remote

danrabbit:
Converting to draft since there hasn't been any movement here in a couple months

danirabbit:
Closing since there hasn't been any movement here in a long time

PR #1835, opened 2022-04-27
Core: fix appstream deprecations

danrabbit:
* We can no longer modify cache flags
* `metadata_locations` functions are replaced by `extra_data_locations`
* `add_component` is replaced by `add_components`


tintou:
You technically don't have to break support for Odin with the use of `#if HAS_APPSTREAM_0_15`

danrabbit:
@tintou dunno why I didn't do that originally :sweat_smile:  fixed

tintou (approved):

PR #1837, opened 2022-04-28
Homepage: more accurate navigation

danrabbit:
Helps with #1769

* Replace a pretty opaque method with some unrelated functions with smaller methods. This makes things more explicit and easier to follow.
* Make navigation more accurate and DRY


src/MainWindow.vala
```diff
@@ -444,35 +441,35 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         selected_package = package;
     }
 
-    private void view_opened (string? return_name, bool allow_search, string? custom_header = null, string? custom_search_placeholder = null) {
+    public void set_return_name (string? return_name) {
         if (return_name != null) {
             if (return_button_history.peek_head () != return_name) {
                 return_button_history.offer_head (return_name);
             }
```
jeremypw, line 448: The two if clauses could now be combined.

jeremypw (approved):
Nice clarification of the code!  No regressions found.

PR #1838, opened 2022-04-29
AppInfoView: new icon and copy for old runtimes

### LINKED ISSUE #1812: Outdated runtime icons have opposite implication
### What Happened?

When apps are built against outdated runtimes, they receive "update-available" namespaced icons. However, the problem is that there _isn't_ an update available. 

### Steps to Reproduce

Screenshot from 2022-01-11 13 08 47

### Expected Behavior

We should probably have a different icon. Maybe an update with a slashy would work

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

danrabbit:
Fixes #1812 

Screenshot from 2022-04-29 11 40 41

* Remove superfluous references to OS
* Why waste time say lot word when few word do trick?


micahilbery (approved):
Copy is shorter and sounds better. This icon is cute :spider_web: :black_heart:

PR #1856, opened 2022-05-25
CategoryView: highlight recently updated apps

danrabbit:
Show recently updated apps at the top of category views


src/Views/CategoryView.vala
```diff
@@ -130,23 +152,52 @@ public class AppCenter.CategoryView : Gtk.Stack {
                 child.destroy ();
             }
 
+            grid.add (recently_updated_box);
```
jeremypw, line 155: Populate can be called more than once but `recently_updated_box` is never removed from the grid so a second attempt to add it would be an error.

Is it possible there are zero recently updated packages?  In that case this grid should be hidden.

jeremypw (changes requested):
In general looks good and works as expected. One possible problem noted inline.

jeremypw:
What is the criterion for a "recently updated app"? Using this PR and looking at the "Maths, Science & Engineering" category, the first entry in "Recently Updated" (AlphaPlot) was last updated in 2016 (!).

danrabbit:
@jeremypw okay, I've made it so we don't show apps that haven't been updated in over 6 months and we now make sure to remove/not add the flowbox if it's empty

jeremypw (approved):
This now looks good to go.  I noticed the "Recently updated" set on the Home page appears to used different criteria from this app - is that intentional?  There are apps appearing in this set (e.g. "Gnonograms") that are not included in the Recently updated set in the corresponding Catergory View.

PR #1857, opened 2022-05-27
Do not show another package before transition complete

jeremypw:
This prevents a second click on e.g. the banner or a recently updated item on the Home Page from loading duplicate views.


danrabbit (approved):
Makes sense to me! I can't reproduce the issue you're experience but this seems like a sensible solution.

PR #1858, opened 2022-05-27
Provide default case for switches

jeremypw:
Silences some compile warnings.  Assumes missed cases do not need handling.


danrabbit (changes requested):
for the appstream icons, it might be better to explicitly include `AppStream.IconKind.UNKOWN` in case a new IconKind is introduced we probably would want this to warn again

I wonder if it would be better to do the same with `Pk.ProgressType` so we're at least documenting the status' we're ignoring and making sure we know it's intentional: https://valadoc.org/packagekit-glib2/Pk.ProgressType.html

jeremypw:
I added a warning for unknown AppStream.IconKind but I have held off from doing the same for Pk.ProgressType as that resulted in the terminal getting spammed with lots of warnings for each package updated.

danrabbit:
@jeremypw in that case maybe we shouldn't touch the Pk.ProgressType since that sounds like maybe an issue that needs to be addressed and hiding the warning here just hides the issue instead of actually fixing it

jeremypw:
Sorry I misunderstood your original comment. Hope I have got it right now.

tintou (approved):

PR #1859, opened 2022-05-27
AppListView → SearchView

danrabbit:
Renames AppListView to SearchView because this is the only thing it's used for. With this cleared up, we can move search internals from HomePage to SearchView in a future branch.

No logic changes in this branch, git just isn't picking up on the rename because of indentation changes due to the namespace change


jeremypw (approved):
Makes sense.

Maybe in future the Search view could be divided into Paid, Curated and non-Curated like Category view?

danrabbit:
I'm a little worried about having search view do anything that's not presenting the closest possible match. Something that's really annoying for example is search on the iOS AppStore where you get sponsored results up top or you search for the exact name of an app and get a match for a competitor. It can be a not great experience

jeremypw:
That's fair.  I agree the closest matches should appear first. Maybe the paid/curated/non-curated info could also be exposed in the view?

danrabbit:
Yeah or filters might be another way to accommodate that desire

PR #1860, opened 2022-05-29
AppListUpdateView: move list population from InstalledView to here

danrabbit:
`InstalledView` only handles navigation while `AppListUpdateView` is the real view. Move populating the list to the list itself. In the future branch this makes it easier to removed `InstalledView` so that we only have one Hdy.Deck to navigate all views

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -24,6 +24,8 @@ namespace AppCenter.Views {
     public class AppListUpdateView : AbstractAppList {
         private Gtk.SizeGroup action_button_group;
         private bool updating_all_apps = false;
+        private Cancellable refresh_cancellable;
+        private AsyncMutex refresh_mutex = new AsyncMutex ();
 
         construct {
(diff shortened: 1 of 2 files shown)


src/Views/AppListUpdateView.vala
```diff
@@ -67,6 +69,40 @@ namespace AppCenter.Views {
...
+                });
+            });
+        }
+
+        private async void get_apps () {
+            refresh_cancellable.cancel ();
+
+            yield refresh_mutex.lock ();
+
+            refresh_cancellable.reset ();
```
jeremypw, line 91: According to Valadoc.org:
> Note that it is generally not a good idea to reuse an existing cancellable for more operations after it has been cancelled once, as this function might tempt you to do. The recommended practice is to drop the reference to a cancellable after cancelling it, and let it die with the outstanding async operations. You should create a fresh cancellable for further async operations.

However, this is existing code so not necessarily blocking for this PR.

jeremypw (approved):
No regressions or additional warnings found.

PR #1863, opened 2022-06-21
AppListUpdateView: Don't re-use cancellable

danrabbit:
As per https://github.com/elementary/appcenter/pull/1860#discussion_r901883333 

Does this make sense? I'm assuming the original intent here is that if the method is called again before it finishes that we want to cancel the original one?


jeremypw:
@danrabbit I have taken the liberty if amending this PR according to my previous comment.  Is it OK?

danrabbit:
@jeremypw thanks for following up on this! This looks good to me :)

PR #1744, opened 2021-11-03
AppInfoView: Add scalable, full-width screenshots

isantop:
Adds scalable full-width screenshots to the AppInfoView. This allows the screenshots to shrink at narrow window sizes while still displaying large, beautiful screenhots at larger sizes.

Requires #1743

PR #1870, opened 2022-08-11
Use DrawingArea for screenshots

danirabbit:
Based on #1744 

Use a Gtk.DrawingArea for screenshots so that they can be resized to fit the window

Todo:
- [ ] Screenshots preserve extra vertical space when resized. Can we shrink that?
- [ ] We've lost the fallback image when screenshots can't be loaded


danrabbit:
Sometimes we get some weird centering:

Screenshot from 2022-08-11 14 18 49

src/Widgets/Screenshot.vala
```diff
@@ -0,0 +1,86 @@
...
+
+    protected override void get_preferred_height_for_width (int width, out int min, out int nat) {
+        min = width * pixbuf_height / pixbuf_width;
+        nat = min;
+    }
+
+    protected override void get_preferred_width_for_height (int height, out int min, out int nat) {
+        min = height * pixbuf_width / pixbuf_height;
+        nat = min;
+    }
```
jeremypw, line 85: Not sure whether this needs casting to double and back in order to avoid undesired rounding? Depends whether the division or multiplication takes precedence.

danirabbit:
Gonna close this since it seems like we'll be merging #1896 imminently

PR #1873, opened 2022-08-15
MainWindow: Use icons for view mode switcher and center search

danrabbit:
Part of the redesign discussion at https://github.com/elementary/appcenter/discussions/1467#discussioncomment-2676916

Make sure the search bar is usable at small window widths—As brought up by @isantop—by shrinking the size of the navigation buttons. Also prioritize search over navigation since automatic updates should reduce the need to visit the updates page.

Also removes deprecated Granite.ModeButton as a bonus

Screenshot from 2022-08-15 12 39 28
Screenshot from 2022-08-15 12 39 45
Screenshot from 2022-08-15 12 43 19


jeremypw (changes requested):
The only thing I've noticed so far is that the "Installed View" button is not highlighted when on that view - the highlight remains on the "Home" button.
Screenshot from 2022-08-23 10 12 46

Probably outside the scope of this PR but not sure the search entry and its placeholder should still appear when it cannot be used (the view mode buttons are hidden when unusable)

danrabbit:
@jeremypw That's the keyboard focus ring, so even if I changed them to be toggle buttons it would stay the same you'd just also have a ring on the toggled button:

Screenshot from 2022-08-23 11 46 51

Regarding the search entry, I definitely want to change that in a future branch but I was trying to keep diffs relatively small

jeremypw:
Should the viewmode icons accept keyboard focus if they cannot be activated with the keyboard? Also, on the Home page, if the keyboard focus is moved to one of the "Recently Updated" entries then the homeview icon is clicked on, the keyboard focus moves to the icon. But if you then click the other icon the focus does not move.  It seems rather confusing.  However handling of keyboard focus is not consistent in other apps as well so I do not think it is blocking for this PR.

jeremypw (approved):

danrabbit:
View mode buttons are keyboard activatable on this end! 

Yeah I'm not sure how Gtk decides whether or not to move the focus ring when something is clicked :sweat_smile:

PR #1874, opened 2022-08-21
Single deck nav

danrabbit:
Builds on #1873 

Use a single deck for navigation and make the updates view just another view instead of its own deck


jeremypw:
Looks like a nice simplification and works as expected so far.

danrabbit:
@jeremypw Do you think it's worth trying to merge #1873 first as a smaller branch, or is it manageable enough here that I should just close that other PR?

jeremypw:
Yeah, #1873 should probably be merged first - sorry I missed that.  I have now review that PR.

jeremypw:
The search entry moves noticeably between the home and installed views which is a bit jarring but I do not think it is enough to block this PR. Approving.

jeremypw (approved):

PR #1875, opened 2022-08-21
InstalledPackageRowGrid: release notes in a dialog

### LINKED ISSUE #543: Larger changelog expand click targets.
Right now expanding the changelog for updates is difficult because the click targets to do so are small. I miss clicked into the app page multiple times using a mouse. I would imagine it even harder on a touch screen.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/51541554-larger-changelog-expand-click-targets?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1137: Preface release description if it starts with a list
If a release description starts with a list instead of a paragraph, we should preface it with a "Details:" paragraph. This will prevent this awkward folding:

Screenshot from 2019-11-15 18-10-46

danrabbit:
Fixes #543 
Fixes #1137 

Fixes some release notes quirks by putting them in a dialog instead of inline in the updates view:

Screenshot from 2022-08-21 14 12 48

Also helps with window width for responsive. Both of these are at their minimum width:

Before:
Screenshot from 2022-08-21 14 16 26

After:
Screenshot from 2022-08-21 14 15 46

### DIFF
--- a/src/Widgets/AppContainers/InstalledPackageRowGrid.vala
+++ b/src/Widgets/AppContainers/InstalledPackageRowGrid.vala
@@ -21,15 +21,8 @@ public class AppCenter.Widgets.InstalledPackageRowGrid : AbstractPackageRowGrid 
     public signal void changed ();
 
     private AppStream.Release? newest = null;
-    private Gtk.Expander release_expander;
     private Gtk.Label app_version;
-    private Gtk.Label release_description;
-    private Gtk.Label release_expander_label;
-    private Gtk.Label release_single_label;
-    private Gtk.Revealer release_stack_revealer;
-    private Gtk.Stack release_stack;
-
-    private static Gtk.SizeGroup info_size_group;
+    private Gtk.Revealer release_button_revealer;
 
     public InstalledPackageRowGrid (AppCenterCore.Package package, Gtk.SizeGroup? action_size_group) {
         Object (package: package);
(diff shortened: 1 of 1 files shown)


lokix22:
I think it would look cleaner if info icon would be inline and just after the package name.

jeremypw:
How it looks with info button inline, next to app name.

Screenshot from 2022-08-22 16 43 32

jeremypw:
It could be argued that the info is more about the update than the program itself so should be nearer the update button.

jeremypw:
I would be happy to approve this PR as is and maybe tweak the appearance in another PR if required.

lokix22:
I agree, It doesn't look as inconspicuous as in "Up to x GB" line. Original PR is better that what was before for sure. The other option is tiny "view changes" flat button with rounded rectangle outline right next to version number, preferably at same distance as "Up to x GB" line. But this is all nit picking.

danrabbit:
@jeremypw If you wouldn't mind! I'd like to go through a few iterations in this view.

One idea I had in mind was making the version show the difference like we did in the firmware updates view with "x.x.x → y.y.y" and maybe have that be a text button that opens the dialog even. Could go a few different directions. The dialog itself could be spiced up in future versions too. So I'd just think of this as an intermediate step

jeremypw (approved):

PR #1878, opened 2022-08-23
AppInfoView: Add background and caption to screenshots

danrabbit:
Add an accent colored background behind screenshots and display the screenshot caption
* Makes screenshot captions more obvious instead of being hidden in a tooltip
* Better exposes off-screen screenshots
* Makes the draggable area more obvious
* Style "Screenshot not available" text so that the white space preserved looks less awkward"

Screenshot from 2022-08-23 13 04 47
Screenshot from 2022-08-23 13 05 25
Screenshot from 2022-08-23 13 05 16
Screenshot from 2022-08-23 13 04 26


jeremypw (approved):
Looks good to me and works as expected.

There seems to be a problem downloading some screenshots e.g. for Files but that also occurs in  master.  Will investigate.   May conflict with the DrawingArea PR.

PR #1880, opened 2022-08-23
Banner, AppInfoView: Use a single accent color, auto text color

danrabbit:
Supersedes #1722 

In the interest of adopting the single branding color that's now supported upstream in AppStream: https://github.com/ximion/appstream/issues/187


src/Widgets/Banner.vala
```diff
@@ -102,24 +102,28 @@ public class AppCenter.Widgets.Banner : Gtk.Button {
...
+                } else {
+                    var bg_rgba = Gdk.RGBA ();
+                    bg_rgba.parse (bg_color);
 
-        try {
+                    text_color = Granite.contrasting_foreground_color (bg_rgba).to_string ();
+                }
+            }
+
+            var colored_css = BANNER_STYLE_CSS.printf (bg_color, text_color);
```
jeremypw, line 122: Is it OK to (potentially) print `null` into the css here?

jeremypw (commented):
Works as expected during testing.  Just a question about the effect of printfing a null string into the css string.

src/Widgets/Banner.vala
danrabbit: Oops you're right, that is problematic. Should be fixed now

jeremypw (approved):
Good to go now.

danrabbit:
Woo, thanks for thorough review!

PR #1882, opened 2022-08-24
Move Deck to MainWindow

danrabbit:
This moves the Hdy.Deck to `MainWindow`, making `Homepage` just another view. Also needed for Gtk4 since you can't subclass Adw.Leaflet


jeremypw:
I havent time to look at this in detail today but it appears new SearchViews keep being created but as far as I can tell are never destroyed.

danrabbit:
Hm, let me know if you have steps to reproduce. I'm seeing views being destroyed on navigate...

jeremypw:
I inserted a message before the `new SearchView ()` line and inserted a `~SearchView ()` function in the SearchView class which also printed a message.  Whenever I do a new search (after returning to Home) a new SearchView is created but the `~SearchView ()` never runs.

jeremypw:
Looks like this also occurs in master and is not limited to this PR so I am inclined approve this anyway. It can be investigated later.

jeremypw (approved):
Works as expected

PR #1884, opened 2022-08-24
AppInfoView: use a flowbox for links

danrabbit:
Use a flowbox so we don't get unreadable labels at small sizes. While we're here, use Gtk.show_uri_on_window to avoid focus stealing prevention

## Before
Screenshot from 2022-08-24 12 59 22

## After
Screenshot from 2022-08-24 12 53 24


jeremypw (approved):
Code looks good and works as expected

PR #1885, opened 2022-08-24
MainWindow: remove window position code

danrabbit:
Since we can't position the window anymore in Wayland/Gtk4 might as well get rid of this now


jeremypw (approved):
Makes sense

PR #1887, opened 2022-08-24
Package: expose ComponentKind as property

danrabbit:
Instead of having several bool properties, just expose ComponentKind as a property. Makes it more obvious what the source of truth is


src/Views/AppListUpdateView.vala
```diff
@@ -122,7 +122,7 @@ namespace AppCenter.Views {
             var needs_update = package.state == AppCenterCore.Package.State.UPDATE_AVAILABLE;
 
             // Only add row if this package needs an update or it's not a font or plugin
-            if (needs_update || (!package.is_plugin && !package.is_font)) {
+            if (needs_update || (!(package.kind == AppStream.ComponentKind.ADDON) && !(package.kind == AppStream.ComponentKind.FONT))) {
```
jeremypw, line 125: Isn't `package.kind != AppStream.ComponentKind.ADDON`  etc a bit clearer than `!(package.kind == AppStream.ComponentKind.ADDON)`?

src/Views/CategoryView.vala
```diff
@@ -236,7 +236,7 @@ public class AppCenter.CategoryView : Gtk.Stack {
         var packages = new Gee.TreeSet <AppCenterCore.Package> ();
         new Thread<void> ("get_packages", () => {
             foreach (var package in AppCenterCore.Client.get_default ().get_applications_for_category (category)) {
-                if (!package.is_plugin && !package.is_font) {
+                if (!(package.kind == AppStream.ComponentKind.ADDON) && !(package.kind == AppStream.ComponentKind.FONT)) {
```
jeremypw, line 239: See previous comment about negated equal.

src/Views/SearchView.vala
```diff
@@ -70,7 +70,7 @@ public class AppCenter.SearchView : AbstractAppList {
 
     private void add_row_for_package (AppCenterCore.Package package) {
         // Don't show plugins or fonts in search and category views
-        if (!package.is_plugin && !package.is_font) {
+        if (!(package.kind == AppStream.ComponentKind.ADDON) && !(package.kind == AppStream.ComponentKind.FONT)) {
```
jeremypw, line 73: See previous comment about negated equal.

jeremypw (changes requested):
Just a tiny comment re code style.

src/Views/AppListUpdateView.vala
danrabbit: omg yeah, why did I write it like that lmao

danrabbit:
@jeremypw oops, thanks. I dunno what I was thinking with that. Should be fixed now

jeremypw (approved):

PR #1888, opened 2022-08-24
Move NetworkInfoBar to MainWindow

danrabbit:
Don't need to create a whole class for an infobar that's only used once


jeremypw (approved):
Code makes sense and works as in master.
Maybe consider a newline between primary (bold) text the secondary text? But that can be done in another PR if required.

PR #1886, opened 2022-08-24
Replace Gtk.Grid with Gtk.Box

danrabbit:
Replace Gtk.Grid with Gtk.Box where applicable in preparation for Gtk 4. Should be able to rebase merge.

PR #1889, opened 2022-08-25
Use explicit properties

danrabbit:
Preparing for Gtk4 port. This can be rebase merged.

Didn't touch a couple classes that are already updated in #1886 to avoid conflicts


jeremypw (approved):
Code appears equivalent and makes sense.

PR #1890, opened 2022-08-25
AppInfoView: show share for appcenter origin

danrabbit:
Instead of checking a bunch of things to decide if a package is shareable, check that the package origin is AppCenter since that's what we actually require


jeremypw (changes requested):
This introduces a change in behaviour - only Appcenter flatpak apps can be shared whereas in master, Elementary native apps (e.g. Files) can also be shared.

Please confirm whether this is desired (or fix).

danrabbit:
Ah I didn't realize apps from the stable PPA do appear to be available on the appcenter website, so closing!

https://appcenter.elementary.io/io.elementary.files/

PR #1891, opened 2022-08-25
Package: Remove should_nag_update

danrabbit:
afaict in all of the places this is used we're already checking for `update_available`


jeremypw (approved):
I have no paid apps installed so cannot fully test but the code appears to be logically equivalent to me.

PR #1892, opened 2022-08-25
Package: use AppCenter origin in should_pay

danrabbit:
`is_native` is kind of a roundabout way of checking package origin, and `is_os_updates` would always be false for the `APPCENTER` origin, so just check that

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -149,7 +149,7 @@ public class AppCenterCore.Package : Object {
 
     public bool should_pay {
         get {
-            if (!is_native || is_os_updates) {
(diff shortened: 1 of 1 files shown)


src/Core/Package.vala
```diff
@@ -149,7 +149,7 @@ public class AppCenterCore.Package : Object {
 
     public bool should_pay {
         get {
-            if (!is_native || is_os_updates) {
+            if (component.get_origin () != APPCENTER_PACKAGE_ORIGIN) {
```
jeremypw, line 152: Don't you need to check the origin is not ELEMENTARY_STABLE_PACKAGE_ORIGIN as well?

jeremypw (approved):
Makes sense

PR #1893, opened 2022-08-26
meson: require appstream >= 0.15.2

decathorpe:
It appears that the "branding" property is new in appstream 0.15.2:
https://github.com/ximion/appstream/blob/v0.15.2/NEWS

The use of this API was recently added here:
https://github.com/elementary/appcenter/pull/1883


jeremypw (approved):
Master is already non-compilable on Elementary6.1 and this PR compiles on Elementary 7 so no obvious problem in merging this.

PR #1895, opened 2022-08-27
CategoryView: don't subclass Gtk.FlowBox

danrabbit:
We can't subclass Gtk.FlowBox in Gtk 4, so make a box subclass instead


jeremypw (approved):
Code looks functionally equivalent and no regressions found

PR #1896, opened 2022-08-27
Port to Gtk 4

### LINKED ISSUE #1590: When an app installs, the cancel button is not properly highlighted at the right edge
## What Happened

See title

## Expected Behavior

Proper highlighting

## Steps to Reproduce

1. Install a big app 
2. Notice the right edge of the Cancel button as the app downloads

## Platform Information

* The version of elementary OS you are using, or other operating system

6.0

* The version of the software you are using such as "1.0", "Compiled from git", or "Latest release" if you're not sure but you have run updates

3.6.2

[Screenshot](https://i.imgur.com/gtnh8TK.png)

### LINKED ISSUE #1795: Port to GTK4

### LINKED ISSUE #1940: Weird previous/next buttons behaviour in screenshots carousel
### What Happened?

After scrolling screenshots to the end, previous screenshot button don't appear until you hover them or don't hover the screenshot

### Steps to Reproduce

1. Open any app with more then 1 image
2. Try use screenshot carousel

### Expected Behavior

Buttons are always visible

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #1590
Fixes #1795
Fixes #1940

I've tried to comment all non-obvious changes here. A lot of the diff is just minor API changes. Hide whitespace changes might be useful since there's a few try/catches removed or null or type checks added

Categories:
* css: Fixes for some changes to how Gtk.CSS renders gradients

Application:
* Use new method for window save/restore size
* New way to check focus
* No more destroy, connect to `close`

Package:
* It looks like we never copied remote icons before? But now this causes a crash when Gtk tries to set a Paintable from a File. So removed for now since this does nothing already

StripeDialog:
* Spinbutton no longer has `activates_default` so use a key controller
* Spinbutton no longer has `primary_icon_name` so parent a Gtk.Image

MainWindow:
* Leaflet handles mouse back button, so no need to do it ourselves

AppInfoView:
* There was some overcomplicated arrow mouse event stuff here. afaict the motion controller hides and shows arrows more accurately in this branch than master with fewer signals
* I used Gtk.Picture for screenshots instead of Gtk.Image which seems to obsolete #1870 

AbstractAppContainer:
* Made overflow visible in a revealer here so that button shadows don't get clipped


danrabbit:
Couple notes for myself:
- [ ] I think in the spin button key controller I need to check if the text entry has focus otherwise you won’t be able to use the buttons with the keyboard maybe
- [ ] I think in appinfo view dim label is being set on the entire box for content warnings and it should only be on the body text
- [ ] I might be able to make the arrows in the screenshot carousel simpler by setting icon size in CSS

lenemter:
You forgot to change the dependencies in GitHub Workflow: https://github.com/elementary/appcenter/blob/cf8af4baa7937a8aad9160df93f6a57e71898228/.github/workflows/main.yml#L21-L23

src/Dialogs/StripeDialog.vala
```diff
@@ -270,8 +281,17 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
...
+        custom_amount_controllerkey.key_released.connect ((keyval, keycode, state) => {
+            if (keyval == Gdk.Key.Return || keyval == Gdk.Key.KP_Enter) {
+                activate_default ();
+            }
+        });
+
         email_entry.changed.connect (() => {
-            email_entry.text = email_entry.text.replace (" ", "").down ();
+            if (" " in email_entry.text) {
+                email_entry.text = card_expiration_entry.text.replace (" ", "");
```
zeebok, line 292:
```suggestion
                email_entry.text = email_entry.text.replace (" ", "");
```

zeebok (approved):
The code looks good! I haven't cooked and tested. If you'd like me to do that, let me know and I'll happily find time to do that.

zeebok:
By and large everything is looking good. I did notice that when you go to a package's page it would have the title highlighted:
Screenshot from 2022-10-20 16 00 53

zeebok:
Also if you pick a different source (appcenter, system-wide, etc) it seems to place a new page on the stack (or leaflet is it called here?) so when you go back to the same app page for as many times as you changed the source.

danirabbit:
@zeebok thanks for testing! I'll look into those.

I'm thinking maybe it would be better anyways to release 4.0 and then we can merge after release so we have more time for testing in daily

danirabbit:
@zeebok it looks like that navigation bug also happens in master, can you open an issue report for that one please? :)

danirabbit:
I believe I fixed that label selection issue as well

meisenzahl:
I haven't had enough time to take a closer look at the code, but when testing under OS 7 with this AppCenter branch, I noticed the following errors:

## Missing app icons on the first start of AppCenter

Bildschirmfoto von 2022-11-01 22 44 21

```sh
...
(io.elementary.appcenter:7540): Gtk-WARNING **: 22:43:58.785: Failed to load icon /home/marius/.cache/io.elementary.appcenter/flatpak-metadata/user/icons/appcenter/64x64/com.github.eksanos.eksanos.png: Error opening file /home/marius/.cache/io.elementary.appcenter/flatpak-metadata/user/icons/appcenter/64x64/com.github.eksanos.eksanos.png: No such file or directory
...
```

## AppCenter is not responding and needs to be killed

```sh
** (io.elementary.appcenter:9522): CRITICAL **: 23:14:38.142: Package.vala:724: Unable to download remote icon: Not Found
```

## Random crashs when opening a category while "Checking for updates"

I get random memory access errors when opening a category while "Checking for updates".


https://user-images.githubusercontent.com/10796736/199353069-c74db336-4e7a-443b-b2e7-89e823aac262.mp4

zeebok:
> @zeebok it looks like that navigation bug also happens in master, can you open an issue report for that one please? :)

Sorry got real busy. Do you need me to open an issue still? I am not actually not seeing it with stable right now on my 6.1 install.

src/Core/Package.vala
```diff
@@ -709,10 +708,22 @@ public class AppCenterCore.Package : Object {
...
+                        // FIXME: Copy this async
+                        try {
+                            server_file.copy (local_file, FileCopyFlags.OVERWRITE, null, null);
+
+                            icon = new FileIcon (local_file);
+                            current_size = icon_width;
+                            current_scale = icon_scale;
+                        } catch (Error e) {
+                            critical ("Unable to download remote icon: %s", e.message);
+                        }
```
meisenzahl, line 725: I get random freezes of AppCenter with a warning in the console:

```sh
** (io.elementary.appcenter:9522): CRITICAL **: 23:14:38.142: Package.vala:724: Unable to download remote icon: Not Found
```

The issue could be related to this code.

src/Dialogs/StripeDialog.vala
```diff
@@ -270,8 +281,17 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
...
+            if (keyval == Gdk.Key.Return || keyval == Gdk.Key.KP_Enter) {
+                activate_default ();
+            }
+        });
+
         email_entry.changed.connect (() => {
-            email_entry.text = email_entry.text.replace (" ", "").down ();
+            if (" " in email_entry.text) {
+                email_entry.text = email_entry.text.replace (" ", "");
+            }
```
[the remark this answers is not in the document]
danirabbit: Yes because this now only runs if there is a space so converting to lowercase is not really relevant, and we also convert what we actually submit to lowercase in `on_pay_clicked ()` so this isn't really necessary

danirabbit:
Turns out gestures are currently broken in Gtk4/Adwaita for touchpad, so gonna shelve this for now

src/Views/AppInfoView.vala
```diff
@@ -187,7 +187,7 @@ public class AppCenter.Views.AppInfoView : AppCenter.AbstractAppContainer {
         };
 
         unowned var uninstall_button_context = uninstall_button.get_style_context ();
-        uninstall_button_context.add_class (Gtk.STYLE_CLASS_RAISED);
+        uninstall_button_context.add_class ("raised");
```
danirabbit, line 190: Double check this is working as intended

src/Views/Homepage.vala
```diff
@@ -25,31 +25,29 @@ public class AppCenter.Homepage : Gtk.Box {
...
         hexpand = true;
         vexpand = true;
 
-        banner_carousel = new Hdy.Carousel () {
-            allow_long_swipes = true
-        };
+        var banner_motion_controller = new Gtk.EventControllerMotion ();
 
-        banner_motion_controller = new Gtk.EventControllerMotion (banner_carousel) {
-            propagation_phase = CAPTURE
```
danirabbit, line 49: Double check this works without settings propagation phase

src/Widgets/ReleaseRow.vala
```diff
@@ -87,13 +88,13 @@ public class AppCenter.Widgets.ReleaseRow : Gtk.Box {
 
             var issue_linkbutton = new Gtk.LinkButton (issue.get_url ());
             issue_linkbutton.get_child ().destroy ();
-            issue_linkbutton.add (issue_label);
+            issue_linkbutton.child = issue_label;
```
danirabbit, line 91: Double check if this works or there's a simpler way to do this

danirabbit:
```
(io.elementary.appcenter:245070): Adwaita-CRITICAL **: 12:19:02.256: adw_carousel_get_nth_page: assertion 'n < adw_carousel_get_n_pages (self)' failed

(io.elementary.appcenter:245070): Adwaita-CRITICAL **: 12:19:02.257: adw_carousel_scroll_to: assertion 'GTK_IS_WIDGET (widget)' failed
```

danirabbit:
@leolost2605 since AppCenter is somewhere you touched recently could I get a review here? I'd like to do some more UI work but I really want to land GTK4 porting first :)

src/MainWindow.vala
```diff
@@ -348,36 +325,14 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
             }
         });
 
-        delete_event.connect (() => {
+        close_request.connect (() => {
             ((AppCenter.App) application).request_background.begin (() => destroy ());
 
-            return Gdk.EVENT_STOP;
+            return false;
```
leolost2605, line 331: Maybe this can be moved into the override bool close_request?
Also I think this should probably return true?

src/Views/CategoryView.vala
```diff
@@ -228,13 +226,13 @@ public class AppCenter.CategoryView : Gtk.Box {
...
+            flowbox.append (package_row);
         }
 
         public void clear () {
-            foreach (unowned var child in flowbox.get_children ()) {
-                child.destroy ();
-            }
+            while (flowbox.get_first_child () != null) {
+                flowbox.get_first_child ().destroy ();
+            };
```
leolost2605, line 235:
```suggestion
            }
```

leolost2605 (changes requested):
A few things I noticed during testing, I think most of them aren't blocking though

- If you've got only a few packages the packages grid in the update view is bottom aligned but I think that's also present in master
- Also I'm being greeted by a restart required infobar
image

- The backbutton doesn't have the back style applied but I've seen the same thing in settings so I think that's a granite/stylesheet issue?
- Switching between remotes plays an leaflet animation 
- There are a lot of deprecation warnings but I guess those can be fixed in follow up PRs

src/MainWindow.vala
danirabbit: Yeah I'm not sure why this was separate in master. Moved!

danirabbit:
> * If you've got only a few packages the packages grid in the update view is bottom aligned but I think that's also present in master

Yup that's in master too. Definitely wanna iterate here in a future branch

> * Also I'm being greeted by a restart required infobar

Ah because widgets are visible by default in GTK4 and we forgot to remove this when we removed the updates recently in another branch. Fixed!

> * The backbutton doesn't have the back style applied but I've seen the same thing in settings so I think that's a granite/stylesheet issue?

Yeah, pushed a fix to stylesheet :sweat_smile: 

> * Switching between remotes plays an leaflet animation

Yeah this might be a leaflet problem :thinking: When we move to NavigationView there's a property there for disabling animations, but I was trying to keep from blowing up the diff. Is this okay to fix in a follow up?

> * There are a lot of deprecation warnings but I guess those can be fixed in follow up PRs

Yeah I was trying to keep the diff here as small as I can

leolost2605 (approved):
Seems good enough to iterate on! 🎉

danirabbit:
Let's fucking goooo

PR #1898, opened 2022-08-27
Homepage: absorb CategoryFlowBox

danrabbit:
Alternative to #1897 

Homepage is a much smaller class now that it doesn't contain nav, so maybe this is more reasonable


jeremypw (approved):
Appears functionally equivalent and no regressions found

PR #1899, opened 2022-08-27
AppInfoView: absorb ReleaseListBox

danrabbit:
ListBox can't be subclassed in Gtk 4. This is such a simple class anyways there's not really a point. Fix a style issue while we're here by adding `background` class


jeremypw (approved):
Code appears functionally equivalent.

PR #1900, opened 2022-08-28
SharePopover: code style and DRY

danrabbit:
* Bump copyright
* Code style
* Since these are all `image-button` they don't also need `flat`
* construct the email button all at once
* DRY uri showing and use Gtk.show_uri_on_window to avoid focus stealing prevention
* animate with `popdown` instead of `hide`


jeremypw (approved):
LGTM - no regressions found.

PR #1903, opened 2022-08-29
Application: use active_window

danrabbit:
Gtk.Application already keeps track of the main window with `active_window` so we don't need to do it ourselves

* Use `active_window` in place of `main_window` where applicable
* Don't create style providers every time we activate, only when we create a new `MainWindow`
* Remove dead `HOMEPAGE` ifdef
* We always check for `show_updates` and `present` whether or not `active_window` is null so DRY
* Move updates badge code to `MainWindow` since there's no reason for it to be here


Marukesu:
> * Don't create style providers every time we activate, only when we create a new `MainWindow`

This should be done on `startup()`, since it's lives the entire lifetime of the Application.

danrabbit:
@Marukesu today I learned about `startup`! Will have to make lots of PRs

Marukesu:
You need to chain-up at the beginning of the function, see for example, [notifications](https://github.com/elementary/notifications/blob/91e6dcd61aced65c50624431efe9ef47464768db/src/Application.vala#L54-L76).

danrabbit:
@Marukesu that worked, thank you! Let me know if there's other things here that would make sense to move to `startup ()` from `construct` or `activate`

Marukesu:
The `Hdy.init()` call on main window can be moved to `startup()` too. Also, `Gtk.IconTheme.add_resource_path()` don't need to be called with the `/io/elementary/appcenter/icons/` path, the Gtk.Application class already do this to us.

src/MainWindow.vala
```diff
@@ -274,14 +274,22 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
             maximize ();
         }
 
+        show_all ();
```
Marukesu, line 277: IMHO, calling `show_all()` inside `construct()` is a bad pattern, since not all properties is actually with the right values.

Doing it for the box is fine though, since in Gtk4 child widgets are shown by default, and `present()`/`present_with_time()` will call `show()` internally.

```suggestion
        box.show_all ();
```

Marukesu (approved):
Code seems fine, not on OS7 so can't actually test this.

PR #1905, opened 2022-08-30
StripeDialog: code style and boxes

danrabbit:
Should reduce the diff when porting to Gtk 4
* Code style
* Use boxes
* `window.set_default ()` instead of `widget.has_default`. `has_default` is read-only in Gtk 4


jeremypw (approved):
Appears functionally equivalent. No regressions found in appearance.

PR #1906, opened 2022-08-30
Dialogs: use present and connect to response

danrabbit:
`dialog.run ()` and `show_all ()` no longer exist in Gtk 4. Can be rebase merged

Application: you can see InstallFailDialog already connects its response to `destroy ()`

InstalledPackageRowGrid: use a Granite.Dialog instead of a Hdy.Window. Don't know why it wasn't this way in the first place. Brain must have fell out of my head :upside_down_face: Saves a few lines and window type hints aren't available in Gtk 4 anyways

### DIFF
--- a/src/Widgets/AppContainers/InstalledPackageRowGrid.vala
+++ b/src/Widgets/AppContainers/InstalledPackageRowGrid.vala
@@ -81,7 +81,7 @@ public class AppCenter.Widgets.InstalledPackageRowGrid : AbstractPackageRowGrid 
             var releases_dialog = new ReleaseDialog (package) {
                 transient_for = (Gtk.Window) get_toplevel ()
             };
-            releases_dialog.show_all ();
+            releases_dialog.present ();
         });
     }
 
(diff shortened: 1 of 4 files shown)


src/Widgets/AppContainers/InstalledPackageRowGrid.vala
```diff
@@ -147,26 +146,20 @@ public class AppCenter.Widgets.InstalledPackageRowGrid : AbstractPackageRowGrid
...
                 margin_end = 12,
-                margin_bottom = 12,
-                margin_start = 12
+                margin_start = 12,
+                vexpand = true
             };
             releases_dialog_box.add (releases_title);
             releases_dialog_box.add (release_description);
-            releases_dialog_box.add (dialog_close_button);
+            releases_dialog_box.show_all ();
```
zeebok, line 156: Does Box really still use show_all instead of present?

zeebok (approved):

src/Widgets/AppContainers/InstalledPackageRowGrid.vala
danrabbit: Yeah because this branch is still gtk3

PR #1907, opened 2022-09-01
Homepage categories

danrabbit:
Reduces diff for Gtk4 a bit:

* Don't add an empty image widget to category cards with no icon
* Use calc to avoid rounding errors
* Set card padding and margin in CSS to make our lives a little simpler

### DIFF
--- a/data/styles/categories.css
+++ b/data/styles/categories.css
@@ -179,27 +179,27 @@
     background-image:
         linear-gradient(
             to right,
-            transparent 15px,
-            alpha(white, 0.05) 15px,
-            alpha(white, 0.05) 16px,
-            transparent 16px
+            transparent calc(0.5em - 1px),
+            alpha(white, 0.05) calc(0.5em - 1px),
+            alpha(white, 0.05) 0.5em,
+            transparent 0.5em
         ),
         linear-gradient(
             to bottom,
-            transparent 15px,
-            alpha(white, 0.05) 15px,
-            alpha(white, 0.05) 16px,
-            transparent 16px
+            transparent calc(0.5em - 1px),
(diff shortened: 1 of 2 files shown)


data/styles/categories.css
```diff
@@ -27,6 +27,9 @@
         0 3px 5px alpha(black, 0.1);
     font-size: 32px;
     font-weight: 300;
+    margin: 1.333rem; /* 12px */
+    padding: 1.333rem;
+    min-height: 3em;
```
jeremypw, line 32: You changed ems to px elsewhere but used ems here - is that OK?

src/Views/Homepage.vala
```diff
@@ -416,38 +407,33 @@ public class AppCenter.Homepage : Gtk.Box {
...
             };
-            box.add (display_image);
-            box.add (name_label);
 
             content_area.attach (box, 0, 0);
 
             if (category.icon != "") {
-                display_image.icon_name = category.icon;
+                var display_image = new Gtk.Image.from_icon_name (category.icon, Gtk.IconSize.DIALOG) {
+                     halign = Gtk.Align.END,
```
jeremypw, line 424: Incorrect indent.

jeremypw (commented):
No regression in appearance.  Just a couple of minor comments.

data/styles/categories.css
danrabbit: Yeah it works for margins and padding, but when used as stops in gradients we get weird rounding sometimes that makes gradients either too hard or lines disappear etc

danrabbit:
After thinking about it for a second, we can use `calc` to avoid rounding errors and keep the background image scalable :tada:

jeremypw (approved):

PR #1908, opened 2022-09-02
CategoryView: change append to add_package

danrabbit:
Want to make sure we don't have to override `append` actually because we need to use it during `construct`. This simplifies things anyways because we don't need to construct the package row in `populate` because it's only used in the flowbox


jeremypw (approved):
Functionally equivalent

PR #1909, opened 2022-09-02
AbstractAppList: Remove list.add connect

danrabbit:
Can't connect to `add` in Gtk 4, so just make sure we do these things when we add


jeremypw (approved):
Functionally equivalent and all uses covered.

PR #1910, opened 2022-09-02
Move remove_package from AbstractAppList to AppListUpdateView

danrabbit:
Only used here, so remove it from the abstract class


jeremypw (commented):
Functionally equivalent.

jeremypw (approved):

PR #1911, opened 2022-09-02
AbstractAppList: Remove sort func

danrabbit:
Neither implementation uses the abstract sort function, so remove it

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -37,6 +37,7 @@ namespace AppCenter.Views {
             );
             loading_view.show_all ();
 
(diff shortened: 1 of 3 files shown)


jeremypw (approved):
Appears functionally equivalent. 

I noticed the AppListUpdateView does not sort the already up to date packages in name order (the order seems to change every time depending on when the view is opened)  However a similar issue appears in master so is not blocking for this PR.

jeremypw:
The SearchView does appear in name order though.

danrabbit:
Yeah the sort func in the updates view is pretty complicated and I'm assuming broken

PR #1912, opened 2022-09-02
Move get_packages from AbstractAppList to AppListUpdateView

danrabbit:
Only used in one implementation so move it from the abstract list. We have a null check with a continue here so I'm assuming this type of cast is intentional, maybe because row headers are listbox children?

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -329,12 +329,9 @@ namespace AppCenter.Views {
         private Gee.Collection<AppCenterCore.Package> get_packages () {
             var tree_set = new Gee.TreeSet<AppCenterCore.Package> ();
             foreach (unowned var child in list_box.get_children ()) {
-                unowned var row = child as Widgets.PackageRow;
-                if (row == null) {
-                    continue;
+                if (child is Widgets.PackageRow) {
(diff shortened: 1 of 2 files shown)


src/Views/AppListUpdateView.vala
```diff
@@ -326,6 +326,20 @@ namespace AppCenter.Views {
             updating_all_apps = false;
         }
 
+        private Gee.Collection<AppCenterCore.Package> get_packages () {
+            var tree_set = new Gee.TreeSet<AppCenterCore.Package> ();
+            foreach (unowned var child in list_box.get_children ()) {
+                unowned var row = child as Widgets.PackageRow;
+                if (row == null) {
+                    continue;
+                }
```
jeremypw, line 335: Maybe better to replace this with 
```
if (!child is Widgets.PackageRow) {
    continue;
}

row = (Widgets.PackageRow)child;
```

danrabbit: Yeah that's much more explicit. Thanks!

jeremypw (changes requested):
Functionally equivalent but see code style suggestion.

jeremypw (approved):

PR #1914, opened 2022-09-02
Handle more componentkinds

jeremypw:
In connection with https://github.com/elementary/icons/pull/1128, try to reduce need for a completely generic app icon, which is hard to represent, by using generic icons for categories of apps/packages.

Most apps in the AppCenter do supply their own icons.  The fallback generic icon is mainly used for system components, most of which are identified as a "generic" appstream component kind.  Presumably this is due to lack of information in the appdata?
A couple of component kinds have been given example icons but these are not ideal.

Another place the generic application icon is used extensively is when Files displays the contents of e.g. `usr/bin` for executables.  In that case it would be better to have an icon for a generic executable as these are not necessarily (desktop) apps.

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -724,10 +724,28 @@ public class AppCenterCore.Package : Object {
         }
 
         if (icon == null) {
-            if (component.get_kind () == AppStream.ComponentKind.ADDON) {
-                icon = new ThemedIcon ("extension");
-            } else {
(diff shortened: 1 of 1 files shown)


jeremypw:
@danrabbit Not sure how useful this is - your input welcome.

danrabbit (approved):
This looks good to me and makes it way clearer which component kinds we're covering. Nice job!

PR #1915, opened 2022-09-02
AbstractAppList: Remove redundant package_changing

danrabbit:
It looks like this only thing this does is call `invalidate_sort ()`, but we already send the `changed ()` signal in `PackageRow` which automatically calls `invalidate_sort ()` on the ListBox, so it doesn't seem like there's a need to call it twice.

I tested installing a package and manually checking for updates and afaict the listbox sorts when packages change without this


jeremypw (approved):
I guess if you have the list open and then update a package externally the sort may not update but this does not seem to very likely to occur in practice.

PR #1916, opened 2022-09-02
Remove AbstractAppList

danrabbit:
After:
- [x] #1912
- [x] #1915

With these views now no longer sharing a common class we can have more freedom with the layout (such as showing installed apps with no updates in a flowbox maybe?). Also we can see how small this class is that it doesn't really do any major heavy lifting to abstract here


jeremypw (dismissed):
Makes sense as these classes diverge.  No regressions found.

jeremypw:
Sorry - accidentally reviewed while still draft.

jeremypw (approved):

Issue #262, opened 2017-05-10
Toasts


Timecraft:
With the launch of the new version of Granite, I think it would be a great idea to add in a Toast widget to let the user know when an app is finished updating/downloading.

harisvsulaiman:
Isn't that what notifications are for?

danrabbit:
Notifications are sent if the window is unfocused but there's currently no feedback if the window is focused but the updates view is not

harisvsulaiman:
@danrabbit Didn't know that, in that case toasts on operations-complete is a welcome feature.

PR #1918, opened 2022-09-03
MainWindow: don't rely selected_package

danrabbit:
This isn't a reliable way to make sure we're not on the appinfoview of an app being installed. Instead, we can just get the currently visible child of the deck and see if we're looking at the package


jeremypw (approved):
Code looks good works as expected.

I would maybe expect a similar notification on uninstalling but that is out of scope for this PR.

danrabbit:
@jeremypw thats a good suggestion! Can you open an issue report?

jeremypw:
@danrabbit It is more or less covered by issue #262 I think

PR #1919, opened 2022-09-03
Show current job type in overlaybar

### LINKED ISSUE #394: Add tooltip to loading spinner
It might not be clear what exactly is loading when the spinner in the headerbar is going. We could reveal some more information as a tooltip for those curious.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47102514-add-tooltip-to-loading-spinner?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #394 

Screenshot from 2022-09-03 12 24 34

* Changes the spinner to an overlaybar, which saves us some horizontal space in the headerbar and keeps elements from moving around.
* Show a label describing the current job type in the overlaybar


jeremypw (approved):
Code looks good and works as expected.

I noticed that the spinner periodically stalled apparently indefinitely but would restart if the overlay was hovered to make it move to the other side.  This likely to be a Gala or other system issue as the spinner in the overlay is created spinning and never stopped (just hidden) but it is annoying and potentially confusing.

It would be nice to show more information like which package is being updated so that the text changed over time to give assurance that something is actually happening, but that can wait for another PR.

danrabbit:
Yeah we definitely have some issues with the whole UI freezing while updates are being fetched and I'm hoping that's the root cause of that

jeremypw:
Yes, it is not surprising it freezes occasionally when the cpu is overloaded but I would expect it to restart when the cpu load decreases. It seems not to restart until the overlay is moved.

PR #1920, opened 2022-09-04
AppListUpdateView: use flowboxes

danirabbit:
Use flowboxes instead of a list to show updates and installed apps.
* Simplify sort and header handling
* Show more apps on screen at once
* Easier to see which info and update button goes to which app

## BEFORE
Screenshot from 2022-09-04 13 14 58

## AFTER
Screenshot from 2022-09-04 13 13 12


danrabbit:
Haha yeah I did this and was like hmmm maybe need to start with just the header 😅

jeremypw:
I look forward to it being completed as it makes much better use of space.

PR #1921, opened 2022-09-05
AppListUpdateView: make update all header sticky

danrabbit:
Don't create the whole updates header over and over again, just create it once and put it in a revealer. Also use the update count information from Client and calculate size when this changes instead of in the header func. Give it a cool sticky behavior.

Side note, in a future branch we should probably calculate size, unpaid apps, flatpak usage, etc while we're counting updates so this page can grab that info from Client too.

## Not scrolled
Screenshot from 2022-09-05 14 48 32

## Scrolled
Screenshot from 2022-09-05 14 48 43


src/Views/AppListUpdateView.vala
```diff
@@ -95,11 +125,52 @@ namespace AppCenter.Views {
...
+
+                            if (!using_flatpak && package.is_flatpak) {
+                                using_flatpak = true;
+                            }
+
+                            update_real_size += package.change_information.size;
+                        }
+                    }
+
+                    size_label.update (update_real_size, using_flatpak);
```
jeremypw, line 149: I think you also need to update the size label in the constructor else it may not show if the view is opened after checking for update has finished.

jeremypw (changes requested):

jeremypw (approved):
Works as expected, code looks good.

Nice improvement!

PR #1922, opened 2022-09-06
UpdateManager: add update-all action

### LINKED ISSUE #1545: Add "Update All" button to update notification
It'd be slick to be able to kick off installing updates from the notification without opening the AppCenter UI itself.

danirabbit:
Fixes #1545 

Move the "Update All" func into UpdateManager and create a GLib.Action that triggers it. It looks like the notification button isn't being added when a default action is set. Not sure if this is a bug in our notifications server or a "Feature" in GLib


danrabbit:
I'm gonna convert this back to a draft for a second because there's also some stuff about like setting rows insensitive etc that could probably be simplified

PR #1924, opened 2022-09-08
UpdateManager: count updates size, flatpak, unpaid

danirabbit:
Reduce the amount of times we iterate over lists in AppListUpdateView

### DIFF
--- a/src/Core/UpdateManager.vala
+++ b/src/Core/UpdateManager.vala
@@ -20,6 +20,9 @@
 public class AppCenterCore.UpdateManager : Object {
     public bool restart_required { public get; private set; default = false; }
     public Package os_updates { public get; private set; }
+    public int unpaid_apps_number { get; private set; default = 0; }
+    public uint64 updates_size { get; private set; default = 0ULL; }
+    public bool has_flatpak_updates { get; private set; default = false; }
 
     private const string RESTART_REQUIRED_FILE = "/var/run/reboot-required";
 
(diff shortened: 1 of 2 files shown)


jeremypw (commented):
Couple of issues noticed so far:
* Started with some System Updates + some "normal" updates.  Size label indicated "3.4GB".
* Debug messages indicated that the system updates contained flatpaks (runtimes & locales) but the size label did not say "Up to".
* Installed the System Updates while monitoring network traffic.
* Only about 300MB was downloaded
* After the system updates completed, the size label disappeared even though there was still another 7 updates to perform (none of them flatpaks)

jeremypw:
It appears that the size label disappears because the `change_information.size` property is returning `0` for the updating packages, so this not down to this PR.

danrabbit:
@jeremypw okay looking at this again after taking a moment away I realized I was doing some extra stuff here!

Now I'm waiting again for `installed_apps_changed`, which means all of this can go in `get_apps` and should mean we only update the header once we get all of the completed relevant information and not while we're still counting.

I can confirm the `Up to` label when including flatpaks is now present

Downloading a lot less data than the maximum size is expected with Flatpak. There's no way I know of to guess a closer size.

We should now only update the size label when we're ready to and in tandem with revealing the header so that should be fixed

jeremypw:
When I first ran this it began installing updates automatically.  I cancelled installing the non-system updates and ended up with this:

Screenshot from 2022-09-22 11 15 35


After turning off automatic updates and refreshing with `Ctrl+R` there were 26 updates outstanding including system updates.

So far the PR seems to work OK with automatic updates off.

jeremypw:
I'll run this PR for a while under normal operating conditions (as a daemon) to see whether the above issue recurs or was just an artifact.

danirabbit:
@jeremypw any additional feedback? The situation in master is pretty ugly with #1925. I believe this branch resolves it

jeremypw:
OK, lets merge - I do not think there are any serious issues.

jeremypw (approved):

PR #1926, opened 2022-09-14
 AppListUpdateView : Fix visibility and content of size label

### LINKED ISSUE #1925: Update size may not appear or may be incorrect
### What Happened?

This was mainly noticed when restarting AppCenter after installing from source.  Although the updateable packages indicated remained the same, the download size label sometimes did not appear and sometimes showed different values

### Steps to Reproduce

* Keep restarting AppCenter (after killing the background process) 
* Switch to Update list view at different stages during the refresh process
* Note the presence and text of the size label

### Expected Behavior

The size label should always appear when there are updates and the value displayed should be correct.

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1925 

Only update the header *after* updating the app list.
Update the app list and header together when the `installed-apps-changed` signal received.

Since updates are refreshed whenever the AppCenter window is opened there is no need to update/show the header in advance of the `installed-apps-changed` being received, which takes longer than constructing the view.

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -118,19 +118,13 @@ namespace AppCenter.Views {
             add (header_revealer);
             add (scrolled);
 
-            get_apps.begin ();
-
             unowned var client = AppCenterCore.Client.get_default ();
             client.installed_apps_changed.connect (() => {
-                Idle.add (() => {
-                    get_apps.begin ();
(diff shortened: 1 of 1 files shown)


jeremypw:
Closing in order to rebase locally.

PR #1933, opened 2022-09-25
ReleaseRow: cleanup and design update

danirabbit:
* Code style
* DRY and simplify where possible
* Use relative date time for release dates
* Use separate labels so we can use css to style
* Add an icon

## BEFORE
Screenshot from 2022-09-25 13 51 45

## AFTER
Screenshot from 2022-09-25 13 47 48


JoseExposito (approved):
LGTM, thanks!

PR #1935, opened 2022-09-30
Translate "Updates & installed apps"

lenemter:


tintou:
The `C_` macro is a translation marker with a context, the string should already be available or it is a bug

lenemter:
Oh I see. Translation files haven't been updated in a while, so this string is not on Weblate

danirabbit:
I'll regenerate pot files today! Hadn't done it since we were landing a lot of UI changes

PR #1936, opened 2022-10-01
fix: Segfault from error message being freed in dialogs

mmstick:
These error dialogs will cause the appcenter to crash from a memory safety violation if there's a dependency conflict or network error. It seems that `show_error_details (error.message)` is consistently given an error whose `message` has already been freed. This change will take the message string into the object initialization to ensure the string survives. Dialogs are only using the string anyway.

Can be easily replicated by editing the `/etc/hosts` file and setting the apt repository to localhost.


jeremypw (changes requested):
Makes sense. Its similar to #1877 which fixed a different dialog. 
I would suggest changing the property name in the dialog from `error` to `error_message` as the other PR did.

jeremypw (approved):

PR #1943, opened 2022-10-31
Install PackageKit based Operating System Updates as Offline Updates

meisenzahl:


danirabbit:
Alright, so I thinking about this, probably the smallest change to the UI would be to add another header that's like "Waiting to Install" or something and list the apps there that have been downloaded but not installed yet and on the header have a button "Restart and Install" similar to how we have the "Update All" button.

And then I think we probably need to change the labels on the buttons for non-flatpak packages from "Update" to "Download" so that it's more clear which ones will be installed immediately and which ones will be installed after restarting.

And then from there I think we can iterate on doing a more complicated redesign after the feature lands, but that sounds to me like the smallest change to make it usable. What do you think?

meisenzahl:
All packages provided by PackageKit are grouped as "Operating System Updates". The button label is changed to "Download". If the packages are downloaded and prepared for update on next boot by PackageKit a info is shown that "A restart is required to finish installing updates". This was accomplished by extending existing logic.

https://user-images.githubusercontent.com/10796736/208926905-38e134de-108a-4baf-a88d-d97e2f864554.mp4

After you prepared "Operating System Updates", closed and opened AppCenter again, the info is shown again that a restart is required.

If you restart your device the packages will be installed during boot like shown here by the early Operating Upgrade Protoype: https://youtu.be/RRN05MwFJ5Y?t=415

meisenzahl:
@danirabbit I would love to get feedback from you.

A few more things you could look out for in the review:

1. when applying the updates during the boot process, a progress bar with a percentage value is shown on my device, but no status text in the middle of the display. This worked under OS 6 on my device. (see the video of the upgrade prototype) Is this also the case for you? If so, we may need to update/patch Plymouth.

2. `BackendAggregator.get_prepared_applications ()` returns a list of prepared packages. Currently, we do not use the list. Would you like to show more information about the prepared packages with a follow-up PR? If not, I could probably remove the unused functions.

danirabbit:
As far as I can tell this worked exactly as expected, nice job!

I can confirm that there's no label like "Installing Updates" on the plymouth screen. It looks like the patch we used in Focal was built for Jammy, so I can make sure that gets done: https://github.com/elementary/os-patches/issues/240

I think in a future branch we should probably rework that OS Updates page not to use an AppInfoView so we can do something more appropriate like GNOME Software with the list of additions/removals/upgrades, but I agree that can wait for another branch

meisenzahl:
https://github.com/elementary/appcenter/pull/1943/commits/94e23271c66134e0e80bcf8eef0c79e1c060c908 removes code that we currently don't use. But it could be useful for a follow-up PR.

meisenzahl:
My bad, so actually we need the changes I tried to remove in https://github.com/elementary/appcenter/pull/1943/commits/94e23271c66134e0e80bcf8eef0c79e1c060c908.

These are responsible for checking to see if packages are ready for an update, if they are, it will indicate that a reboot is necessary.

Currently we do not use the returned list of packages. However, it should be very useful for follow-up PRs.

danirabbit:
I'm building Plymouth now, it looks like we carried a patch last release to set fonts, so I bet the text didn't show because we're missing the Ubuntu font: https://code.launchpad.net/~elementary-os/+recipe/plymouth-jammy

danirabbit (changes requested):
One issue I discovered here is that if I click "Download" and let that finish, then either close AppCenter and reopen it, or click the "Check for Updates" menuitem, the Operating system updates row will appear again with the download button and the Restart infobar will disappear even though we've already downloaded those updates

meisenzahl:
@danirabbit thanks for your review.

PackageKit deletes all information about offline update actions and prepared updates (stored in `/var/lib/PackageKit/`) if a refresh of it's cache is triggered. https://github.com/elementary/appcenter/pull/1943/commits/6d4065d036716a310d61c6e9dc251b2ba0ec8d61 prevents a refresh of the cache if prepared updates are available.

## "Download" ⟶ "Check for Updates"

https://user-images.githubusercontent.com/10796736/209296838-e07bb2de-4285-4c65-a923-94a3c6969fb8.mp4

## "Download" ⟶ Reopen AppCenter

https://user-images.githubusercontent.com/10796736/209296844-91b8aab4-5437-4d44-8e61-672b05745bf1.mp4

danirabbit (approved):
Nice work! I say we go ahead and get this in so we can keep iterating on it and get wider testing etc :)

PR #1956, opened 2022-12-20
Flatpak Runtimes

meisenzahl:
In preparation for https://github.com/elementary/appcenter/pull/1943 and potentially for https://github.com/elementary/os/pull/582, I propose that we remove updates for Flatpak Runtimes from the "Operating System Updates" category and group them in the new "Runtime Updates" category.

This would have the advantage that "Operating System Updates" would only be managed by PackageKit. This would simplify handling and presentation for https://github.com/elementary/appcenter/pull/1943.

| Before  | After  |
|:---:|:---:|
|   |  |
|  |  |

To test the changes, you can downgrade `io.elementary.Platform/x86_64/7.1`:

```bash
sudo flatpak update \
    --commit=995cc9d6c6b1ea57a4869bc4e59a7d325750899a24a1dc1043438d7e840ca466 \
    io.elementary.Platform/x86_64/7.1
```

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -83,6 +83,7 @@ public class AppCenterCore.Package : Object {
     }
 
     public const string OS_UPDATES_ID = "xxx-os-updates";
+    public const string RUNTIME_UPDATES_ID = "xxx-runtime-updates";
     public const string LOCAL_ID_SUFFIX = ".appcenter-local";
     public const string DEFAULT_PRICE_DOLLARS = "1";
 
@@ -102,7 +103,7 @@ public class AppCenterCore.Package : Object {
     private bool _installed = false;
     public bool installed {
         get {
-            if (component.get_id () == OS_UPDATES_ID) {
+            if (is_os_updates || is_runtime_updates) {
                 return true;
             }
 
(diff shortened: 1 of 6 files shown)


meisenzahl:
`application-x-sharedlib` probably is a better choice:

meisenzahl:
I was not aware of `application-vnd.flatpak`:

danirabbit:
Makes sense to me! I'm +1 on the Flatpak bundle icon. Nice work :)

danirabbit:
Actually thinking about it, since there's unlikely to be more than a handful of runtime updates, do we really need to group them? I think the original idea behind grouping OS Updates was because there could easily be like a dozen or more of them at a time and they're all very cryptic and don't have appdata etc. but Flatpak runtimes should have appstream data and you probably will only ever see a couple of them at a time

meisenzahl:
@danirabbit I will prepare an alternative PR without grouping.

Marukesu:
> Actually thinking about it, since there's unlikely to be more than a handful of runtime updates, do we really need to group them? I think the original idea behind grouping OS Updates was because there could easily be like a dozen or more of them at a time and they're all very cryptic and don't have appdata etc. but Flatpak runtimes should have appstream data and you probably will only ever see a couple of them at a time

Flatpak extensions are runtimes too, and most of, if not all, don't ship with metainfo. we should be sure that we can export metainfo files for at least `.Locale` and `.Debug` extensions (since those are generated automatically by flatpak-bulder, instead of by alternatives manifest) and recommend that for third-party applications.

however, metainfos aren't required for runtimes and extensions in flatpak-builder, external repos still can have them and we would need to have a way to show them here as we have with apt packages.

danirabbit (approved):
In that case, let's go ahead with this branch so we can unblock @meisenzahl and we can always revisit it :)

Issue #1959, opened 2022-12-27
OS Updates item shows when there's no updates available


danirabbit:
### What Happened?

![Screenshot from 2022-12-27 15 30 04](https://user-images.githubusercontent.com/7277719/209736247-03393338-e51c-49c9-afa0-265cb4f2d739.png)

### Steps to Reproduce

Install all your updates and then view the updates page :)

### Expected Behavior

Seems kinda weird and unnecessary to show this item if there's nothing in it. It doesn't display any useful information. Probably better to skip adding it if there's nothing to show

### OS Version

7.x (Early Access)

### Software Version

Latest release (I have run all updates)

PR #1960, opened 2022-12-28
Show Operating System Updates and Runtime Updates only if available

meisenzahl:
Fixes https://github.com/elementary/appcenter/issues/1959


danirabbit (approved):
Works as expected. Nice job!

Issue #1310, opened 2020-05-12
Get Flatpak remote names from Title field


isantop:
## Prerequisites
- [*] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**

AppCenter displays the name of the Flatpak remote in the UI when viewing a Flatpak App. It's getting this from the name in the config, which is a cli-friendly (i.e. lowercase, no spaces) name of the remote. Flatpak also provides the `xa.title` field which can get a UI-Friendly display title of a remote, which would be better than using the name as it's intended for UI display to users.

**Describe the solution you'd like**

Use the Title field to display remote names in AppCenter

**Existing work**

https://github.com/pop-os/repoman uses Titles for displaying remotes to users (and falls back on the name when a title is unavailable, as it's an optional field):

![image](https://user-images.githubusercontent.com/5883565/81724684-07553b00-9442-11ea-80c3-bd76dcb7c746.png)

`libflatpak` provides the `FlatpakRemote.get_title()` method to get the title if one is available.

**Describe alternatives you've considered**

It looks like GNOME Software hardcodes UI-Friendly names in the code (e.g. `if remote_name == 'flathub': remote_name = 'Flathub'`; this works, but we probably want to avoid doing that except in very extreme circumstances.

**Additional context**

[Flatpak remote configuration docs](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-remote)
[`libflatpak` FlatpakRemote docs](https://docs.flatpak.org/en/latest/libflatpak-api-reference.html#flatpak-remote-get-title)

PR #1962, opened 2022-12-29
Pretty output Flatpak package origins

meisenzahl:
Closes https://github.com/elementary/appcenter/issues/1310


danirabbit:
Did we already try getting the title as in #1310? It sounds like there's a user-facing remote title string we should be using that we don't have to hard code

meisenzahl:
@danirabbit thanks for mentioning https://github.com/elementary/appcenter/issues/1310 :)

danirabbit (approved):
Nice job! Works as expected here. Fallbacks look good. And no hardcoding :)

PR #1965, opened 2023-01-05
Improve SearchView

meisenzahl:
Only show the Flathub link if the Flathub remote is enabled.


davidmhewitt:
Why? I thought the point was that you could go to flathub in a browser and install an app, and doing so would add the flathub remote.

davidmhewitt:
Ah, my bad. I misunderstood your description. This only shows the link if Flathub *is not* enabled, right?

meisenzahl:
@davidmhewitt my bad I messed up the description.

Of course, this only shows the link if Flathub is not enabled.

PR #1966, opened 2023-01-08
actions/checkout@v3

meisenzahl:


danirabbit (approved):
I think the gettext action files to push with v3, but let's try it

PR #1974, opened 2023-01-26
AppListUpdateView: use flowbox for installed

### LINKED ISSUE #1522: Multiple update headers can appear
## What Happened

I just saw an instance where I had updates and up-to-date apps split among a bunch of different headers and mixed together. I saw a structure that was something like
```
22 updates available
    *a few apps listed*
Up to Date
    *a few apps listed*
22 updates available
    *more apps*
Up to Date
and so on...
```
None of these lists contained the full collection of up-to-date or update-able apps, so I think each app just appeared once. I didn't get a chance to check though.
(Sorry, I didn't get a screenshot before the updates started finishing)

## Expected Behavior

All available updates should be at the top, followed by all up to date apps.

## Steps to Reproduce

Not sure, to be honest... Maybe it'll appear if you just have enough updates?

**Speculation:** My guess is that this is a result of some bug in the ListBox sort function. I'm guessing that it sorts the up-to-date and update-able apps together. Then when each is assigned a header it results in the headers being scattered around, making duplicate headers. In this case, it should just appear if you have enough updates at once. (I had waited for a week or 2, so most of the apps on my system had updates available. That might have helped.)

### Logs

n/a

## Platform Information

Up to date as of 4/21/21 in Odin daily

danirabbit:
Screenshot from 2023-01-26 15 47 38

Closes https://github.com/elementary/appcenter/discussions/1658
Fixes #1522
* Much better use of space for installed components on large displays


lenemter (changes requested):
When opening an app and immediately going to updates i get this:
image

No "Checking for Updates" placeholder, only "Up to Date" header at the very bottom

---

`master`:
image

danirabbit:
@lenemter re-added the placeholder! Sorry it took so long to follow up

lenemter (approved):
Looks good

Issue #1976, opened 2023-01-28
"May Not Be Translated" shows on OS Updates page


lenemter:
### What Happened?

![image](https://user-images.githubusercontent.com/80143868/215261559-4b7507ae-e66f-4d39-9c72-e183d048b237.png)

### Steps to Reproduce

1. Open OS Updates page

### Expected Behavior

No warning

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

PR #1977, opened 2023-01-28
Fix warnings for native packages, OS and Runtime updates

meisenzahl:
Closes https://github.com/elementary/appcenter/issues/1976

### Before

Bildschirmfoto von 2023-01-28 13 04 24
Bildschirmfoto von 2023-01-28 13 04 32

### After

Bildschirmfoto von 2023-01-28 13 05 50
Bildschirmfoto von 2023-01-28 13 05 57

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -181,36 +181,38 @@ namespace AppCenter.Views {
             }
 #endif
 
-            var percent_translated = package_component.get_language (
-                // Expects language without locale
-                package_component.get_active_locale ().split ("_")[0]
-            );
+            if (!package.is_native && !package.is_os_updates && !package.is_runtime_updates) {
+                var percent_translated = package_component.get_language (
+                    // Expects language without locale
+                    package_component.get_active_locale ().split ("_")[0]
+                );
 
-            if (percent_translated < 100) {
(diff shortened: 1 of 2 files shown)


meisenzahl:
@elementary/ux do we want to filter more info for native packages, OS and Runtime updates like ratings?

meisenzahl:
The last commit changes the stable package origin to `jammy`:

### Before
Bildschirmfoto von 2023-01-28 13 11 48

### After
Bildschirmfoto von 2023-01-28 13 14 53

src/Views/AppInfoView.vala
```diff
@@ -181,36 +181,38 @@ namespace AppCenter.Views {
...
-                    );
-
-                    oars_flowbox.add (locale);
-                } else if (percent_translated == 0) {
-                    var locale = new ContentType (
-                        _("Not Translated"),
-                        _("This app is not available in your language"),
-                        "metainfo-locale"
-                    );
+            if (!package.is_native && !package.is_os_updates && !package.is_runtime_updates) {
```
danirabbit, line 184: I think filtering out OS and runtime updates pages makes sense while those are still using AppInfoView, but I don't think we should exempt our own packages from these warnings

danirabbit (approved):

Issue #1981, opened 2023-02-01
Runtime Updates don't seem to update


guythomson:
### What Happened?

Ran updates and when I returned to the Appcenter I saw

> Runtime Updates
> No runtimes with updates

with an update button:

![Screenshot from 2023-02-01 17 28 44](https://user-images.githubusercontent.com/99534283/216122173-0812b92d-9106-4ba0-8eff-ffd14b66464f.png)

![Screenshot from 2023-02-01 17 28 49](https://user-images.githubusercontent.com/99534283/216123484-fe585f92-e1db-4b10-8e63-8df936f8654c.png)

Pressing the Update button seems to run something but then the row remains to be updated again.

Ran

> sudo apt update && sudo apt upgrade

Reopened Appcenter and problem still exists.

### Steps to Reproduce

- Opened Appcenter
- Click the update cog
- Press update

### Expected Behavior

- Click update
- Updates happen
- Update disappears and reappears as an installed component

### OS Version

7.0 (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell
### Loading

(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.620: Received 1406 bytes
(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.624: Loaded indexed summary file 2551be2af70183606080438223db200bd2d301264419fc9b5f0c5051dddcbd5f from cache for remote ‘appcenter’
(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.624: Transaction: update appcenter:app/io.elementary.calculator/x86_64/stable[*]
(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.624: Transaction: update appcenter:app/io.elementary.camera/x86_64/stable[*]
(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.625: Transaction: update appcenter:app/io.elementary.capnet-assist/x86_64/stable[*]
(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.625: Transaction: update appcenter:app/io.elementary.music/x86_64/stable[*]
(io.elementary.appcenter:8812): flatpak-DEBUG: 18:01:00.625: Transaction: update appcenter:app/io.elementary.screenshot/x86_64/stable[*]
[...]

Issue #1741, opened 2021-11-01
updates are not installing 


Taasin:
### What Happened?

2 components with updates:
 • org.freedesktop.Platform Flatpak runtime
 • org.freedesktop.Sdk Flatpak runtime

this two updates are not installing in my elementary os. i tried to reintall but it failed everytime, 

### Steps to Reproduce

2 components with updates:
 • org.freedesktop.Platform Flatpak runtime
 • org.freedesktop.Sdk Flatpak runtime

### Expected Behavior

2 components with updates:
 • org.freedesktop.Platform Flatpak runtime
 • org.freedesktop.Sdk Flatpak runtime

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

Issue #1523, opened 2021-04-23
Attempt to recover from Flatpak issues


cassidyjames:
## Problem

If Flatpak gets into a state where it needs to be repaired (e.g. if an install was interrupted), AppCenter fails to install or update Flatpaks until `flatpak repair` is run (mentioned in #1234 and https://github.com/pop-os/shop/issues/248#issuecomment-778692701)

## Proposal

AppCenter should attempt to recover from Flatpak issues by automatically repairing.

## Prior Art

I'm not sure if GNOME Software or KDE Discover do anything special in this situation.

davidmhewitt:
There aren't any repair methods exposed in the public libflatpak API so achieving this would likely be a case of asking if they can be (or why they shouldn't be), or running `flatpak repair` as a subprocess (which wouldn't really be the preferred solution)

cassidyjames:
@davidmhewitt ah, that makes sense. Yeah I think it's worth asking Flatpak devs if they can/should be exposed, as I understand our current options are to silently error (like now), throw a dialog asking the user to run Terminal commands (eww), or spawning a subprocess (also eww).

hanaral:
I think `flatpak repair` should have an indication on whether it's running or not since it disables your ability to open any apps (and can also take a very long time if you have lots of app and/or are on an HDD)

danirabbit:
I think this should be considered fixed with #1983 since we now offer a Flatpak repair option as a subprocess

PR #1983, opened 2023-02-02
Flatpak repair

meisenzahl:
Fixes https://github.com/elementary/appcenter/issues/1981
Fixes https://github.com/elementary/flatpak-platform/issues/108
Fixes https://github.com/elementary/appcenter/issues/1741
Fixes https://github.com/elementary/appcenter/issues/1986

As  @mwleeds mentioned in https://github.com/elementary/appcenter/issues/1523#issuecomment-1048005581 we should patch

- https://github.com/flatpak/flatpak/pull/4765
- https://github.com/ostreedev/ostree/pull/2549

in https://github.com/elementary/os-patches.

The files provided in this PR could be a solution to repair broken Flatpak installations.


meisenzahl:
I was about to create patches for Flatpak and OSTree in https://github.com/elementary/os-patches. But apparently, the fixes are already included in the default Ubuntu packages:

https://github.com/elementary/os-patches/blob/flatpak-jammy/common/flatpak-dir.c#L5948

https://github.com/elementary/os-patches/blob/ostree-jammy/src/libostree/ostree-repo-pull.c#L2247

@mwleeds do you have any other idea what could be the reason that Flatpak installation breaks for users?

For example in https://github.com/elementary/appcenter/issues/1981

meisenzahl:
The last commit adds a quick working prototype to fix Flatpak installations. It runs `flatpak --system repair` and `flatpak --user repair` in the background.

The functionality is implemented via an `Action`. We could add a button in case of an error during an installation or update that suggests the repair and triggers the action.

https://user-images.githubusercontent.com/10796736/216595280-1b58d354-973b-4ee2-a17a-a515cf6e1414.mp4

@elementary/ux would this be a solution we want to work on further?

src/Core/FlatpakBackend.vala
```diff
@@ -1690,6 +1693,64 @@ public class AppCenterCore.FlatpakBackend : Backend, Object {
...
+        job.results_ready ();
+    }
+
+    public async bool repair (Cancellable? cancellable = null) throws GLib.Error {
+        var job_args = new RepairArgs ();
+        job_args.cancellable = cancellable;
+
+        var job = yield launch_job (Job.Type.REPAIR, job_args);
+        if (job.error != null) {
+            warning ("Could not repair user Flatpak: %s", job.error.message);
```
danirabbit, line 1747: We should probably not silently fail and make sure we throw an error dialog if we're unable to repair

src/MainWindow.vala
```diff
@@ -220,12 +225,24 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
...
+        repair_menuitem.get_child ().destroy ();
+        repair_menuitem.add (repair_accellabel);
+
         var menu_popover_box = new Gtk.Box (Gtk.Orientation.VERTICAL, 0) {
             margin_bottom = 6,
             margin_top = 6
         };
         menu_popover_box.add (automatic_updates_button);
         menu_popover_box.add (refresh_menuitem);
+        menu_popover_box.add (repair_menuitem);
```
danirabbit, line 245: I'm not sure this makes sense to show preemptively. If we give the option for repairing when there's an error, it kind of seems like this menu item wouldn't really be used or it would be confusing why it's there and how to make it go away

meisenzahl:
https://user-images.githubusercontent.com/10796736/217006730-a49566f3-797f-47e5-96a3-ac0d574c4a14.mp4

meisenzahl:
@danirabbit  after a fresh installation of elementaryos-7.0-stable.20230129rc.iso in VirtualBox, I also have the runtime error you see in the video. With the steps in the video I am able to fix the error using this branch.

danirabbit (approved):
I can confirm that this fixes the issue in a clean virtual machine, so let's goo :rocket:

PR #1994, opened 2023-02-07
Add meson options to disable backends

meisenzahl:
Closes https://github.com/elementary/appcenter/issues/1076


meisenzahl:
> @sysrich had expressed interest at GUADEC in using a Flatpak-only AppCenter for a project he was working on. It could be useful to provide a flag to disable the PackageKit backend and only talk to Flatpak.

@sysrich if you're still interested. Great talk at FOSDEM by the way :)

src/Core/Package.vala
```diff
@@ -327,6 +327,15 @@ public class AppCenterCore.Package : Object {
...
+            if (backend is FlatpakBackend) {
+                var fp_package = this as FlatpakPackage;
+                if (fp_package == null) {
+                    return origin;
+                }
+
+                return fp_package.remote_title;
+            }
+#if PACKAGEKIT_BACKEND
             if (backend is PackageKitBackend) {
```
meisenzahl, line 339:
```suggestion
            else if (backend is PackageKitBackend) {
```

zeebok (changes requested):
This looks good, but there is a conflict with master to tackle first

sysrich:
I'm interested.. I've packaged gnome-software to behave like this but once this gets merged I'd see which does the job better :)

src/meson.build
```diff
@@ -76,6 +73,25 @@ if appstream.version().version_compare('>=0.16')
...
+endif
+
+if get_option('ubuntu_drivers_backend')
+    if not get_option('packagekit_backend')
+        error('Option ubuntu_drivers requires option packagekit_backend')
+    endif
+
+    appcenter_files += files('Core/UbuntuDriversBackend.vala')
+
+    args += '--define=UBUNTU_DRIVERS_BACKEND'
```
zeebok, line 92: I don't see this getting used in this PR. What is the intent with this define?

zeebok (approved):
Just the one question for clarification, otherwise it seems good

src/meson.build
meisenzahl: This is used in `src/Core/BackendAggregator.vala` to check if the backend should be added.

Issue #148, opened 2017-03-22
Add an Author page


danrabbit:
The developer/author/project group name in an app listing should be clickable and lead to a page that shows all the apps published by that author

stsdc:
![author](https://user-images.githubusercontent.com/6031763/28325495-44120a1c-6bde-11e7-931f-d4ff362c4ece.png)

danrabbit:
Something that might be better than doing a GitHub tie-in here might be to use custom appdata tags, like we do with the stripe key and accent colors. We could add a tag for an avatar URL or whatever other kinds of info would make sense here.

I was thinking as well that maybe instead of doing a dedicated author page that we could expand the "Other apps" section at the bottom of the page into more of the author section. It could be a natural place to include things like the author's contact tag. Home page could possibly be moved here as well as the donation url. Keeping as part of the AppInfo page would avoid problems with trying to resolve conflicts between the information in different packages

cassidyjames:
Updating from our call: we could totally use a Gravatar from the contact URL (and make sure to update documentation to tell developers that's where an avatar comes from).

cassidyjames:
I don't think it makes sense to move the homepage or possibly even the donate URL here because these may be app-specific; for example if I have separate pages on my website for my apps, and different donation URLs/URL parameters to track which app was being funded.

PR #1997, opened 2023-02-07
Create AuthorView

danirabbit:
Setting the foundation for #148 by making this its own class


jeremypw (approved):
Code looks good as a first step - no regressions found.

PR #2006, opened 2023-02-09
Update POTFILES

ryonakano:
The newly added file in #1983


meisenzahl (approved):
I guess I forgot about that. Thanks a lot :)

PR #2007, opened 2023-02-09
Show Flatpak permissions

### LINKED ISSUE #1012: Display Permissions Requested by Flatpak Apps
It would be good to provide a list of permissions requested by Flatpak Apps, now that that support is present. There's almost certainly some design needed to ensure this looks good and works well from a UI and UX standpoint. But providing this information is necessary for users to know what a specific application has access to and whether they're okay with the security implications of that app. 

I believe this functionality already exists in GNOME Software and is provided by libflatpak, so we should be able to get that information for any given flatpakref fairly easily.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74817406-display-permissions-requested-by-flatpak-apps?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F57667267&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danirabbit:
Work towards #1012

This is already a good amount to review just for the backend stuff, so I want to propose this first and then build out the UI in another branch. For now, only show full filesystem access warning just to prove that it works

Screenshot from 2023-02-09 17 22 46


src/Core/Package.vala
```diff
@@ -82,6 +83,26 @@ public class AppCenterCore.Package : Object {
         REMOVING
     }
 
+    public enum PermissionsFlags {
```
davidmhewitt, line 86:
```suggestion
    [Flags]
    public enum PermissionsFlags {
```

If you do this, you can remove all of the `1 << x` parts. Probably don't need `LAST` in here either.

danirabbit: Oh that's great! Much more maintainable

src/Core/Package.vala
```diff
@@ -45,6 +45,7 @@ public class AppCenterCore.Package : Object {
     public const string APPCENTER_PACKAGE_ORIGIN = "appcenter";
     private const string ELEMENTARY_STABLE_PACKAGE_ORIGIN = "elementary-stable-jammy-main";
 
+    public PermissionsFlags permissions_flags { get; set; default = PermissionsFlags.UNKNOWN; }
```
zeebok, line 48: It isn't necessarily clear in this PR but is there a scenario were the permission *actually* is unknown? Just based on this PR it seems like it makes more sense to default to None but I feel like there very well may be unknowns I am not aware/thinking of

zeebok (approved):

PR #2024, opened 2023-05-03
AppInfoView: rework vertical spacing

danirabbit:
Meant to address some concerns about app descriptions being really far down the page. Also helps out at small window widths and just makes some of the padding and margins more consistent and pleasant, alignment of buttons etc as well

## BEFORE

Screenshot from 2023-05-02 18 19 10
Screenshot from 2023-05-02 18 18 47

## AFTER
Screenshot from 2023-05-02 18 19 48

Screenshot from 2023-05-02 18 13 10

### DIFF
--- a/data/io.elementary.appcenter.gresource.xml
+++ b/data/io.elementary.appcenter.gresource.xml
@@ -2,6 +2,7 @@
 <gresources>
   <gresource prefix="/io/elementary/appcenter">
     <file alias="application.css" compressed="true">styles/application.css</file>
+    <file alias="AppInfoView.css" compressed="true">styles/AppInfoView.css</file>
     <file alias="AppListUpdateView.css" compressed="true">styles/AppListUpdateView.css</file>
     <file alias="arrow.css" compressed="true">styles/arrow.css</file>
     <file alias="badge.css" compressed="true">styles/badge.css</file>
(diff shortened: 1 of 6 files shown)


lenemter:
I like this!
The issues I noticed:
* The spacing between app name and its description is changing when app info is loaded
image
image
* Why not make uninstall button a square, IMO it would look better (if this is simple enough to do in gtk ).

aitor-gomila:
Perhaps it could be made even tighter (when needed) by showing only the icon, the title of the app, the flatpak remote and price? All inline.

Although maybe this would overcomplicate the code :)

danirabbit:
@lenemter 
> The spacing between app name and its description is changing when app info is loaded
Yeah it's because of the height of the little info icon next to the size label, when that pops it in changes the height of that row. Probably what we can do is make sure to always show a label there with some placeholder text, but I think that might be better handled in a separate branch since this is also an issue in master

> Why not make uninstall button a square
I think this is actually pretty non-trivial in Gtk :sweat_smile: I'll give it a try haha

@aitor-gomila
> Perhaps it could be made even tighter (when needed) by showing only the icon, the title of the app, the flatpak remote and price? All inline.

I think it would make it significantly more complicated to make that work at smaller window widths. I'd like to make the package origin selector a lot more compact in a separate branch, but now the limiting factor here is actually the 128px icon that we use so there's not a significant height difference between having it and not

danirabbit:
@lenemter so I tried with Gtk.AspectFrame but it seems like I can't get the button to fill height anymore so it's square but too small. So I did a little hack in CSS and it's not perfectly square, but it's more square :sweat_smile:

lenemter (approved):
Now uninstall button looks a lot nicer, thanks! Great job :+1:

PR #2028, opened 2023-05-04
FlatpakBackend: don't purity check featured packages

### LINKED ISSUE #2026: Show recommended apps from flathub
### Problem

The feed only shows applications from appcenter, which is good. However having to search applications from flathub from outside is not very convenient.

### Proposal

Having the opportunity to search them is already good. But maybe we could show them too in the feed?  By feed I mean the welcome page that shows lots of applications

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #2026 

Allow packages from any remote to appear on the homepage


lenemter (commented):
Now no curated apps show in Recently Updated and in Carousel. Is it is supposed to be this way? I think we still need to prioritize native elementary OS apps and show them on the main page.

danirabbit:
Hm hm maybe we can bias the main carousel for AppCenter and then have the recently updated flow box unbiased?

danirabbit:
There's something funny going on with package ordering

danirabbit:
Dunno what's happening here but sorting is really fucked up and this makes it more obvious somehow

PR #2030, opened 2023-05-04
Homepage placeholder

danirabbit:
Bring back the placeholder on the homepage from before #1662 

This prevents the page contents from changing so much on startup. Also make sure we don't start the timer until we finish loading banners. Fixes some terminal warnings about switching to a non-existent widget


src/Views/Homepage.vala
```diff
@@ -271,6 +273,11 @@ public class AppCenter.Homepage : Gtk.Box {
         });
     }
 
+    private void page_changed_handler () {
+        banner_carousel.remove (appcenter_banner);
+        banner_carousel.page_changed.disconnect (page_changed_handler );
```
lenemter, line 278:
```suggestion
        banner_carousel.page_changed.disconnect (page_changed_handler);
```

lenemter (changes requested):
Works perfectly. One small nitpick in the code

lenemter (approved):

PR #2041, opened 2023-05-25
AppInfoView: warn about direct geoclue access

danirabbit:
Screenshot from 2023-05-24 17 28 20

Now that we have a location portal, shame apps for accessing geoclue directly.

While we're here, prevent a bunch of keyfile errors that caused permissions not to be displayed correctly


zeebok (approved):
Were the warnings from simply getting the list on a key that wasn't there?

danirabbit:
@zeebok yeah exactly it would throw an error and bail out early without checking any of the other keys

zeebok:
Okay, makes sense. Guess now there's a conflict with master though lol

danirabbit:
@zeebok yeah it’s just adding files to the gresource. Should be fixed :)

zeebok (approved):

PR #2039, opened 2023-05-23
AppInfoView: warn about sandbox escape

danirabbit:
Screenshot from 2023-05-24 17 32 24

View with hide whitespace changes. Same keyfile fixes as #2041
Fixes #1012

PR #2044, opened 2023-05-25
AppInfoView: Remove the curated warning

danirabbit:
Goes with:
- [x] #2039
- [x] #2040
- [x] #2041
- [x] #2043
- [ ] https://github.com/elementary/appcenter/discussions/2045

With comprehensive flatpak sandbox warnings, it's not clear what the use of this warning really is


danirabbit:
I think I'm actually going to close this. I think what we want instead is to warn about a non-elementary Platform, not a non-elementary remote

PR #2053, opened 2023-06-06
FlatpakBackend: Add null checks for user and system installation

meisenzahl:
Fixes a bug where AppCenter crashes if only an user or a system installation of Flatpak remotes is available.


davidmhewitt (approved):
Looks good, thanks!

PR #2057, opened 2023-06-27
AppInfoView: use child property instead of add

danirabbit:
Gtk 4 prep


jeremypw (approved):
Code makes sense, no regressions found.

PR #2063, opened 2023-07-27
Fix deletion of search term when replacing text with text

### LINKED ISSUE #1987: AppCenter search box dropping chatacters
### What Happened?

When performing a subsequent search, the first few characters you enter are dropped.

### Steps to Reproduce

1. Enter a search term
2. After search results have loaded, highlight the text you entered and start typing
3. The main page will briefly load
4. The search results will display
5. The first 3 characters of your search term will be missing

### Expected Behavior

The search box should not drop characters

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2059: Search bar refreshes when typing over existing text
### What Happened?

In the App Center, if you have text in the search bar, select it all and start typing a new search query, then the first few characters get removed. 

**My assumption...** 
The search bar (alongside the whole interface) in the App Center refreshes when the search bar is cleared. This makes sense if you click the icon on the right of the search bar to clear the field. Refresh the whole screen to show the default view. However if you are replacing text with new text, that refresh is delayed by some number of ms and causes the first few characters to disappear.

### Steps to Reproduce

1. Type something into the search bar
2. hit Ctrl + A to select all
3. start typing a new search query - eg "testing"
4. depending on how fast you type you'll see that 1 or more characters get cleared, leaving your search term as something like "ting"

### Expected Behavior

Should be able to overwrite text in that field and retain the full search term.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #2059 
Fixes #1987 

It changes placeholder description to "The search term must be at least 3 characters long" when searching for something and then replacing that text with something else that is < 3 characters long.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -34,7 +34,7 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
 
     private bool mimetype;
 
-    private const int VALID_QUERY_LENGTH = 3;
+    public const int VALID_QUERY_LENGTH = 3;
 
     public static Views.AppListUpdateView installed_view { get; private set; }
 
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
Fixes the related issues.

PR #2064, opened 2023-07-27
Fix deck navigation when changing package origin

### LINKED ISSUE #1947: Changing install source adds to navigation stack
### What Happened?

As I mention [in this comment](https://github.com/elementary/appcenter/pull/1896#issuecomment-1286101038) when you change the install source it seems to add to the navigation stack so going back goes to the same app for each time you select from the list.

### Steps to Reproduce

1. Go to an app that can be installed from multiple source (appcenter, system-wide, flathub, etc)
2. Select a new source any number of times
3. Click the back button

### Expected Behavior

Clicking the back button returns to the previous view such as the search or category view

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #1947

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -435,15 +435,24 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         deck.add (app_info_view);
         deck.visible_child = app_info_view;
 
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
Works as expected, no regressions found.

PR #2065, opened 2023-08-03
Meson: use gnome.post_install to compile schemas

danirabbit:


meson.build
```diff
@@ -2,7 +2,7 @@ project(
     'io.elementary.appcenter',
     'vala', 'c',
     version: '7.3.0',
-    meson_version: '>=0.56.0'
+    meson_version: '>=0.58.0'
```
bobby285271, line 5: Do you aware of anything that already requires 0.58.0 here? Otherwise `gnome.post_install()` should be a [0.57.0 feature](https://mesonbuild.com/Gnome-module.html#gnomepost_install) I think :thinking:

danirabbit: Ah for some reason I thought 0.58 😅 my bad

tintou (approved):

bobby285271 (approved):

PR #2067, opened 2023-08-05
Fix flickering on AppInfoView when getting install size

leolost2605:
Getting the install size caused the grid rows to change height and the action button to change size. 
The height change is fixed in this PR and the button size change was reduced by quite a bit. It's not quite perfect but IMHO definitely better than before.

Current master:
ac-height-master

This PR:
ac-height-new-pr


jeremypw (commented):
An alternative, simpler solution might be to not add the size label to the action stack size group.  It is this that is causing the button to change width when the size label updates. After doing this the appearance is:
Screenshot from 2023-08-07 11 52 46

jeremypw:
There is a slight sideways movement of the button when the size label updates when it is right-aligned. It would have to be centre-aligned to stop that I guess.

leolost2605:
> An alternative, simpler solution might be to not add the size label to the action stack size group. It is this that is causing the button to change width when the size label updates. After doing this the appearance is:
> Screenshot from 2023-08-07 11 52 46

If that's ok from a UX perspective it definitely would be simpler

jeremypw:
Definitely needs UX sign-off.  The problem is not knowing the final width of the size label (or even if it is visible) when the action button box is realized.  There may be a way of forcing the button box to be wider than its natural width so that it is always larger than the size label but I couldn't get that to work.

danirabbit:
Oops, I thought I left a comment here. I'm in favor of removing the label from size group as suggested by Jeremy. Would be much simpler and I'm not sure there's a great reason these need to be the same size

leolost2605:
It doesn't add the sizelabel to the size group now and uses a revealer. It also doesn't use the icon revealer now because that would cause allocation warnings and it wasn't used for a transition anyway

danirabbit (approved):
LGTM. Nice work!

PR #2068, opened 2023-08-19
StripeDialog: Gtk4 prep

danirabbit:
Partial changes from #1896

* Set pixel sizes for images
* Entries crash when their text is replaced on every keystroke, so only replace spaces if there's actually spaces
* Only convert email to lowercase when we submit it, set input hint instead
* Use `child` instead of `add` where appropriate


src/Dialogs/StripeDialog.vala
```diff
@@ -274,7 +277,10 @@ public class AppCenter.Widgets.StripeDialog : Granite.Dialog {
         });
 
         email_entry.changed.connect (() => {
-            email_entry.text = email_entry.text.replace (" ", "").down ();
+            if (" " in email_entry.text) {
+                email_entry.text = email_entry.text.replace (" ", "");
```
jeremypw, line 281: Did you intend to stop converting the email entry to lowercase?

jeremypw: Oh, I see you do that later. Ignore me.

jeremypw (commented):
This looks the same as before.

There is a slight behaviour change in that uppercase characters in the email address are not converted to lowercase on the fly (but spaces are removed on the fly).  Not sure of the reason for this change.  Invalid email addresses (e.g. multiple @ ) are not corrected or marked invalid but that is the same as master.

danirabbit:
@jeremypw I believe it was also causing crashes in Gtk4 and we really only need to change to lowercase before submitting. I don't think there's any value to changing the case while someone is typing

jeremypw (approved):

PR #2070, opened 2023-08-19
MainWindow: gtk4 prep

danirabbit:
Changes from #1896 

* Use bind in Application for window maximized
* Use separate height/width keys because in Gtk4 this becomes a property bind
* Remove `Hdy.NavigationDirection` namespace. Saves line length and no need to port to Adw namespace :)
* Use `child` instead of `add` where appropriate
* Use EventControllerKey
* move `installedview.clear` to delete_event override instead of in destroy.connect
* Replace complicated `deck.get_children ().find (installed_view)` with `installed_view.parent` since it's the same thing


danirabbit:
@jeremypw fixed! Not sure why but the window seems to unset maximized when closing in Gtk3

jeremypw (approved):

PR #2073, opened 2023-08-20
CategoryView: Gtk4 prep

danirabbit:
Changes from #1896 

* Can't subclass Stack in Gtk4, so subclass Box
* Use `child` instead of `add`
* Fix spinner center align

### DIFF
--- a/src/Views/CategoryView.vala
+++ b/src/Views/CategoryView.vala
@@ -15,11 +15,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-public class AppCenter.CategoryView : Gtk.Stack {
+public class AppCenter.CategoryView : Gtk.Box {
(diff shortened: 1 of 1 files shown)


src/Views/CategoryView.vala
```diff
@@ -45,17 +46,21 @@ public class AppCenter.CategoryView : Gtk.Stack {
...
         };
         spinner.start ();
 
-        add (spinner);
-        add (scrolled);
+        stack = new Gtk.Stack ();
+        stack.add (spinner);
+        stack.add (scrolled);
+
+        add (stack);
```
davidmhewitt, line 63: Should this not be `child = stack` ?

danirabbit: I don't think we can use `child` since a box can have more than one child?

davidmhewitt (approved):

PR #2074, opened 2023-08-20
CardNumberEntry: gtk4 prep

danirabbit:
Changes from #1896 

Review with hide whitespace changes

* Remove unused timeout
* Make sure contents is actually different from string builder before changing them
* Use `set_position` instead of `move_cursor`


jeremypw (approved):
Looks sane and entry works as before.

PR #2075, opened 2023-08-20
Application: Set allow_replacement flag

danirabbit:
Allow replacing the running instance with `--gapplication-replace`


jeremypw:
What would happen if the current instance is installing updates in the background?

danirabbit:
@jeremypw I believe it does the same thing as when you click the close button. This is just a convenience for development so that you don't have to kill a background process when running from Terminal

jeremypw (approved):

PR #2076, opened 2023-08-22
SearchView: bind liststore to listbox

danirabbit:
Remove the very clever loading on scroll thing. Probably a bit slower when there's a ton of results, but that will be resolved in Gtk4


jeremypw (approved):
Nice simplification.  I do not think this optimisation is necessary with modern hardware anyway.

PR #2085, opened 2023-09-17
Use portal for background and autostart

lenemter:


leolost2605 (dismissed):
Looks good to me! 🎉

danirabbit:
My only concern about this is, if someone never opens appcenter does that mean they’ll never get update notifications?

danirabbit (approved):
Since we're at the beginning of a new cycle, this is the time for experimentation! So let's try it and see if updates become a concern during testing

Issue #2088, opened 2023-09-20
Don't check for updates in the guest session


danirabbit:
### What Happened?

AppCenter runs in the background and checks for updates when run in the guest session

### Steps to Reproduce

1. Have some updates available
2. Start a guest session
3. AppCenter sends an updates notification

### Expected Behavior

Guests shouldn't be managing updates. We definitely shouldn't show an updates notification to guests, and tbh we probably should not even show them anything to do with updates in AppCenter at all

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

PR #2090, opened 2023-09-20
Don't check for updates in the guest session

meisenzahl:
Closes https://github.com/elementary/appcenter/issues/2088


danirabbit:
I think we probably want to also remove or disable the "Check for updates" button, and possibly avoid showing the update list entirely in the installed view

zeebok:
If you do that, you might want to just remove the settings menu since the other option pertains to installing updates

meisenzahl:
Thank you very much for your feedback. @danirabbit could you please explain again in more detail what you mean by:

> possibly avoid showing the update list entirely in the installed view

src/MainWindow.vala
```diff
@@ -212,8 +212,12 @@ public class AppCenter.MainWindow : Hdy.ApplicationWindow {
         };
         headerbar.set_custom_title (search_clamp);
         headerbar.pack_start (return_button);
-        headerbar.pack_end (menu_button);
-        headerbar.pack_end (view_mode_revealer);
+        if (!Utils.is_running_in_guest_session ()) {
+            headerbar.pack_end (menu_button);
+            headerbar.pack_end (view_mode_revealer);
+        } else {
+            headerbar.pack_end (new Gtk.Label (""));
```
meisenzahl, line 219: If I do not call `headerbar.pack_end`, the headerbar is broken. Is there a better way than adding an empty widget?

danirabbit (approved):

PR #2093, opened 2023-10-03
Add unit tests for Stripe and Houston

davidmhewitt:
Removes the Stripe and Houston API calls and logic from the `StripeDialog` class and moves them into the core and adds unit tests.

Also abstracts the Soup calls out into an interface so we can mock them and check the results in the unit tests. But also means it'll be easier to upgrade the Soup version since there's just one place to change the Soup calls now instead of separate Soup sessions for both Stripe and Houston.

I've tested that I can still pay for apps with this branch.


src/Core/HttpClient.vala
```diff
@@ -0,0 +1,37 @@
...
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+public interface AppCenterCore.HttpClient : Object {
+    public class Response : GLib.Object {
+        public string? body { get; set; }
+        public Gee.HashMap<string, string>? headers { get; set; }
```
tintou, line 23: Can we use GLib.HashTable here?

tintou: (and thus remove the gee dependency from core)

src/Core/Stripe.vala
```diff
@@ -0,0 +1,289 @@
...
+                case "api_connection_error":
+                case "api_error":
+                case "authentication_error":
+                case "invalid_request_error":
+                default:
+                    return _(INTERNAL_ERROR_MESSAGE);
+            }
+        }
+
+        private string get_decline_reason () {
```
tintou, line 99:
```suggestion
        private unowned string get_decline_reason () {
```

src/Core/Stripe.vala
```diff
@@ -0,0 +1,289 @@
...
+        public override unowned ParamSpec? find_property (string name) {
+            // Map the "type" field from JSON to "error_type" property
+            if (name == "type") {
+                return get_class ().find_property ("error_type");
+            }
+
+            return get_class ().find_property (name);
+        }
+
+        public string to_friendly_string () {
```
tintou, line 54:
```suggestion
        public unowned string to_friendly_string () {
```

tintou (approved):

tintou:
Nice!

PR #2094, opened 2023-10-20
Update to Soup 3.0

davidmhewitt:
Adds tests for the Soup client to check its performing HTTP posts correctly, and then upgrades to the 3.0 version.

Please don't squash these commits when merging, I'd prefer to keep them separate.


danirabbit (approved):
Seems very reasonable, thanks for adding tests!

PR #2098, opened 2023-11-11
Suggest packages

meisenzahl:
As mentioned in https://github.com/elementary/os/discussions/651, we install our Flatpak applications via apt. This allows us to switch back to a native application if there are any issues.

However, since we now install all native packages as offline updates, this mechanism fails because we have no Internet connection during the update and therefore cannot install any Flatpak packages. This problem also blocks offline system upgrades.

This branch searches a user and a system directory for configuration files that suggest installing Flatpak packages.

You can place configuration files under `$HOME/.config/appcenter/flatpak.d/` to suggest an installation for the current user. You can also create configuration files under `/etc/appcenter/flatpak.d` to suggest system installations.

A configuration file must be formatted as follows:

```ini
[install]
origin=appcenter
type=app
id=com.github.manexim.insomnia
branch=stable
```

Currently, the packages are only suggested for installation and displayed on the update page.

I tested the changes under Fedora 38. Due to https://github.com/elementary/appcenter/pull/2094, I could not test the changes under OS 7, as I received the following error message: `libsoup2 symbols detected. Using libsoup2 and libsoup3 in the same process is not supported`. Unfortunately, I could not install a daily build of OS 8 (`elementaryos-8.0-daily.20231110.iso`).

### DIFF
--- a/src/Core/BackendAggregator.vala
+++ b/src/Core/BackendAggregator.vala
@@ -104,6 +104,22 @@ public class AppCenterCore.BackendAggregator : Backend, Object {
         return apps;
     }
 
+    public async Gee.Collection<Package> get_suggested_applications (Cancellable? cancellable = null) {
+        var apps = new Gee.TreeSet<Package> ();
+        foreach (var backend in backends) {
+            if (cancellable.is_cancelled ()) {
+                break;
+            }
+
+            var suggested = yield backend.get_suggested_applications (cancellable);
+            if (suggested != null) {
+                apps.add_all (suggested);
+            }
+        }
+
+        return apps;
+    }
+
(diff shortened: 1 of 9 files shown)


meisenzahl:
This is a working prototype. However, the UI still needs a lot of improvement.

@danirabbit what do you think of the approach from a UX perspective?

danirabbit:
@meisenzahl have you seen this proposal? Does this cover the same problem? https://github.com/flatpak/flatpak/issues/5579

meisenzahl:
I just saw Jorge's post and wanted to mention it as well. Yes, this would cover the same problem.

PR #2105, opened 2023-12-29
Show ongoing operations

leolost2605:
Whenever a install button is clicked we add the app to the updates ListStore on the AppListUpdateView. There it shows a progressbutton with cancel ability and gets added to the installed apps list once its done as then the `installed_apps_changed` signal gets fired and the view updated shortly after.

Regarding the current behavior I'm not sure but I think it's wrong anyways because currently an app gets added to the update liststore only after it was installed and the installed_apps_changed signal was then already fired so that it also doesn't get removed but stays there.

To test this you might want to use a large app like GNOME Builder etc. to have enough time to switch over

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -436,15 +436,8 @@ namespace AppCenter.Views {
             updating_all_apps = false;
         }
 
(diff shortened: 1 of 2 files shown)


leolost2605:
Ok nvm the installed_apps_changed signal doesn't get fired at all

PR #2111, opened 2024-02-06
Remove OS updates

danirabbit:
This doesn't completely remove packagekit because we use it for UbuntuDrivers

### DIFF
--- a/src/Core/UpdateManager.vala
+++ b/src/Core/UpdateManager.vala
@@ -21,7 +21,6 @@ public class AppCenterCore.UpdateManager : Object {
     public Package runtime_updates { public get; private set; }
     public int unpaid_apps_number { get; private set; default = 0; }
     public uint64 updates_size { get; private set; default = 0ULL; }
-    public bool has_flatpak_updates { get; private set; default = false; }
 
     construct {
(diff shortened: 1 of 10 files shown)


danirabbit:
@leolost2605 I don't think runtime updates should be asking for a reboot right? Should we remove the restart notification completely?

leolost2605:
@danirabbit I can't comment on runtime updates but AFAIK the only thing that would need a reboot are drivers and with elementary/settings-daemon#132 they can be removed as well so we might just remove it know 🤷

src/Core/UpdateManager.vala
```diff
@@ -71,10 +58,6 @@ public class AppCenterCore.UpdateManager : Object {
             installed_package.update_state ();
         }
 
-
-        uint os_count = 0;
-        string os_desc = "";
-
 #if PACKAGEKIT_BACKEND
```
leolost2605, line 61: I think we can remove all of this?

leolost2605 (changes requested):
Looks good for now but I think we can remove even a bit more? For example we could just make packagekit return in cache refresh since it's not needed for ubuntu drivers i think?

leolost2605 (approved):
LGTM

PR #2114, opened 2024-02-14
Use add/remove_css_class where possible

danirabbit:
This is going to require a larger rewrite to really fix all the way, but this at least reduces the use of get_style_context as much as possible


leolost2605 (approved):
LGTM

PR #2117, opened 2024-02-15
AppInfoView: replace combobox with dropdown

danirabbit:
Knocks out a couple deprecation warnings by using dropdown and Glib.ListStore


leolost2605 (changes requested):
Works as it should however when clicking the dropdown button its background becomes transparent so that it has the banner background color

danirabbit:
@leolost2605 fixed! Just a reminder, you need to do `ninja clean` to rebuild CSS

src/Views/AppInfoView.vala
```diff
@@ -1218,6 +1202,25 @@ public class AppCenter.Views.AppInfoView : AppCenter.AbstractAppContainer {
...
+        list_item.set_data ("title", title);
+        list_item.set_child (title);
+    }
+
+    private void origin_bind_factory (Object object) {
+        var list_item = object as Gtk.ListItem;
+
+        var package = (AppCenterCore.Package) list_item.get_item ();
+
+        var title = list_item.get_data<Gtk.Label> ("title");
```
leolost2605, line 1220: I think instead of doing this manually you can just cast the `list_item.child` as a label

leolost2605 (changes requested):

src/Views/AppInfoView.vala
danirabbit: You got it!

danirabbit:
@leolost2605 can I get a re-review? :)

leolost2605 (approved):
Sorry kinda forgot about this 😔

LGTM!

PR #2118, opened 2024-02-15
Make pages subclass Adw.NavigationPage

danirabbit:
This will be important later to replace Leaflet with NavigationView. It looks like we need some more refactoring before we can make AppInfoView into a NavigationPage


leolost2605 (approved):
Feels much better than checking type :)

PR #2119, opened 2024-02-15
UpdateManager: take over updates_number from client

danirabbit:
Fixes a bug that caused things relying on the client class for updates number to break


leolost2605 (approved):
Nice fix! I noticed that unity badges currently don't show up but it's the same in main so i guess something is wrong elsewhere?

PR #2120, opened 2024-02-15
Add a toggle to show/hide third party apps

### LINKED ISSUE #2113: Doesn't respect flatpak's enumerate/don't enumerate settings
### What Happened?

Flatpak has the ability to mark certain repositories as "no-enumerate." Here is the setup on my machine:
```$ flatpak remotes
Name        Options
appcenter   system
flathub     system
freedesktop system,no-enumerate,filtered
appcenter   user
flathub     user,no-enumerate```

### Steps to Reproduce

To reproduce, install an app from FlatHub that requires dependencies. Flatpak will automatically add the flathub repository and therefore all of the apps from flathub will appear in AppCenter. This is something that most people sideloading flatpak apps probably want --- however, for users like me that are completely comfortable with manually sideloading flatpak apps, the settings of flatpak should be respected, so that the AppCenter only contains the elegant and curated apps from the elementary team.

### Expected Behavior

From `man 1 flatpak-remote-modify`:

`--no-enumerate`
Mark the remote as not enumerated. This means the remote will not be used to list applications, **for instance in graphical installation tools**. It will also not be used for runtime dependency resolution (as with --no-use-for-deps).

With this setting enabled, as described, AppCenter should not list from this repository in the search listing or in the different category lists

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

robog-two:
Fixes #2113 ... sort of

#2113 was about respecting flatpak's dont-enumerate setting, but this is another way of choosing to only see first-party apps from the elementary team, and it uses a switch in the settings menu.


danirabbit:
I think instead of doing this it would be more flexible to have a way to manage remotes so you could choose from a list which remotes to show apps from.

also, to be clear, the majority of the apps in ApCenter are third party. Only the apps that are published with elementary as the developer are 1st party :)

robog-two:
While that is a valid point I think this is a fine way of going about it for now --- managing repositories is sort of difficult and would require much more design, as well as actually respecting flatpak's settings, which AppCenter doesn't even have a means to check as of now. (This is why I added the switch instead of reading no-enumerate) Also, I think it would be possible/better to fork the tool from Pop!_OS, I believe it's called repoman, that does _excatly_ this and is already integrated with their AppCenter fork (Pop!_Shop)

PR #2122, opened 2024-02-16
ComponentValidator: drop hiddenapps function

### LINKED ISSUE #438: Blacklist does not support comments

danirabbit:
Closes #438 

This is from a time when we were pulling apps from the Ubuntu repos and we had to specifically hide components that we didn't want to show like imagemagick and other utilities


leolost2605:
Maybe we could just make `validate` static and move it to Utils or something? Having a whole separate class for just a (static) method seems a bit overkill 🤷

danirabbit:
@leolost2605 I moved it to BackendInterface since it's only used by the backends. How's that? :)

leolost2605 (approved):
Nice cleanup!

PR #2123, opened 2024-02-16
show_uri → UriLauncher

danirabbit:


ryonakano (approved):
Looks good and works as expected on Daily 8. Note that we can no longer release AppCenter for Horus after merging this branch.

If possible, we might want to update the docker image in the following line to `development-target`, to prevent the gettext actions from being failing, since the current `next-unstable` image [points to `horus-unstable`](https://github.com/elementary/docker/blob/b6a3cfec9a1f0984d439f348552a86dface35c38/next-unstable)

https://github.com/elementary/appcenter/blob/405211349df7c3c0b77225d4f91976861f6673b1/.github/workflows/gettext.yml#L11

danirabbit:
@ryonakano looks like it broke anyways. I proposed https://github.com/elementary/actions/pull/81

PR #2125, opened 2024-02-16
Separate ProgressButton from AbstractAppContainer

danirabbit:


leolost2605 (changes requested):
Looks really great, especially that you can now also see progress in the appinfoview. It has the same issue as the dropdown button though with the background becoming transparent when working.

danirabbit:
@leolost2605 good catch! Fixed :)

leolost2605 (approved):
LGTM!

PR #2127, opened 2024-02-22
Create ActionStack widget

danirabbit:
So that we can make AppInfoView a NavigationPage


leolost2605 (changes requested):
For already installed apps the uninstall button isn't revealed. I.e. if I install an app and watch it install it gets revealed but when I close appcenter and open it again it stays hidden.

Other than that it's looking good :)

leolost2605 (approved):
LGTM!

PR #2130, opened 2024-02-29
MainWindow: Adw.Leaflet → Adw.NavigationView

danirabbit:
Replaces deprecated Adw.Leaflet with Adw.NavigationView

* Handles mouse buttons already
* Knows not to remove pages with `add` when they're popped
* Can't remove/pop arbitrary children, can only navigate back to that page
* Automatically removes popped children

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -507,11 +507,6 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
     }
 
     private void show_category (AppStream.Category category) {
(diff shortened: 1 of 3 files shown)


leolost2605 (changes requested):
Much cleaner and works as expected except for that it reintroduces #1947

danirabbit:
@leolost2605 I think I fixed it!

leolost2605 (approved):
LGTM!

PR #2133, opened 2024-03-19
AppInfoView: badge for appcenter instead of uncurated

danirabbit:
Screenshot from 2024-03-18 19 26 13

Finish off the last of the "Curated" language by giving apps that were made for elementary OS a nice badge instead of a warning on apps that weren't


lenemter (approved):
This looks so much better ❤️

PR #2135, opened 2024-03-19
AppInfoView: Fix parsing locale with newer AppStream

### LINKED ISSUE #2134: AppCenter says an app is "Not Translated" that provides translation info
### What Happened?

AppCenter reports an app is that provides translation info as "Not Translated" in the app detail page.

### Steps to Reproduce

Maybe you should set your system locale to other than English.

1. Open AppCenter
2. Type `reco` in the search entry
3. Open the app detail page of Reco
4. See there are "Not Translated" label in your language is shown:

Screenshot from 2024-03-19 23-49-24

Reco already provides translation info, so this shouldn't be shown.

https://github.com/ryonakano/reco/blob/1f248f1231456a10fe69db0b58faed5ae5fff4b3/data/reco.metainfo.xml.in#L6

### Expected Behavior

AppCenter doesn't report "Not Translated" for apps with translation info.

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

ryonakano:
Fixes #2134

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -284,17 +284,21 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
 #endif
 
 #if HAS_APPSTREAM_1_0
-            var active_locale = "en-US";
+            const string DEFAULT_LOCALE = "en-US";
+            const string LOCALE_DELIMITER = "-";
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
This is cleaner and makes sense. Nice job!

PR #2136, opened 2024-03-19
Drop options for Pop! Shop

danirabbit:
System76 has their own app store now, so drop options we were carrying for Pop! Shop


zeebok (approved):
This seems to look alright to me

PR #2151, opened 2024-05-31
AppInfoView: Use AppStream API to get license info

ryonakano:
- Get license name from AppStream if >= 1.0 (if not SPDX license identifiers are used as before)
- Get license URL from AppStream instead of manually constructing


danirabbit (approved):
Very cool!

PR #2153, opened 2024-06-04
Package: assume all packages are flatpak

danirabbit:
Don't check if we're a flatpak. Packages are always a flatpak


zeebok (approved):
This looks good. Does make me wonder, are there any default apps that are still deb packages and potentially not show up in the Installed View?

danirabbit:
@zeebok Yeah, but nothing we can really do about that for now. Anything installed via deb is considered part of the OS

PR #2154, opened 2024-06-05
Move validate () to FlatpakBackend

danirabbit:
This is the only place this is used now


zeebok (approved):

zeebok:
The Backend class is starting to be less and less useful :P

PR #2156, opened 2024-06-05
Remove BackendInterface

danirabbit:
Only one backend so we don't need to save which backend an update is from


zeebok (approved):
Code looks good. Plenty of compiler warnings but they seem to all be deprecation warnings than anything pertaining to this.

danirabbit:
Ah yeah there’s a bunch of stuff related to style contexts but I’m not sure it’s really easily solvable until we get newer Gtk

PR #2162, opened 2024-06-12
Move updates_liststore to UpdateManager

danirabbit:
* Make sure other classes are talking to UpdateManager not to a piece of UI
* Make sure we're not getting the list of updates twice. We can add updates to the liststore as we get them in the UpdateManager. Faster
* Update the UI for the list of installed apps separately from the list of apps with updates
* Fix the loading screen
* Make sure the count of the number of updates is based on the number of packages in the liststore
* Use a simplified comparison func for installed apps
* Only try to get runtime updates size if we counted more than 0
* Remove unused runtime updates strings. We don't add the package to the list unless there are updates


src/Core/UpdateManager.vala
```diff
@@ -25,9 +25,10 @@ public class AppCenterCore.UpdateManager : Object {
     public signal void installed_apps_changed ();
     public signal void cache_update_failed (Error error);
 
+    public ListStore updates_liststore { public get; private set; }
     public Package runtime_updates { public get; private set; }
     public int unpaid_apps_number { get; private set; default = 0; }
-    public uint updates_number { get; private set; default = 0U; }
+    public uint updates_number { get; set; default = 0U; }
```
leolost2605, line 31: Absolutely not necessary but to make it even more expressive you could do 
```suggestion
    public uint updates_number { get { return updates_liststore.n_items; } }
```

danirabbit: I think it has to be settable because there are things that bind to this property so they need the set notify signal

leolost2605 (approved):

PR #2164, opened 2024-06-12
InstalledPackageRowGrid: only show release notes buttons if available

### LINKED ISSUE #2035: Only show update info icon when there is info available?
### Problem

Currently we get update info icons next to almost all updates:
Screenshot from 2023-05-10 13 16 05@2x

But often, beside a date, there is no real info available:
Screenshot from 2023-05-10 13 15 55@2x

It's rather annoying to have to open this window to see whether there is info. 

### Proposal

Shouldn't we only show the icon when there is info available?

Maybe we can communicate the release data in a different way? 

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #2035


zeebok (approved):
Not sure if there is a simple way to test but the code looks good

danirabbit:
Ah yeah I just kinda of keep some updates on the queue for the past week to test things lol

PR #2165, opened 2024-06-12
CSS: make categories all use the same font

### LINKED ISSUE #736: Fonts of some categories don't support some diacritics
In czech language a letter "ě" is used, that is not supported by those fonts. It looks ugly. (This will probably affect some other languages too)

snimek z 2018-07-09 09 25 41

danirabbit:
Use the system font for all the category cards and make them much smaller so they can all fit on the homepage at smaller window sizes

Fixes #736 

Screenshot from 2024-06-12 13 55 19


leolost2605 (approved):
Even though I quite like the different fonts I guess this is the best way forward regarding compatibility and accessibility :(

PR #2167, opened 2024-06-13
SearchView: make search_entry internal

### LINKED ISSUE #554: Can't search from an external source while open on an app from search
If I have AppCenter open up to an app's page that I arrived to from a search in AppCenter, using an external search source (like slingshot) puts AppCenter into a weird state: it stays on the app page but fills the search bar with my search term. I then have to clear the search (which takes me to wherever I was last in AppCenter) and manually re-search.

On app page after searching "spice" in AppCenter:
screenshot

After searching for "test search term" from slingshot:
screenshot

This seems to be affected by whether or not you used search within AppCenter to get to the app page. If you didn't, it works as expected. If you did, the search is not performed.

danirabbit:
Fixes #554

* Makes it so you can start a search from any page
* Makes more searching logic internal to the search page
* You can have multiple searches without messing up your navigation history
* Will make it easy for each page to handle its own headerbar in a future branch

### DIFF
--- a/src/Views/SearchView.vala
+++ b/src/Views/SearchView.vala
@@ -22,15 +22,42 @@
 public class AppCenter.SearchView : Adw.NavigationPage {
     public signal void show_app (AppCenterCore.Package package);
 
-    public string? current_search_term { get; set; default = null; }
+    public const int VALID_QUERY_LENGTH = 3;
+
+    public string search_term { get; construct; }
+    public bool mimetype { get; set; default = false; }
 
     private GLib.ListStore list_store;
+    private Gtk.SearchEntry search_entry;
+    private Granite.Placeholder alert_view;
+
+    public SearchView (string search_term) {
+        Object (search_term: search_term);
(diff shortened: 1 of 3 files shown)


src/Views/SearchView.vala
```diff
@@ -48,28 +75,85 @@ public class AppCenter.SearchView : Adw.NavigationPage {
...
+                add_packages (found_apps);
+            } else {
+                var category = update_category ();
+
+                found_apps = flatpak_backend.search_applications (search_entry.text, category);
+                add_packages (found_apps);
+            }
+
+        } else {
+            alert_view.description = _("The search term must be at least 3 characters long.");
```
leolost2605, line 137: On first opening this doesn't get triggered

leolost2605 (changes requested):
Only one thing apart from that looks pretty good :)

leolost2605 (approved):
LGTM!

PR #2168, opened 2024-06-13
Application: don't try to get parent when backgrounding

### LINKED ISSUE #2166: Segfault
### What Happened?

No idea when exactly this happens but on latest main and (only?) when running under wayland i get pretty reliable segfaults.

### Steps to Reproduce

Idk but it's pretty common

### Expected Behavior

Dont segfault

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

```shell
#0  0x00007ac9a342c80b in gdk_wayland_toplevel_unexport_handle () from /lib/x86_64-linux-gnu/libgtk-4.so.1
[Current thread is 1 (Thread 0x7ac9a0571240 (LWP 5732))]
(gdb) bt
#0  0x00007ac9a342c80b in gdk_wayland_toplevel_unexport_handle () at /lib/x86_64-linux-gnu/libgtk-4.so.1
#1  0x00007ac9a3b915c9 in ??? () at /lib/x86_64-linux-gnu/libportal.so.1
#2  0x00007ac9a3b917b4 in ??? () at /lib/x86_64-linux-gnu/libportal.so.1
#3  0x00007ac9a3dc2a30 in ??? () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#4  0x00007ac9a3f4a48e in ??? () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#5  0x00007ac9a3fa9717 in ??? () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#6  0x00007ac9a3f49a53 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#7  0x00007ac9a3da188d in g_application_run () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#8  0x000064c872e360a7 in _vala_main (args=0x7ffd12b6e0b8, args_length1=1) at ../src/Application.vala:400
#9  0x000064c872e360ee in main (argc=1, argv=0x7ffd12b6e0b8) at ../src/Application.vala:398
```

### Hardware Info

_No response_

danirabbit:
Fixes #2166 

It seems like this is pretty much always going to be null? Maybe the background request should be moved to when the window is shown or something? But the way it is right now I don't think this can ever get a valid window

### DIFF
--- a/data/appcenter.metainfo.xml.in
+++ b/data/appcenter.metainfo.xml.in
@@ -87,6 +87,7 @@
         <issue url="https://github.com/elementary/appcenter/issues/2035">Only show update info icon when there is info available</issue>
         <issue url="https://github.com/elementary/appcenter/issues/2088">Don't check for updates in the guest session</issue>
         <issue url="https://github.com/elementary/appcenter/issues/2134">AppCenter says an app is "Not Translated" that provides translation info</issue>
+        <issue url="https://github.com/elementary/appcenter/issues/2166">Segfault on wayland</issue>
       </issues>
     </release>
 
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -233,15 +233,13 @@ public class AppCenter.App : Gtk.Application {
     public async void request_background () {
         var portal = new Xdp.Portal ();
 
-        Xdp.Parent? parent = active_window != null ? Xdp.parent_new_gtk (active_window) : null;
-
         var command = new GenericArray<weak string> ();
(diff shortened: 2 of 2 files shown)


leolost2605:
Maybe we should return true in close_request so that the window stays? I think in mail we do it that way

danirabbit:
@leolost2605 ah yeah that makes sense but I'm just getting a freeze when I try that :thinking:

PR #2172, opened 2024-06-16
AppListUpdateView: set loading placeholder only when loading

danirabbit:
Try to set the loading placeholder only at the most appropriate times and don't avoid showing loaded content


zeebok:
Code looks good. I am seeing this gap between the updates section and installed app section. Is that a new issue from this?
Screenshot from 2024-06-16 22 10 57

danirabbit:
Nope that's been like that for quite a while 😅

zeebok (approved):
The code looks good. With this

PR #2173, opened 2024-06-17
InstalledPackageRowGrid: make more compact

danirabbit:
Makes it so more packages fit on screen at smaller window sizes:
* Wrap app names
* Make version labels small
* Use a reveal method that doesn't preserve space when hidden
* Make sure release notes revealers don't get margins when hidden

## AFTER
Screenshot from 2024-06-17 12 26 53

## BEFORE
Screenshot from 2024-06-17 12 30 18

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -103,8 +103,7 @@ namespace AppCenter.Views {
 
             installed_flowbox = new Gtk.FlowBox () {
                 column_spacing = 24,
-                homogeneous = true,
-                max_children_per_line = 4,
+                max_children_per_line = 5,
                 row_spacing = 12
             };
             installed_flowbox.bind_model (installed_liststore, create_installed_from_package);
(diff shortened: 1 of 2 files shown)


zeebok (commented):
Looks good. I am noticing some oddities that I am not sure is something to be concerned about or not. If the daemon is not running (I had killed it to install the compiled version) and you go to the update view, it will show the placeholder and then show installed apps with 2 per row. Then when it finishes finding updates and reveals, changes the installed apps section to 3 per row?

If this isn't a use case that we are worried about, then I will approve the PR

danirabbit:
@zeebok yeah I noticed something like it seems the list of installed apps sometimes contains apps that have updates until it finalizes. So you get rows with the release notes button which is a wider row. I'm not sure why that happens but I think that would be more another branch. This whole view kind of gets fucky while loading so we still have plenty of work to do here 😅

zeebok (approved):

PR #2174, opened 2024-06-17
PackageRow: rm PackageRow.installed

danirabbit:
Remove another extra layer of abstraction


zeebok (approved):
Doesn't appear to break anything

PR #2180, opened 2024-06-20
UpdateManager: remove runtime size check

danirabbit:
Fixes runtime updates not showing in updates list

This size check was added [here](https://github.com/elementary/appcenter/pull/1960), But we no longer create a "No runtime updates" package so we don't need this. A package is only inserted into the list if we find updates already


zeebok (approved):
Dang you found the cause quick. The fix works a treat!

danirabbit:
I've been digging around in this muck for a bit now :sweat_smile: Thanks for the quick review!

PR #2185, opened 2024-06-20
Fast start

leolost2605:
The installed check is queued and doesn't run until some more time intensive operations finish. However it is actually quite fast so IMO we can do it sync (I didn't notice any bad performance with this). Alternatively we can do it async but I think this might be worth doing on its own thread instead of queuing it as a job.

### DIFF
--- a/src/Views/Homepage.vala
+++ b/src/Views/Homepage.vala
@@ -332,9 +332,7 @@ public class AppCenter.Homepage : Adw.NavigationPage {
             var installed = false;
             foreach (var origin_package in package.origin_packages) {
                 try {
-                    warning ("CHECK INSTALLED");
                     if (AppCenterCore.FlatpakBackend.get_default ().is_package_installed (origin_package)) {
-                        warning ("CHECKED INSTALLED");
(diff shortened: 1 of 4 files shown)


src/Views/Homepage.vala
```diff
@@ -332,7 +332,9 @@ public class AppCenter.Homepage : Adw.NavigationPage {
             var installed = false;
             foreach (var origin_package in package.origin_packages) {
                 try {
-                    if (yield AppCenterCore.FlatpakBackend.get_default ().is_package_installed (origin_package)) {
+                    warning ("CHECK INSTALLED");
+                    if (AppCenterCore.FlatpakBackend.get_default ().is_package_installed (origin_package)) {
+                        warning ("CHECKED INSTALLED");
```
danirabbit, line 337: Probably should remove these or at least set to debug if it's necessary :sweat_smile:

leolost2605: Oops yeah they were just for testing :)

danirabbit (approved):
SO fast :rocket:

PR #2187, opened 2024-06-20
FlatpakBackend: set action enabled in main thread

danirabbit:
Accidentally introduced a crash in #2181


davidmhewitt:
I'm reviewing on my phone so it's hard to see the full context. But I think the working variable is changed from another thread so the notify signal may run in another thread context.

It would be better to wrap the GTK code in an Idle.add which would ensure it runs on the GTK thread.

danirabbit:
@davidmhewitt wrapped in an Idle instead. Seems to work. Thanks for the tip!

davidmhewitt (approved):

PR #2193, opened 2024-06-21
AppInfoView: redesign links

### LINKED ISSUE #1134: Show donation URL button if payments not enabled
I think we've talked about this but I couldn't find an issue or code for it, so filling for discussion. If an app does not have monetization enabled but provides a donation URL, we could consider showing that along with other metadata buttons in the footer. 

Pro: apps from developers who cannot use Stripe i.e. in unsupported countries could have an easier route for monetization

Con: developers in supported countries could try to do this to bypass the regular AppCenter payment flow

I think the advantages of the AppCenter payment flow are incentive enough for developers to use it instead of choosing to go the donation URL route but I'm up to discuss it.

danirabbit:
Fixes #1134

Show more links, add some color. Group links related to contributing/giving vs requesting/contacting/receiving

Screenshot from 2024-06-21 11 46 44
Screenshot from 2024-06-21 11 47 37

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -660,62 +660,15 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
             load_extensions.begin ();
         }
 
-        var links_flowbox = new Gtk.FlowBox () {
-            column_spacing = 12,
-            row_spacing = 6,
-            hexpand = true
-        };
-
-        var project_license = package.component.project_license;
-        if (project_license != null) {
-            string? license_copy = null;
-            string? license_url = null;
-
-            parse_license (project_license, out license_copy, out license_url);
-
-            var license_button = new UrlButton (_(license_copy), license_url, "text-x-copying-symbolic");
-
-            links_flowbox.append (license_button);
-        }
-
-        var homepage_url = package_component.get_url (AppStream.UrlKind.HOMEPAGE);
-        if (homepage_url != null) {
-            var website_button = new UrlButton (_("Homepage"), homepage_url, "web-browser-symbolic");
-            links_flowbox.append (website_button);
-        }
-
-        var translate_url = package_component.get_url (AppStream.UrlKind.TRANSLATE);
-        if (translate_url != null) {
-            var translate_button = new UrlButton (_("Translate"), translate_url, "preferences-desktop-locale-symbolic");
-            links_flowbox.append (translate_button);
-        }
-
-        var bugtracker_url = package_component.get_url (AppStream.UrlKind.BUGTRACKER);
-        if (bugtracker_url != null) {
-            var bugtracker_button = new UrlButton (_("Send Feedback"), bugtracker_url, "bug-symbolic");
-            links_flowbox.append (bugtracker_button);
-        }
-
-        var help_url = package_component.get_url (AppStream.UrlKind.HELP);
-        if (help_url != null) {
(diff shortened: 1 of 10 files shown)


ryonakano (commented):
Really cute and nice improvements! Leaving some comments.

src/Widgets/LinkListBox.vala
```diff
@@ -0,0 +1,286 @@
...
+        if (contact_listbox.get_first_child () != null) {
+            flowbox.append (contact_listbox);
+        }
+
+        // Don't let invisible container get focus
+        var flowbox_child = flowbox.get_first_child ();
+        while (flowbox_child != null) {
+            flowbox_child.focusable = false;
+            flowbox_child = flowbox_child.get_next_sibling ();
+        }
```
ryonakano, line 153: I'm not sure what this actually does. The comment says that this prevents invisible flowboxes from being focused but the code doesn't check for the visibility of flowboxes.

ryonakano:
Also, there are extra margin right to the links if contribute_listbox is not appened.

Screenshot from 2024-06-23 11-28-00

src/Widgets/LinkListBox.vala
```diff
@@ -0,0 +1,286 @@
...
+                component.get_name (),
+                component_id,
+                uri_or_key
+            ) {
+                modal = true,
+                transient_for = ((Gtk.Application) Application.get_default ()).active_window
+            };
+
+            stripe_dialog.download_requested.connect (() => {
+                if (stripe_dialog.amount != 0) {
```
leolost2605, line 278: This will cause a mem leak :(
(Using a local variable inside a lambda that's connected to a signal)

leolost2605 (changes requested):

src/Widgets/LinkListBox.vala
danirabbit: Updated! Is that better?

src/Views/AppInfoView.vala
```diff
@@ -1185,80 +1095,6 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
         title.label = package.origin_description;
     }
 
-    class UrlButton : Gtk.Box {
-        public UrlButton (string label, string? uri, string icon_name) {
-            add_css_class (Granite.STYLE_CLASS_DIM_LABEL);
-            tooltip_text = uri;
```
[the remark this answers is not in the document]
danirabbit: I decided to add the description label directly to the box as GNOME Software is doing here so that it's also accessible for touch

src/Widgets/LinkListBox.vala
danirabbit: Moved this bit to the stripe dialog itself which is more DRY anyways

danirabbit:
Updated! I added description labels directly to the box here and also made sure to update the icon and a11y for stripe buttons. You can see here how it differs between a link and a payment that's handled by stripe:

Screenshot from 2024-06-24 18 17 39
Screenshot from 2024-06-24 18 20 19
Screenshot from 2024-06-24 18 22 42

leolost2605 (approved):
Looks good from my side :)

src/Widgets/LinkListBox.vala
ryonakano: Thank you, now I understand what we're doing here.

ryonakano (approved):
Maybe we can create the LinkRow class in Granite for consistent design, because the same design is used in https://github.com/elementary/switchboard-plug-about/pull/325 too, but we can discuss that later.

PR #2195, opened 2024-06-25
AppListUpdateView: Avoid mentioning OS updates in user-facing texts

ryonakano:
Because OS updates has been removed from AppCenter


danirabbit (approved):
Good catch!

PR #2199, opened 2024-07-02
AppListUpdateView: DRY Update All

danirabbit:
UpdateManager:
* Make `update_all` its own function that is cancellable
* Make sure to use class-wide cancellable for auto updates

AppListUpdateView:
* Use update_all function from update_manager
* Make sure to set action buttons back to sensitive if update_all is canceled or errors

### DIFF
--- a/src/Core/UpdateManager.vala
+++ b/src/Core/UpdateManager.vala
@@ -185,7 +185,7 @@ public class AppCenterCore.UpdateManager : Object {
         return updates_number;
     }
 
-    public async void update_all (Cancellable? cancellable = null) throws Error {
+    public async void update_all (Cancellable? cancellable) throws Error {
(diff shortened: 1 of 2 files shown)


zeebok (approved):
Not entirely sure what the action sensitivity does but things appear to work as normal

danirabbit:
@zeebok if you use the "Update All" button in the interface it'll make all the update buttons insensitive so you can't just keep clicking update over and over. It won't have any affect if the update was triggered automatically however. Still needs more work to handle that

zeebok:
@danirabbit Gotcha, I thought maybe it was making all buttons insensitive until it finished looking for updates or something, but wasn't seeing that happen

zeebok:
Can confirm that is the behavior, and is cancellable

PR #2203, opened 2024-08-07
Add support for new developer id

### LINKED ISSUE #2102: Use new "developer ID" metainfo
### Problem

We currently identify "apps by the same developer" using the `developer_name` tag. Since this is in theory localizable and not necessarily always typed the same or unique, it could cause issues:

https://github.com/elementary/appcenter/blob/6e78d5d0b990f05cff27a2548080099defe542c5/src/Core/FlatpakBackend.vala#L500-L533

The AppStream spec now has a solution for this in the form of an ID field for the developer:
https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer

### Proposal

Support the developer ID and use that to identify other packages by the same developer. Fallback to author name if not available. I believe we do this package search in both the `FlatpakBackend` code (above), and the `PackageKitBackend` so it would need to be updated in at least two places. Though a code search for `developer_name` would help here.

### Prior Art (Optional)

_No response_

kerunaru:
This is a first approach to support the new developer id (closes #2102). Although I added a new property to `Package.vala` populating that id, it is really uncommon that this info is available.

The "More apps from this developer" feature is kind of broken right now because of this... Should I search by author name like before in case no more apps are found?

I left some `warning` logs just to check the realted properties to this functionality in `FlatpakBackend.vala`.

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -297,6 +297,23 @@ public class AppCenterCore.Package : Object {
         }
     }
 
+    private string? _author_id = null;
+    public string? author_id {
+        get {
+            if (_author_id != null) {
+                return _author_id;
+            }
+
+            _author = component.get_developer ().get_id ();
+
+            if (_author_id == null) {
+                return _author;
+            }
+
+            return _author_id;
+        }
+    }
+
     private string? _author_title = null;
     public string author_title {
         get {
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -304,7 +304,9 @@ public class AppCenterCore.Package : Object {
                 return _author_id;
             }
 
-            return component.get_developer ().get_id ();
+            author_id = component.get_developer ().get_id ();
+
+            return author_id;
         }
     }
 
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -304,9 +304,9 @@ public class AppCenterCore.Package : Object {
                 return _author_id;
             }
 
-            author_id = component.get_developer ().get_id ();
+            _author_id = component.get_developer ().get_id ();
 
-            return author_id;
+            return _author_id;
         }
     }
 
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -304,13 +304,7 @@ public class AppCenterCore.Package : Object {
                 return _author_id;
             }
 
-            _author = component.get_developer ().get_id ();
-
-            if (_author_id == null) {
-                return _author;
-            }
-
-            return _author_id;
+            return component.get_developer ().get_id ();
         }
     }
 
--- a/src/Views/AuthorView.vala
+++ b/src/Views/AuthorView.vala
@@ -23,9 +23,7 @@ private class AppCenter.AuthorView : Gtk.Box {
             return;
         }
 
-        Gee.Collection<AppCenterCore.Package> author_packages;
-
-        author_packages = package.author_id == null
+        var author_packages = package.author_id == null
             ? AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX)
             : AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author_id (package.author_id, AUTHOR_OTHER_APPS_MAX);
 
--- a/src/Views/AuthorView.vala
+++ b/src/Views/AuthorView.vala
@@ -23,11 +23,11 @@ private class AppCenter.AuthorView : Gtk.Box {
             return;
         }
 
-        var author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author_id (package.author_id, AUTHOR_OTHER_APPS_MAX);
+        Gee.Collection<AppCenterCore.Package> author_packages;
 
-        if (author_packages.size <= 1) {
-            author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX);
-        }
+        author_packages = package.author_id == null
+            ? AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX)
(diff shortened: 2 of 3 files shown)


src/Core/Package.vala
```diff
@@ -297,6 +297,23 @@ public class AppCenterCore.Package : Object {
...
     }
 
+    private string? _author_id = null;
+    public string? author_id {
+        get {
+            if (_author_id != null) {
+                return _author_id;
+            }
+
+            _author = component.get_developer ().get_id ();
```
ryonakano, line 307:
```suggestion
            _author_id = component.get_developer ().get_id ();
```

Maybe a copy & paste error

src/Core/Package.vala
```diff
@@ -297,6 +297,23 @@ public class AppCenterCore.Package : Object {
...
+
+            if (_author_id == null) {
+                return _author;
+            }
+
+            return _author_id;
+        }
```
ryonakano, line 314: This fallback means the `author_id` property never returns null, but you check for null in AuthorView. Maybe we shouldn't fallback to `_author` here?

ryonakano (commented):
Thank you for your contribution! I'm leaving some comments, although your PR is still draft.

src/Core/Package.vala
kerunaru: True. I'll change this a little bit just to include your other suggestion.

kerunaru:
> Thank you for your contribution! I'm leaving some comments, although your PR is still draft.

Yes. It is still a draft because I want some feedback from your part about the questions on this PR description. :smiley:

src/Core/Package.vala
```diff
@@ -297,6 +297,17 @@ public class AppCenterCore.Package : Object {
...
     }
 
+    private string? _author_id = null;
+    public string? author_id {
+        get {
+            if (_author_id != null) {
+                return _author_id;
+            }
+
+            return component.get_developer ().get_id ();
```
ryonakano, line 307:
```suggestion

            _author_id = component.get_developer ().get_id ();
            return _author_id;
```

You didn't need to remove substitution to `_author_id`, otherwise we need to run `component.get_developer ().get_id ();` every access to the `author_id` property

kerunaru: Fixed!

src/Views/AuthorView.vala
```diff
@@ -23,7 +23,12 @@ private class AppCenter.AuthorView : Gtk.Box {
             return;
         }
 
-        var author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX);
+        var author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author_id (package.author_id, AUTHOR_OTHER_APPS_MAX);
+
+        if (author_packages.size <= 1) {
+            author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX);
+        }
+
```
ryonakano, line 31: I guess both `package.author_id` and `package.component.get_developer ().get_id ()` would be null and it would match in the get_packages_by_author_id method if the developer id is not provided

```suggestion
        Gee.Collection<Package> author_packages;
        if (package.author_id != null) {
            author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author_id (package.author_id, AUTHOR_OTHER_APPS_MAX);
        } else {
            author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX);
        }

```

ryonakano: Warnings shown when I navigate to an app page that does not provide developer id:

```
** (io.elementary.appcenter:5675): CRITICAL **: 22:39:07.544: app_center_core_flatpak_backend_get_packages_by_author_id: assertion 'author_id != NULL' failed

** (io.elementary.appcenter:5675): CRITICAL **: 22:39:07.545: gee_collection_get_size: assertion 'self != NULL' failed
```

kerunaru: Can you give me an example of application which has no `author_id`?

ryonakano: Apps that does not have the `<developer id="foo.bar">` tag. For example, [Pin It!](https://appcenter.elementary.io/com.github.ryonakano.pinit/) published on AppCenter is not updated recently and thus does not have it.

kerunaru: Good! Those logs dissapeared with the last changes. :smile:

src/Views/AuthorView.vala
```diff
@@ -23,7 +23,12 @@ private class AppCenter.AuthorView : Gtk.Box {
             return;
         }
 
-        var author_packages = AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX);
+        Gee.Collection<AppCenterCore.Package> author_packages;
+
+        author_packages = package.author_id == null
+            ? AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX)
+            : AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author_id (package.author_id, AUTHOR_OTHER_APPS_MAX);
+
```
ryonakano, line 31: I personally prefer using `if` than `foo ? bar : baz` here because the part of `bar` and `baz` is too long in this case. But I don't insist on that.

Instead, if you use `foo ? bar : baz` here you don't need to separate declaration and substitution of `author_packages`:
```suggestion
        var author_packages = package.author_id == null
            ? AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author (package.author, AUTHOR_OTHER_APPS_MAX)
            : AppCenterCore.FlatpakBackend.get_default ().get_packages_by_author_id (package.author_id, AUTHOR_OTHER_APPS_MAX);

```

kerunaru: Done!

ryonakano (approved):
Looks good and works as expected for me.

The `get_packages_by_author_id` and `get_packages_by_author` methods do almost similar thing so maybe we can work on DRY on them, but we can revisit later if we want.

kerunaru:
Woah! This is the first time I contribute to an OS! Thanks for accepting my contribution :smile:

PR #2207, opened 2024-08-29
Banner: create providers for display

danirabbit:
Fixes deprecated adding provider to style context

If a brand color is available, creates a style provider with a css class for that color and adds it to a list of providers. Doesn't add a provider if no brand color available. Re-uses providers if two apps use the same brand color

### DIFF
--- a/data/styles/Banner.scss
+++ b/data/styles/Banner.scss
@@ -3,6 +3,9 @@
  * SPDX-FileCopyrightText: 2017-2024 elementary, Inc. (https://elementary.io)
  */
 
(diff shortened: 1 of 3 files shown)


jeremypw (dismissed):
This differs from main for applications that have not set a custom background (?) e.g. Image Optimizer.   For these, the banner has no background color whereas in main it uses the user's accent color.

danirabbit:
@jeremypw is this different from what you're seeing?

Screenshot from 2024-12-09 17 32 09
Screenshot from 2024-12-09 17 32 56

zeebok (approved):
This works right for me on OS8. Code seems good too

Issue #2210, opened 2024-09-04
Filter screenshots by environment


danirabbit:
### Problem

Apps can provide multiple screenshots for different desktop environments or using light and dark styles. There is a way now in appstream for developers to declare the environment of a screenshot, but AppCenter just shows all screenshots

### Proposal

Filter screenshots to match the current desktop environment and match the dark/light style preference. For example, Harvey should show the light screenshot in light mode and the dark screenshot in dark mode, but not both

danirabbit:
Keeping this issue open since we now have dark mode filtering but not desktop environment filtering

edwood-grant:
I can definitely get the desktop environment via `Environment.get_variable("XDG_SESSION_DESKTOP");` and compare. But I have a question here, actually. 

Let's see, for example, Harvey. The application has two screenshots. Their environment IDs on one is pantheon:dark and the other is pantheon

If I happen to be on GNOME for example, the environment ID will not match the XDG_SESSION_DESKTOP environment variable, and this would result in no screenshots. Would this be an acceptable filter? Or would you rather have at least one screenshot if nothing matches?

It's a bit important since the latter would surely involve a bit of more code to guarantee at least one screenshot if nothing matches... would probably be best to actually filter the screenshots by DE and style (dark, light) with `sort_screenshots` and then just iterate until you don't find anything.

And if you don't find anything then grab the first screenshot at the very least I suppose. I would still not take into account null environment screenshots, though.

Just wanting to understand what would be preferable here. It seems that latter option would be the actual complete solution for this allowing to have all the proper screenshots first and make it easy to verify if there is at least a screenshot that matches the filter (otherwise, leave the first screenshot, whatever it is and leave it like that)

danirabbit:
Yeah I think what we want to do is, if we can get Pantheon screenshots, use those. But if we can't, use whatever is available. Prioritizing the desktop, then light or dark

So for example if we have a list that's `["Pantheon", "Pantheon:Dark", "GNOME", "GNOME:Dark]`then we should show the Pantheon screenshots in light and dark mode respectively. This is like the ideal case.

If we have `["Pantheon", "GNOME", "GNOME:Dark]` we should prefer the Pantheon screenshot, even in dark mode

If we have `["GNOME", "GNOME:Dark]` then we should just go ahead and use the GNOME screenshots in dark and light mode respectively

PR #2225, opened 2024-12-08
Filter screenshot by environment (elementary#2210)

edwood-grant:
### Short summary
Aims to implement #2210 

Will read the environment property of the screenshot and will filter it out if:

* _Prefer dark mode_ is true, and the screenshot _does not_ have a dark mode environment id (i.e. 'dark' in the environment ID). 
* _Prefer dark mode_ is false, and the screenshot _does_ have a dark mode environment id.

This won't consider null environments in the screenshots.

### Tests

I have only tested this on the [Harvey](https://appcenter.elementary.io/com.github.danrabbit.harvey/) application, since this is the only app that I know has this implemented. I also tried a random assortment of titles with and without screenshots without any problems/crashes, though.


danirabbit (approved):
Looks good to me! Nice work :tada:

PR #2226, opened 2024-12-09
Parse all branding colors from appstream (elementary#2140)

### LINKED ISSUE #2140: Theme specific branding colors
### Problem

It seems like appcenter will always take the branding color without a scheme

https://github.com/elementary/appcenter/blob/f8db5192b69da0846fa9e62f789e15e0dbf0cf29/src/Core/Package.vala#L826

(I might just no be understanding vala)

Flathub apps will almost always have a `scheme_preference` that should be read instead.  Those will look similar to the spec https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-branding

```
    <branding>
      <color type="primary" scheme_preference="light">#ff00ff</color>
      <color type="primary" scheme_preference="dark">#993d3d</color>
    </branding>
```

But you would also need to handle
```
    <branding>
      <color type="primary" scheme_preference="light">#ff00ff</color>
      <color type="primary">#993d3d</color>
    </branding>
```

And this one, that's probably already working
```
    <branding>
      <color type="primary">#993d3d</color>
    </branding>
```

If this is not done, we might run into comparability problems.

For e.g. we had an app with the following (kinda nonsensical definition)

```
<branding>
    <color type="primary">#f9c440</color>
    <color type="primary" scheme_preference="light">#d48e15</color>
    <color type="primary" scheme_preference="dark">#ffe16b</color>
  </branding>
```

Presumably cause they wanted it to work in flathub and appcenter.

https://github.com/ryonakano/konbucase/pull/128#issuecomment-2034641821

### Proposal

Parse all branding colors from appstream including the `scheme_preference`

### Prior Art (Optional)

_No response_

edwood-grant:
### Short summary
Aims to fix #2140 

This will parse all the branding possibilities of the metadata info according to the freedesktop docs [here](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-branding).

The branding tag has a must-have color tag with `primary_color` attribute (always set its value to `primary`), with an optional `scheme_preference` attribute which shows a preference for a particular color scheme (withe light or dark) and this should take precedence.

But, as the issue says, there can be multiple setups here, and this PR aims to account for all of these in the following order:

* If the color tag has a `scheme_preference`, either set to light or dark, it should be taken as precedent, according to the desktop current color scheme (via `gtk_application_prefer_dark_theme`).
* If not, look for the color tag that has no `scheme_preference` (i.e. unknown). This was the original call.
* If the last call returns null, use the local color scheme setting.

### Tests

Apps that I tested with this change:

* [Reco](https://appcenter.elementary.io/com.github.ryonakano.reco/): Has all color branding setups (light, dark and without `scheme_preference` attribute). Different colors per tag.
* [KonbuCase](https://appcenter.elementary.io/com.github.ryonakano.konbucase/): Only has light and dark branding tags. Different colors per tag.
* [Larawan](https://appcenter.elementary.io/io.github.xchan14.larawan/): I _believe_ it has no branding tag of any kind.
* [Mail](https://appcenter.elementary.io/io.elementary.mail/): Has all tags, but all with the same color.

I couldn't find an application that had only the no `scheme_preference` attribute, but it was the default original call, so it should cover that case as well.

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -755,13 +755,13 @@ public class AppCenterCore.Package : Object {
             var branding = component.get_branding ();
             if (branding != null) {
                 var gtk_settings = Gtk.Settings.get_default ();
-                if (gtk_settings.gtk_application_prefer_dark_theme){
+                if (gtk_settings.gtk_application_prefer_dark_theme) {
                     color_primary = branding.get_color (AppStream.ColorKind.PRIMARY, AppStream.ColorSchemeKind.DARK);
                 } else {
                     color_primary = branding.get_color (AppStream.ColorKind.PRIMARY, AppStream.ColorSchemeKind.LIGHT);
                 }
 
-                if (color_primary == null){
+                if (color_primary == null) {
                     color_primary = branding.get_color (AppStream.ColorKind.PRIMARY, AppStream.ColorSchemeKind.UNKNOWN);
                 }
             }
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -754,7 +754,16 @@ public class AppCenterCore.Package : Object {
         } else {
             var branding = component.get_branding ();
             if (branding != null) {
-                color_primary = branding.get_color (AppStream.ColorKind.PRIMARY, AppStream.ColorSchemeKind.UNKNOWN);
+                var gtk_settings = Gtk.Settings.get_default ();
+                if (gtk_settings.gtk_application_prefer_dark_theme){
+                    color_primary = branding.get_color (AppStream.ColorKind.PRIMARY, AppStream.ColorSchemeKind.DARK);
+                } else {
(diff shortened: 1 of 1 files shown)


danirabbit (changes requested):
Thanks for your branch! This looks really good besides one problem: it caches the brand color so if you change between light and dark mode you get the old color even if you exit and re-enter the page.

I think the way to fix this would be to cache each color separately and then only check if its null after we figure out which color we're looking for

edwood-grant:
Got it!

So what I did now was to create a separate color values for each primary color, that is:

* Primary light
* Primary Dark
* Primary Unknown (the one without any scheme preferences)
* Fallback (when everything fails, the settings appcenter primary)

This is encapsulated in a private method, which runs before asking for available colors. This should guarantee it to work every time you switch dark mode and then enter/exit the page.

Hopefully this works better! :+1:

danirabbit (approved):
Works here!

PR #2227, opened 2024-12-10
 Made issue_box widget a grid (elementary#2224)

### LINKED ISSUE #2224: Visual bug in release widgets
### What Happened?

image
image

### Steps to Reproduce

Open App Detail

### Expected Behavior

-

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

edwood-grant:
## Short summary

Aims to fix #2224

The main issue was that the Issue box `LinkedButton` Label widget broke the natural sizing within the box when `wrap` was set to true. I'm not sure why this is the case, though. 

I could have removed the wrapping, but it would let issues too long and the boxes looked too odd. However, I managed to fix this by changing the `issue_boxes` GTK Widget to a Grid. This does detect and respect the natural size height changes when enabling wrap within a `LinkedButton` Label. 

This PR also contains three small additions:

* Changed the alignment of the cute bug icon so it centers better with the corresponding `LinkedButton` label (set `valign` to `Gtk.Align.BASELINE_CENTER`).
* Added a `margin_top` value on the "Fixed Issues" header to separate better the issues section from the rest.
* Explicitly set wrapping values via `max_width_chars` to the `LinkedButton` Label and the description label to avoid having too long descriptions and issues that made all the boxes too long. 
    * This is probably the part that might not go well with the whole design of the page? Please, let me know if these changes make sense, or if only enable wrapping should be enabled but not set any wrap max length. I did this i particular for the videos app that ha s particularly long one line description, and it looked odd in my view.

## Tests

Here are the applications that I tested with this change:

#### Before: 

Calendar Before

#### After:
 
Calendar After

##### Before: 

Mail Before

#### After:
 
Mail After

##### Before: 

Screenshot Before

#### After:
 
Screenshot After

### DIFF
--- a/src/Widgets/ReleaseRow.vala
+++ b/src/Widgets/ReleaseRow.vala
@@ -45,6 +45,7 @@ public class AppCenter.Widgets.ReleaseRow : Gtk.Box {
         var description_label = new Gtk.Label (format_release_description (release.get_description ())) {
             selectable = true,
             use_markup = true,
+            max_width_chars = 55,
             wrap = true,
             xalign = 0
         };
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Looks good! nice work :tada:

PR #2234, opened 2024-12-17
Put the release notes description label inside a Gtk.Grid

### LINKED ISSUE #2233: More visual bugs in the release notes widgets
### What Happened?

After doing a fix for #2224 it seems that here are more visual bugs in specific descriptions i the release carrousel widget.

Here are some screenshots as an example:

Here is for example Nimbus:
Nimbus release widget

Journaler:
Journaler release widget

And Reco:
Reco release widget

The reason is that in the last change I introduced the description to wrap and and to have a character wrap limit. And as the previous bug showed, wrapping labels in a `GTk.Box` will not respect natural heights for some reason. But setting them inside a `Gtk.Grid` will.

I'll submit a PR soon, but I set the issue to mark it as existent and refer to it on the PR.

### Steps to Reproduce

1. Open AppCenter
2. Go to any application that has a long description
3. Watch the releases carrousel

### Expected Behavior

The release description should not overflow the carrousel widget.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

edwood-grant:
## Short summary 

This fixes #2233 

This is the same issue that we had for issues, but now for the description label itself. It seems that in a `Gtk.Box` with wrapped labels will not do its height correctly, if there are grids abound.

Putting the label inside a `Gtk.Grid` will make it also respect the natural height and will fix another set of problems found in other description labels.

~~Another change in this PR was a small refactor to compartmentalize each widget creation in its own method. This is just for organization and refactoring purposes only.~~ 
_Edit: Having methods within the constructor is not preferred, so this was set back to how it was before._

## Tests

Here are tests made on the notified app issues:

#### Before: 
Nimbus Before
#### After:
Nimbus After

#### Before: 
Journaler Before
#### After:
Journaler After

#### Before: 
Reco Before
#### After:
Reco After

### DIFF
--- a/src/Widgets/ReleaseRow.vala
+++ b/src/Widgets/ReleaseRow.vala
@@ -69,7 +69,7 @@ public class AppCenter.Widgets.ReleaseRow : Gtk.Box {
             use_markup = true,
             max_width_chars = 55,
             wrap = true,
-            xalign = 0,
+            xalign = 0
         };
         description_label.add_css_class (Granite.STYLE_CLASS_H3_LABEL);
 
--- a/src/Widgets/ReleaseRow.vala
+++ b/src/Widgets/ReleaseRow.vala
@@ -33,12 +33,6 @@ public class AppCenter.Widgets.ReleaseRow : Gtk.Box {
         spacing = 6;
         margin_bottom = 6;
 
-        create_header ();
-        create_description ();
-        create_issues ();
-    }
-
-    private void create_header () {
         var header_icon = new Gtk.Image.from_icon_name ("tag-symbolic");
         var header_label = new Gtk.Label (format_version (release.get_version ())) {
             use_markup = true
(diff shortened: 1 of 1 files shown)


alainm23 (approved):

src/Widgets/ReleaseRow.vala
```diff
@@ -29,8 +29,17 @@ public class AppCenter.Widgets.ReleaseRow : Gtk.Box {
...
 
     construct {
-        var header_icon = new Gtk.Image.from_icon_name ("tag-symbolic");
+        orientation = Gtk.Orientation.VERTICAL;
+        spacing = 6;
+        margin_bottom = 6;
+
+        create_header ();
+        create_description ();
+        create_issues ();
```
lenemter, line 38: We prefer to to not split the constructor into multiple methods.

lenemter (changes requested):

src/Widgets/ReleaseRow.vala
edwood-grant: Got it :+1: Just set removed these methods and put it back to the constructor

alainm23 (approved):

Issue #529, opened 2017-11-07
AppCenter sometimes opens to the Updates view


cassidyjames:
When opening AppCenter by clicking its icon in the dock, it's now opening to the Updates view instead of the Home view.

If I manually kill AppCenter, it doesn't continue doing this. I will try to figure out what causes it, but I'm seeing it in AppCenter on elementary OS and in Pop!_Shop on Pop!_OS.

cassidyjames:
It seems like you can reproduce it like this:

1. Log into your OS, 
2. Click an update notification, opening the app
3. Close the app

Now all subsequent openings of AppCenter open to the updates view, even after running the updates. Killing AppCenter and re-launching seems to temporarily fix it so it opens to the home view again.

edwood-grant:
Should be fixed by #2237

PR #2237, opened 2024-12-22
Various fixes in AppListUpdateView

### LINKED ISSUE #2215: Appcenter still reserves space for system updates in eOS 8
### What Happened?

Appcenter, when maximized fullscreen or on the side, display a large blank space before the apps - for where system updates once were.

Bildschirmfoto von 2024-10-30 20 47 14

### Steps to Reproduce

open appcenter
maximize, or tile to one side

### Expected Behavior

For the list of apps to start at the top, with no strange empty wasted space

### OS Version

8.x (Early Access)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

edwood-grant:
## Summary

This PR make several fixes around the `AppListUpdateView`, and the general Update packages behavior, namely:

1. Fixes #2215 _AppCenter still reserves space for system updates_
Bug 2215

    * The problem was related to the `vexpand` property of the `list_box`. When set to `true`
 it would try to take as much space as possible. It's now set to `false` once the installed panel is visible with any available installed packages.

2. Fixes various issues around not showing properly the updates label.
    * Those are the _"Everything is up-to-date."_ and _"Updates available"_ labels on the top. They should now show/hide properly according to the update behavior.

4. Disable the "Check for updates" menu button (set sensitive to `false`) when clicking on it, and enable it when the Flatpak backend stops again. This is to stop the user for clicking multiple times (when the button is enabled for brief moments in the check for updates process), which would create duplicate items in the updates view. This fix should also disable it if you are updating any package in the updates list box.
Check for updates menu button bug

5. Fixes a bug where if you clicked on an application to update that had no description available (and thus no release notes button) and then cancel it immediately, it would show the release notes button.

https://github.com/user-attachments/assets/7d8e7593-21c0-49b4-8bd3-4460c04a3b31

6. Fixes issue #529 that the application would get "stuck" in the `AppListUpdateView` page and would not open to the main menu until terminated. This also fixes a problem where the view would not show the installed apps if you let the backend check for updates without changing to the updates view until the backend finished.

## Tests

Here is a video of the whole flow with all the fixes applied:

https://github.com/user-attachments/assets/22efbfa5-442f-4927-8c39-b8adb255f3a0

## Comments

I know this is might be a moderate (maybe big?) PR, and I may be doing things wrong here on some places, especially since I don't know the whole picture. I hope it's still good work though to at least keep working on it, or maybe getting closer to something good.

Also, if you'd rather have this split into multiple PRs, or maybe it's not good enough, let me know so I can make it more streamlined or make the necessary changes for everyone.

Thanks a lot,

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -225,6 +225,7 @@ public class AppCenter.App : Gtk.Application {
 
         if (show_updates) {
             ((MainWindow) active_window).go_to_installed ();
+            show_updates = false;
         }
 
         active_window.present ();
--- a/src/Core/UpdateManager.vala
+++ b/src/Core/UpdateManager.vala
@@ -227,6 +227,7 @@ public class AppCenterCore.UpdateManager : Object {
     }
 
     public async void update_cache (bool force = false) {
+        updates_liststore.remove_all ();
         cancellable.reset ();
 
         if (Utils.is_running_in_demo_mode () || Utils.is_running_in_guest_session ()) {
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -145,6 +145,14 @@ public class AppCenter.MainWindow : Gtk.ApplicationWindow {
         backend.notify["job-type"].connect (update_overlaybar_label);
 
         overlaybar.label = backend.job_type.to_string ();
+
+        if (installed_view == null) {
+            installed_view = new Views.AppListUpdateView ();
+
+            installed_view.show_app.connect ((package) => {
+                show_package (package);
+            });
+        }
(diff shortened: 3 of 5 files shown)


leonardo-lemos (approved):
Nice work, code looks good and fixes the issues. Thanks!

danirabbit:
Generally in the future it would be nice to have at least separate commits and do a rebase merge so that if we have to do a git blame later it would be easier for future maintainers

Thank you for resolving these issues!

Issue #721, opened 2018-06-04
daemon locks up apt and hogs network immediately upon log in


codygarver:
appcenter-daemon has no launch delay or any consideration for if it's a good time to check for updates. I don't want my web browser to compete with apt for network resources as soon if I start surfing immediately. I don't want to wait for appcenter-daemon to release apt if I jump into a Terminal and start trying to using apt.

Possible solutions are AppCenter could remember the last time it checked for updates OR some other kind of delay.

mmstick:
Using a timestamp would be a good idea, I think. Perhaps only check once a day?

davidmhewitt:
The daemon now autostarts 60 seconds after the rest of the session and only refreshes its cache if it's >24 hours since it last did it.

Issue #495, opened 2017-10-04
Appcenter punishes disk IO when pantheon starts


Nine-H:
My notebook starts pantheon _significantly_ faster with appcenter disabled in xdg-autostart. it also stays cooler and is more responsive. it's a sandy bridge machine with a HDD.

Appcenter wants to start a daemon and update apt sources immediately at login, and this stresses disk and network while other disk intensive processes are starting and other software and services want to access the network. This makes loki feel _really_ bad on older hardware (I'm not upgrading my notebook: all I do is edit text files and google docs).

There's a few possible ways to solve this:

- kill itself if the machine has a battery
- allow appcenter autostart ("Check for updates") to be disabled in swichboard/apps
- automatically not even try to run if the system is a potato
- wait until everything else is loaded
- only run and update package lists on demand
- all of the above

matzipan:
I think a sort of deferred loading would be a good solution.

peteruithoven:
"potato"?

Nine-H:
I don't know what the industry term is but I think American techies say "toaster". legacy, or budget systems that use low spec drives, weaker chipsets, slower ram, integrated graphics or low clockspeeds.

bksubhuti:
I don't like it either.

davidmhewitt:
AppCenter was set to load 60 seconds after the rest of the desktop environment in this commit:
https://github.com/elementary/appcenter/commit/a3c96a9ba8557e013016f44dae44926ec916b0ef

There are also other issues open about not running the update check on every start (which currently happens) which would also help with this.

I'll close this as I think the delay already helps with a significant portion of this issue.

PR #2253, opened 2025-02-05
Delay initial check for updates by default 60 seconds

vjr:
Related to:

https://github.com/elementary/settings-daemon/issues/158

and

https://github.com/elementary/settings-daemon/pull/177

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -192,18 +192,16 @@ public class AppCenter.App : Gtk.Application {
             request_background.begin ();
 
             NetworkMonitor.get_default ().network_changed.connect ((available) => {
-                schedule_cache_update (!available);
+                schedule_cache_update (!available, false);
             });
 
-            // Don't force a cache refresh for the silent daemon, it'll run if it was >24 hours since the last one
-            update_manager.update_cache.begin (false);
             silent = false;
             return;
         }
 
         if (active_window == null) {
             // Force a Flatpak cache refresh when the window is created, so we get new apps
-            update_manager.update_cache.begin (true);
+            schedule_cache_update (false, true);
 
             var main_window = new MainWindow (this);
             add_window (main_window);
@@ -295,7 +293,7 @@ public class AppCenter.App : Gtk.Application {
     }
 
     private uint cache_update_timeout_id = 0;
-    private void schedule_cache_update (bool cancel = false) {
+    private void schedule_cache_update (bool cancel = false, bool force = false) {
         unowned var update_manager = AppCenterCore.UpdateManager.get_default ();
 
         if (cache_update_timeout_id > 0) {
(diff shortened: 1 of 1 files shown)


danirabbit:
AppCenter no longer uses PackageKit so I'm not sure this is still relevant here

vjr:
> AppCenter no longer uses PackageKit so I'm not sure this is still relevant here

The intent is to delay the flatpak check too - does this PR still make sense?

danirabbit:
I'm not sure what problem you're trying to solve here with the Flatpak side of things. It seems like you should either turn on automatic updates in appcenter or disallow appcenter from auto starting so you can run Flatpak updates manually in terminal. But I'm generally not a fan of adding magic number timeouts to things

vjr:
There's already a `SECONDS_AFTER_NETWORK_UP` constant in here - not intending to add a magic number.

Wanted to (re) solve the (I guess) regression mentioned in https://github.com/elementary/settings-daemon/issues/158 which are the old issues https://github.com/elementary/appcenter/issues/495 and https://github.com/elementary/appcenter/issues/721 which is to not hog the system/network as soon as you startup/login.

Let me know if this is sufficient rationale for this PR - otherwise no worries - we can close/drop this one :-)

danirabbit:
I think in that case it would be better to do something like check for available resources/user activity instead of using a timeout

vjr:
> I think in that case it would be better to do something like check for available resources/user activity instead of using a timeout

argh! that sound like a mini-project on its own lol. can that be a feature/wishlist item for a later time?

just a reminder this is just bringing back the [previously accepted workaround](https://github.com/elementary/appcenter/commit/a3c96a9ba8557e013016f44dae44926ec916b0ef) mentioned in [this comment](https://github.com/elementary/appcenter/issues/495#issuecomment-614545582) but in the code instead of that old config key - im not sure if the old way is still valid.

alainm23 (approved):

leonardo-lemos:
> I think in that case it would be better to do something like check for available resources/user activity instead of using a timeout

@danirabbit Can we go ahead and merge this for now, and leave that feature for a future implementation? I see this as a quick win, while implementing a full user activity monitor to trigger updates would be significantly more complex.

leonardo-lemos:
> I'm personally against adding a fixed timeout here. In practice this just ends up with all our services running at the same time but with a delay after startup. Another solution would be using a systemd unit. Systemd has a mechanism for staggering scheduled events built in

@vjr I agree with Dani's suggestion — using Systemd should be straightforward, as it only requires setting up a timer and allowing AppCenter to run the update check via the command line.

vjr:
thanks for the ideas but i might not be able to dedicate time to work on the systemd approach, maybe someone else like @ryonakano could try especially since they're already working on https://github.com/elementary/settings-daemon/pull/200 ?

PR #2258, opened 2025-02-20
Release 8.1.0

danirabbit:


zeebok:
I did run into this where the title of the app isn't visible, not sure if it was from being run in a VM, after install the new release version. A fresh ISO install doesn't have this.

image

If you don't see this then I may not worry about it

danirabbit:
@zeebok fixed :) https://github.com/elementary/appcenter/pull/2262

zeebok (approved):

PR #2263, opened 2025-03-04
SearchView: use Gtk.ListView

danirabbit:
First step towards #2179 

* Use Gtk.ListView
* Create a new, clean SearchListItem widget so we don't have to modify a bunch of other abstract widgets and we can be more free with designing search results in the future

Screenshot from 2025-03-04 11 14 41


src/Views/SearchView.vala
```diff
@@ -91,12 +100,22 @@ public class AppCenter.SearchView : Adw.NavigationPage {
...
+            list_item.child = new SearchListItem ();
+        });
+
+        factory.bind.connect ((obj) => {
+            var list_item = (Gtk.ListItem) obj;
+            ((SearchListItem) list_item.child).package = (AppCenterCore.Package) list_item.item;
+        });
+
+        list_view.activate.connect ((index) => {
+            show_app ((AppCenterCore.Package) list_store.get_item (index));
```
leolost2605, line 114: Probably better to use the selection model here. Currently not a problem but when we use filtermodels and sorter models etc the original store doesn't match with the index anymore :)

leolost2605 (approved):
Sorry I kinda forgot about that 😬
Haven't tested but this looks good thank you! 
I just added a minor comment :)

src/Views/SearchView.vala
danirabbit: Good call!

danirabbit:
Yeah I forgot about it too! I was going through issue reports and there's a few about search performance so wanted to follow up :sweat_smile:

PR #2271, opened 2025-03-05
SearchListItem: show category

danirabbit:
Show a category icon and label in search results:

Screenshot from 2025-03-05 14 45 23


danirabbit:
Making this a draft because I want to re-use the category data that's current in HomePage

PR #2272, opened 2025-03-06
Show screenshots in search list

danirabbit:
Screenshot from 2025-09-05 08 18 14


danirabbit:
Screenshot from 2025-03-05 16 56 18

GridView is totally interesting here as well

danirabbit:
Extremely satisfying with accent colors

Screenshot from 2025-03-05 17 14 13

jeremypw (changes requested):
The screenshot is often not the correct one.  Also makes list visually jumpy.

jeremypw:
@danirabbit I ran the search "term" and I noticed that many (50%?) of the screenshots did not correspond to the screenshots shown in the app details page and appeared to be of a different app entirely.  For example the "Blackbox" app.

src/Widgets/SearchListItem.vala
```diff
@@ -43,17 +55,54 @@ public class AppCenter.SearchListItem : Gtk.Grid {
...
+    private void load_screenshot (AppCenterCore.Package package) {
+        screenshot_picture.visible = false;
+
+        screenshot_picture.set_branding (package);
+
+        string? screenshot_url = null;
+
+        var screenshots = package.get_screenshots ();
+        foreach (unowned var screenshot in screenshots) {
+            screenshot_url = screenshot.get_image (-1, SCREENSHOT_HEIGHT, scale_factor).get_url ();
```
jeremypw, line 84: Need null check for `screenshot` and `screenshot.get_image() else appcenter crashes for some searches.

jeremypw:
@danirabbit I found this to crash on search now, due to missing null checks (see note).  I am not sure if you are interested in pursuing this but I am not sure it adds much. Loading the screenshots is slow, moreover restarts for every change in search term.  This could be amerliorated by only loading screenshots that will actually be visible in the same way Files does and having similar size placeholder images so the display does not jump about.  But as the screenshots are easily accessible by clicking on those items that look interesting it hardly seems worth the effort.

zeebok:
> @danirabbit I found this to crash on search now, due to missing null checks (see note). I am not sure if you are interested in pursuing this but I am not sure it adds much. Loading the screenshots is slow, moreover restarts for every change in search term. This could be amerliorated by only loading screenshots that will actually be visible in the same way Files does and having similar size placeholder images so the display does not jump about. But as the screenshots are easily accessible by clicking on those items that look interesting it hardly seems worth the effort.

I like the idea of being able to "window shop" an app right from the search list and this concept should be explored more but I do think if there are performance issues having some kind of lazy loading should be put in before merging this. The performance should be the same or better than it is now.

PR #2274, opened 2025-03-06
Add label for in-app purchases

### LINKED ISSUE #2273: Show info about in-app payments
### Problem

We currently don’t warn about in-app payments anywhere 

### Proposal

Show a message when `money-purchasing` is `intense`

### Prior Art (Optional)

In the iOS App Store the line “in-app purchases” is shown below the action button. This association makes sense to me because an app can claim to be free but have an in-app subscription or be a free game that has paid elements etc 

Image

danirabbit:
Fixes #2273 

Screenshot from 2025-03-06 08 39 40

### DIFF
--- a/data/appcenter.metainfo.xml.in
+++ b/data/appcenter.metainfo.xml.in
@@ -65,6 +65,18 @@
   <update_contact>contact_at_elementary.io</update_contact>
 
   <releases>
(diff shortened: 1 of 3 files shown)


zeebok:
Should there also be some kind of badge in the app detail view?

zeebok (approved):

PR #2275, opened 2025-03-06
CategoryView: use listmodels

danirabbit:
First step towards replacing FlowBox with more efficient GridView is to use ListModels


leolost2605 (approved):
LGTM

PR #2287, opened 2025-04-16
Move get_screenshots () to Package

danirabbit:
Move this to Package so we can re-use it in other views such as https://github.com/elementary/appcenter/pull/2272

### DIFF
--- a/src/Core/Package.vala
+++ b/src/Core/Package.vala
@@ -898,6 +898,68 @@ public class AppCenterCore.Package : Object {
         return null;
     }
 
+
+    public GenericArray<AppStream.Screenshot> get_screenshots () {
+        bool has_matching_environment = false;
+        bool has_matching_style = false;
+        var desktop_environment = Environment.get_variable ("XDG_SESSION_DESKTOP");
+        var prefer_dark_style = Gtk.Settings.get_default ().gtk_application_prefer_dark_theme;
+        var desktop_style = prefer_dark_style? AppStream.ColorSchemeKind.DARK.to_string () : AppStream.ColorSchemeKind.LIGHT.to_string ();
+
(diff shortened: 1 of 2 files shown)


src/Core/Package.vala
```diff
@@ -898,6 +898,68 @@ public class AppCenterCore.Package : Object {
         return null;
     }
 
+
```
ryonakano, line 901:
```suggestion
```

src/Core/Package.vala
```diff
@@ -898,6 +898,67 @@ public class AppCenterCore.Package : Object {
...
+            }
+        }
+
+        var screenshots = new GenericArray<AppStream.Screenshot> ();
+
+        foreach (unowned var screenshot in all_screenshots) {
+            var environment_id = screenshot.get_environment ();
+            if (environment_id == null) {
+                screenshots.add (screenshot);
+                break;
```
ryonakano, line 937: I guess this should be `continue;` instead, otherwise only the first screenshot is added?

src/Core/Package.vala
```diff
@@ -898,6 +898,67 @@ public class AppCenterCore.Package : Object {
...
+
+            if (same_environment && same_style) {
+                screenshots.add (screenshot);
+            } else if (same_environment && !same_style && !has_matching_style) {
+                screenshots.add (screenshot);
+            } else if (!same_environment && same_style && !has_matching_environment) {
+                screenshots.add (screenshot);
+            } else if (!has_matching_environment && !has_matching_style) {
+                screenshots.add (screenshot);
+                break;
```
ryonakano, line 955: This should also be `continue;` if you don't mean to change the meaning of the code with this PR. The previous code uses `GenericArray.@foreach ()` and returning in the function given as the argument is equivalent to call `continue;` in the `foreach (var elem in array)` statement.

PR #2291, opened 2025-05-10
Remove packagekit from the dependency chain

tintou:
We then do not need the AppStream data either

Closes: https://github.com/elementary/default-settings/issues/255


danirabbit (approved):
Oh yup we don't need any of this shit

PR #2293, opened 2025-05-14
AppInfoView: show when a game supports controllers

danirabbit:
Initial support for listing things an app supports, starting with controllers. Eventually this will also include things like mobile support, etc. I made this a separate list from content warnings because I think it's less important and doesn't need to be "above the fold" like content warnings do

Screenshot from 2025-05-14 13 15 56

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -272,6 +272,22 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
             maximum_size = MAX_WIDTH
         };
 
+        var supports_flowbox = new Gtk.FlowBox () {
+            column_spacing = 24,
+            row_spacing = 24,
+            selection_mode = NONE
+        };
+        supports_flowbox.add_css_class ("content-warning-box");
+
+        var supports_flowbox_revealer = new Gtk.Revealer () {
+            child = supports_flowbox
+        };
+
+        var supports_clamp = new Adw.Clamp () {
+            child = supports_flowbox_revealer,
+            maximum_size = MAX_WIDTH
+        };
+
         if (!package.is_runtime_updates) {
 #if CURATED
             if (package.is_native) {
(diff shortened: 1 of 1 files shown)


src/Views/AppInfoView.vala
```diff
@@ -472,6 +488,43 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
...
+        var supports = package_component.get_supports ();
+        for (int i = 0; i < supports.length; i++) {
+            var relation = supports[i];
+
+            string title, description, icon_name;
+
+            switch (relation.get_value_control_kind ()) {
+                // An input method for users to control software
+                case GAMEPAD:
+                    switch (relation.get_kind ()) {
```
leonardo-lemos, line 500: I personally don't like a switch inside a case, I think all the gamepad kind logic can be inside a method, then the code will be more organized and prepared to handle more kinds in the future.

src/Views/AppInfoView.vala
```diff
@@ -472,6 +488,43 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
...
+                            title = _("Play With a Controller");
+                    }
+
+                    description = _("Such as a PlayStation, Xbox, or 8BitDo controller.");
+                    icon_name = "input-gaming-symbolic";
+                    break;
+
+                default:
+                    // unhandled
+                    debug (relation.get_value_control_kind ().to_string ());
```
leonardo-lemos, line 516: Can you give more context about this log? Maybe log a phrase like: "Unhandled kind: %s, %s".

leonardo-lemos (changes requested):

src/Views/AppInfoView.vala
danirabbit: In this case "kind" just means "required", "recommended", or "supported" so I don't think there will be any more kinds in the future

danirabbit: Ah sorry I misunderstood. Hows that? :)

leonardo-lemos: Cool, LGTM

leonardo-lemos (approved):

PR #2305, opened 2025-07-30
Use filter list models for updates

### LINKED ISSUE #1822: Remove pending updates from "Installed" tab
### Uninstall applications is slow
If I want to uninstall an application from the **AppCenter**, I need to wait every time that updates are pulled and that the **_Installed_** tab shows them.
Even if I could just remove an application from the **App Menu**, it's not always the best case if I have lots of manually installed applications showing up. So it's useful to have a quick look at the applications installed through the **AppCenter**.

### Adding a new Updates tab
By adding a new tab in the **AppCenter** to just display updates, it would make the **_Installed_** tab faster to load by just displaying the installed applications list. It just make sense to have a separate space where the user can take care of _updates_, but not as separate as in Ubuntu (with dedicated applications) where it starts to be confusing.

### Prior Art (Optional)

_No response_

### LINKED ISSUE #2108: "1 Update Available" while no possible updates are displayed
### What Happened?

"1 Update Available" is displayed in the upper left corner and there is a "1" over the hamburger. There is no listing of packages that need updating. The "Update All" button is green. When pressed it is "checking for updates" and makes no (new) listing, while "1 Update Available" remains. Hamburger refresh does not help

apt update and apt upgrade report no update and no error. Various sudo apt update/upgrade/clear/autoremove commands present nothing (no updates, no error). Purging the package with io.elementary.appcenter and reinstalling it does not help. "dconf reset -f /io/elementary/appcenter/" does not help. "rm -rf ~/.cache/io.elementary.appcenter" does not help. "sudo apt --fix-broken install" does not help. Reboot does not help.

### Steps to Reproduce

1. Start application (io.elementary.appcenter)
2. Click hamburger and "Updates & Installed Software"
3. "1 update Available" shown in upper left corner. "1" is shown over the hamburger.

### Expected Behavior

1. Either there should be no available updates.
2. Or it should display the updates available.
3. Or it should better report what is wrong (in the system) and where it gets the wrong information from.

### OS Version

Other Linux

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2177: Opening AppCenter when updates have already been found has odd UI
### What Happened?

I logged in and was informed there were updates available. When I click the notification to open AppCenter I immediately see this: 
Screenshot from 2024-06-19 00 34 45

Once installed apps are found I see this:
Screenshot from 2024-06-19 00 34 47

And then it finishes looking for updates it starts revealing the same apps:
Screenshot from 2024-06-19 00 34 48

### Steps to Reproduce

1. Let the daemon find apps in the background
2. Open AppCenter (in any way)
3. Go to the Update View if it doesn't open on that view

### Expected Behavior

I can see two possible behaviors:

1. It immediately shows the apps that can be updated, which makes sense given the context, showing something for it loading installed apps, and then not show the placeholder in the updates section. If there are new updates, update the list.
2. It does the same flow as when it doesn't have updates already loaded up

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #2277: Newly installed apps show up as updates
### What Happened?

When installing a new app the app appears under available updates even though it has just been freshly installed so there should not be any updates. If I then click on update all the app just disappears from available updates and nothing else happens.

### Steps to Reproduce

1. Open appcenter.
2. Install any app.
3. Wait for it to have finished installing.
4. An available update will show up for that app.

### Expected Behavior

An app that has just been installed should not show up as available updates because it is already on the most recent version.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

### LINKED ISSUE #2297: Update list repeated entries multiple times
### What Happened?

On the update screen, each app in the list "updates available" was displayed four times in row.

Image

### Steps to Reproduce

1. Open AppCenter
2. Go to updates screen
3. Request a few apps to be updated, not waiting for previous one to finish.

### Expected Behavior

The list shows each app once.

### OS Version

8.x (Circe)

### Session Type

Classic Session (X11, This is the default)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
This is the first of a few PRs i want to propose to move appcenter more towards listmodels and property bindings instead of manual signal connections and always querying the actual flatpak backend for gee.collections.
The biggest advantage of that is consistency. Manually adding things to stores, connecting to signals and updating stuff when they fire is bound to lead to conflicts and forgetting to update a thing here and a thing there. This is especially the case in a code base as big as the appcenter where one might not have an overview about who exactly is connected to what.
Instead we have a single source of truth and can obtain multiple views of that using gtk's wonderful listmodels.

Another thing are performance improvements. Since we always have a list of installed apps we don't need to query the backend so one job less to do when getting updates. Also we immediately have a list of installed apps available without any work from our side so we can show it immediately. When updating apps we don't need to tell the update manager to refetch updates because an app was updated so it shouldn't be shown in the updates anymore. Instead it will just be immediately filtered out.

Fixes #2277 
Fixes #2297 
Fixes #2177
Fixes #2108 
Fixes #1822 

Fixes displayed update size sometimes being really big


src/Core/Package.vala
```diff
@@ -613,20 +607,21 @@ public class AppCenterCore.Package : Object {
         return cached_search_score;
     }
 
-    private string? name = null;
-    public string? get_name () {
-        if (name != null) {
-            return name;
-        }
+    private string? _name = null;
+    public string name {
```
danirabbit, line 611: Can we break out this change into another PR? It's a lot of noise in this PR and probably worth doing as a separate commit

src/Core/Package.vala
```diff
@@ -450,29 +450,19 @@ public class AppCenterCore.Package : Object {
...
-    public async bool update (bool refresh_updates_after = true) throws GLib.Error {
+    public async bool update () throws GLib.Error {
         if (state != State.UPDATE_AVAILABLE) {
             return false;
         }
 
-        var success = yield perform_operation (State.UPDATING, State.INSTALLED, State.UPDATE_AVAILABLE);
-        if (success && refresh_updates_after) {
-            yield UpdateManager.get_default ().get_updates ();
-        }
```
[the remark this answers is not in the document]
leolost2605: this relies on changes in this branch because via the listmodels we automatically filter it from the updates once it changes to installed from updating whereas on main this would require some additional changes to the updatemanager which we would just remove right away :shrug:

leolost2605:
This should be ready for review once #2307 gets merged :)

leolost2605:
Don't worry they weren't that bad :sweat_smile: 

This should be ready for review and testing now :)

src/Views/AppListUpdateView.vala
```diff
@@ -227,21 +213,14 @@ namespace AppCenter.Views {
...
-                on_updates_changed ();
-            }
+            flatpak_backend.updatable_packages.items_changed.connect (on_updates_changed);
+            on_updates_changed ();
 
             flatpak_backend.notify ["working"].connect (() => {
                 if (flatpak_backend.working) {
                     refresh_menuitem.sensitive = false;
-                    header_revealer.reveal_child = false;
-                    updated_revealer.reveal_child = false;
```
danirabbit, line 239: I think this might be a small regression. If we're in the process of checking for updates, we shouldn't be reporting that everything is already up to date

src/Views/Homepage.vala
```diff
@@ -138,6 +140,13 @@ public class AppCenter.Homepage : Adw.NavigationPage {
 
         updates_badge = new Gtk.Label ("!");
         updates_badge.add_css_class (Granite.STYLE_CLASS_BADGE);
+        fp_client.bind_property (
```
danirabbit, line 143: Feel free to say "no" if I'm being unreasonable, but I wonder if we should also break out this change to using property bindings here and in AppListUpdateView into a separate commit.

I'm just really nervous about introducing regressions since this part of the codebase is really complicated and historically has been really fragile 😅

danirabbit (commented):
I'm feeling a bit overwhelmed trying to do a good review here, but this is genuinely good stuff that I'm very excited about. Thank you for tackling this!

src/Views/Homepage.vala
leolost2605: I opened #2308 for the n updatable and has updatable bindings. There are only two other ones which are one liners and require a bunch more work without the listmodels so I left them for here

leolost2605:
I've added a comment above the listmodel creation now, idk if that helps or if it's more confusing, we can remove it :sweat_smile:

danirabbit (approved):
As far as I can tell everything is working as expected and all the changes make sense to me! 🚀

PR #2309, opened 2025-08-01
Introduce can update all property, add bindings

leolost2605:


src/Views/AppListUpdateView.vala
```diff
@@ -79,10 +77,11 @@ namespace AppCenter.Views {
                 "up-to-date", updated_revealer, "reveal-child", SYNC_CREATE
             );
 
-            update_all_button = new Gtk.Button.with_label (_("Update All")) {
+            var update_all_button = new Gtk.Button.with_label (_("Update All")) {
                 valign = Gtk.Align.CENTER
             };
             update_all_button.add_css_class (Granite.STYLE_CLASS_SUGGESTED_ACTION);
+            update_manager.bind_property ("can-update-all", update_all_button, "sensitive", SYNC_CREATE);
```
danirabbit, line 84: I wonder if maybe this shouldn't be an action so we can set the action as disabled and sensitivity will be handled automatically that way?

leolost2605: I switched it to an action, it's much cleaner now :)

danirabbit (approved):
Very nice. This will let us add an update all button to update notifications in the future as well 🚀

Issue #2214, opened 2024-09-29
No cache / time of last update not taken into account


Letterus:
### What Happened?

AppCenter looks for updates too often. That leads to an annyoing experience.

### Steps to Reproduce

1. Start AppCenter. Overview is showing the "Check for updates" spinner.
2. Wait for the task to complete.
3. Click the "update" icon: AppCenter starts to check for updates again, don't showing me the results of the last check. So I have to wait again.

This may be repeated for every start of the app.

### Expected Behavior

Only check for updates once upon a time and cache that result for 2 to 24 hours. Suggestions: 24h on stable, 8h on RC, even less (2h?) on daily.

### OS Version

8.x (Early Access)

### Session Type

Secure Session (Wayland)

### Software Version

Latest release (I have run all updates)

leolost2605:
The reproduction steps in this issue have already been fixed in #2305 but I'd like to expand a bit more on us checking for updates everytime the window is opened.

We currently refresh cache and check for updates everytime the app window is opened.

An important thing to keep in mind is that the user can't really do anything except for browsing before the initial jobs are completed. So the common case where you just quickly want to install an app would require you to wait for the cache refresh and for all updates to be gotten which, depending on the number of updates, can take a really long time.

I would propose that we at least drop the update check and only check for updates if the user asks us to or if the last check was more than 24 hours ago.
This might mean that the user opens the app and sees less updates than are actually available. But regarding security (how fast you get a security fix etc.) it would still behave the same as when the user has automatic updates enabled and never opens the app (which probably is the much more common case). Also the user still has the option to manually refresh and then get all updates.
Dropping the update check already cuts a big chunk of time before the user can do something

Now dropping the cache refresh is what I'm unsure about. This would mean that apps released in the last 24h (worst case) don't show up in search, featured, etc.
However the cache refresh is still a pretty expensive operation which takes quite a while where the user can't install anything.
We could just drop the cache refresh as well which IMO might be acceptable but I'd like to here other opinions on that
Or we could make sure that the cache is only refreshed if it's more than an hour old? (Which in theory is currently what is being done but we still reload the whole appstream pool which takes most(?) of the time)

@elementary/ux any opinions/suggestions?

PR #2314, opened 2025-08-03
Streamline refresh flow

leolost2605:
This PR streamlines the refresh flow. We now do everything in the update manager which will automatically start the timeout if we're less than 24 hours from the last update or else refresh immediately.
We also now only refresh after the network comes available if we tried to refresh while the network wasn't available.

This doesn't include but prepares for more behavioral changes proposed in #2214 
Also prepares for some cleanup of the refresh action enabled logic which is left to another PR

`cancel_updates` was removed because it was only used to cancel updates
1. when disabling automatic updates which isn't necessary anymore since the user now sees that the apps are updating and can cancel the update manually if they wish to
2. when the app window was closed which I don't get anyways why that was done? This could falsely cancel ongoing cache refreshes and automatic updates

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -17,8 +17,6 @@ public class AppCenter.App : Gtk.Application {
         { null }
     };
 
-    private const int SECONDS_AFTER_NETWORK_UP = 60;
-
     public static bool show_updates;
     public static bool silent;
     public static string? local_path;
@@ -211,6 +209,7 @@ public class AppCenter.App : Gtk.Application {
         }
 
         if (active_window == null) {
+            // Force a Flatpak cache refresh when the window is created, so we get new apps
             // TODO: Think about this, maybe only refresh cache?
             AppCenterCore.UpdateManager.get_default ().refresh.begin ();
 
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -211,6 +211,9 @@ public class AppCenter.App : Gtk.Application {
         }
 
         if (active_window == null) {
+            // TODO: Think about this, maybe only refresh cache?
+            AppCenterCore.UpdateManager.get_default ().refresh.begin ();
+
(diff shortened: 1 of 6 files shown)


zeebok:
I must be remembering old behavior! Just confirmed you are right. Sweet!

src/Application.vala
```diff
@@ -220,8 +211,8 @@ public class AppCenter.App : Gtk.Application {
         }
 
         if (active_window == null) {
-            // Force a Flatpak cache refresh when the window is created, so we get new apps
-            update_manager.update_cache.begin (true);
+            // TODO: Think about this, maybe only refresh cache?
```
danirabbit, line 214: I think we should leave the reasoning from the old comment here to make sure we know going forward that the reason we do this refresh is to make sure we fetch new apps when AppCenter is opened

src/Application.vala
```diff
@@ -312,27 +303,6 @@ public class AppCenter.App : Gtk.Application {
...
-        if (cache_update_timeout_id > 0) {
-            Source.remove (cache_update_timeout_id);
-            cache_update_timeout_id = 0;
-        }
-
-        if (cancel) {
-            update_manager.cancel_updates (true); // Also stops timeouts.
-            return;
-        } else {
-            cache_update_timeout_id = Timeout.add_seconds (SECONDS_AFTER_NETWORK_UP, () => {
```
danirabbit, line 328: I think we should also remove this unused constant from this class

danirabbit (approved):
This all makes sense to me. I think we probably will want to replace those timeouts with a systemd timer like we did here: https://github.com/elementary/settings-daemon/pull/200

I just have a couple of minor comments but nothing blocking. Happy to merge this. Nice cleanup :)

zeebok (approved):

PR #2306, opened 2025-07-31
Emit on_metadata_remote_preprocessed always from the main thread

leolost2605:
When we refresh the cache we run on the worker thread and would emit this signal from there.

We connect to it from the appicon though and never go to the main thread so we change gtk properties etc. from the wrong thread which would lead to crashes, sometimes a only white window, app entries dissapearing etc.

PR #2315, opened 2025-08-03
get_installed_applications: mark installed on main thread

leolost2605:
marking installed results in an update state call which in turn results in
1. a notify signal for state to which a bunch of ui components are connected which leads to incosistencies and crashes with gtk not being threadsafe
2. since the list models pr an items_changed in the list models to which a bunch of other stuff is connected with the same problems as 1

I think in general we should be much more careful with doing things like that in the worker thread (e.g. #2306 comes to mind) since I could imagine a good portion of crashes currently are due to us calling methods or emitting signals from the worker thread or not being aware they are emitted from the worker thread.

Fixes a crash that sometimes occurs when trying to start appcenter silently and probably a bunch of incosistencies when appcenter is open while we fetch installed apps (e.g. on refresh)

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -514,12 +514,16 @@ public class AppCenterCore.FlatpakBackend : Object {
         job.results_ready ();
     }
 
-    public async Gee.Collection<Package> get_installed_applications (Cancellable? cancellable = null) {
+    private async void get_installed_applications (Cancellable? cancellable = null) {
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
As far as I can tell, this makes sense.

`get_installed_apps_from_refs` is only called in `get_installed_packages_internal` which is only called in the `worker_func` and the only time `launch_job` is called with `GET_INSTALLED_PACKAGES` is here.

PR #2323, opened 2025-08-25
Add a pretty error icon instead of the broken one (Bug #2240)

### LINKED ISSUE #2240: Use Vector Art for error loading icon
### Problem

Error icons look terrible, make them look nicer

### Proposal

Use SVGs

edwood-grant:
Hey! I remember I was working on this! Whatever happened? Oh well, I finally managed to put my mind on it again.

## Error Summary

This change fixes #2240. Changing the ugly default broken icon to a nicer, dimmed default app icon (the one that has a star). 

Here is a video of the problem in action:

https://github.com/user-attachments/assets/2f640e03-1774-4461-afe8-81a1ca521c1b

This broken icon usually appears when the icon cache gets invalidated, it attempts to load the cached one, but fails, up until it finds the new update one. So its best to keep a nice icon looking here instead of the existing broken one.

Once it finds an icon, it will cross-fade smoothly to the new updated icon so the change looks smooth and nice.

## Test results

Here is a video of what it should look like when the cache gets invalidated. The best way to test this quickly is to:

1. Set the variable `MAX_APPSTREAM_AGE` to 0 (to always force cache invalidation).
2. Run the app
3. Close the app (NOT force quit, just a normal close)
4. Wait for around one-three minutes
5. Open the app again and watch the effect kicking in.

Here is a posted video of the fixed effect in action:

https://github.com/user-attachments/assets/43dd09e8-ba31-43bc-8dcd-e2d609adbe7e

Hopefully this is a small change enough to make all work. I had to make some changes in banner to support `AppIcon` because it was not using them and the banner would show broken icons because it was loading them directly instead of using `AppIcon.vala`.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -14,7 +14,7 @@ msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2025-08-22 18:28+0000\n"
-"PO-Revision-Date: 2025-05-20 10:02+0000\n"
+"PO-Revision-Date: 2025-08-26 12:55+0000\n"
 "Last-Translator: Nathan <bonnemainsnathan@gmail.com>\n"
 "Language-Team: French <https://l10n.elementaryos.org/projects/appcenter/"
 "appcenter/fr/>\n"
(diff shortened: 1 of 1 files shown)


leonardo-lemos (approved):
Nice catch!

wpkelso (approved):
Looks good to me 👍🏻

PR #2324, opened 2025-08-30
Styles: remove placeholder app icon dim

danirabbit:
We should be consistent here. If there needs to be design changes to the default icon we should do that in the icon set and not as a (expensive) CSS filter.

See also: https://github.com/elementary/icons/pull/1380


wpkelso (approved):
Makes sense to me

PR #2325, opened 2025-08-30
SearchView: use GridView

danirabbit:
When used with clamp, GridView preserves a bunch of blank space:

Screenshot from 2025-08-30 09 46 36


danirabbit:
@alice-mkh is there some trick to using GridView with a Clamp? Or should I file an issue report somewhere :sweat_smile:

danirabbit:
Yeah I think there's other things I can do instead anyways, like clamp the contents of a search item row and set max columns in the grid

leolost2605 (approved):
Can't test rn but code lgtm

PR #2339, opened 2025-09-01
ODRSProvider: Result as return value, set out parameters only when succeeds

ryonakano:
- out variables does not need to be initialized explicitly in the calling side
- Use return value instead of out parameters to check if the method succeeds
- Make sure to set values to the out parameters only when the method succeeds

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -285,8 +285,7 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
                 int64 avg;
                 uint n_ratings;
 
-                bool ret = ODRSProvider.fetch_ratings_for_app.end (res, out avg, out n_ratings);
-                if (!ret) {
+                if (!ODRSProvider.fetch_ratings_for_app.end (res, out avg, out n_ratings)) {
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


src/Views/AppInfoView.vala
```diff
@@ -282,11 +282,11 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
...
             ODRSProvider.fetch_ratings_for_app.begin (package_component.id, (obj, res) => {
-                int64 avg = -1;
-                uint n_ratings = 0;
+                int64 avg;
+                uint n_ratings;
 
-                ODRSProvider.fetch_ratings_for_app.end (res, out avg, out n_ratings);
-                if (n_ratings == 0 || avg == -1) {
+                bool ret = ODRSProvider.fetch_ratings_for_app.end (res, out avg, out n_ratings);
+                if (!ret) {
```
danirabbit, line 289: I think this can be:
```suggestion
                if (!ODRSProvider.fetch_ratings_for_app.end (res, out avg, out n_ratings)) {
```

danirabbit (approved):
Ah yeah this is better. Thank you! One small optional comment

PR #2342, opened 2025-09-03
AppListUpdateView: sort installed package by newest release

### LINKED ISSUE #36: Sort installed tab by date installed [$20]

danirabbit:
Fixes #36

* Move the newest release sorter to `Package` so we stay DRY
* Sort installed package by newest release. I don't think we can get the actual install date, so this is the next best thing

### DIFF
--- a/po/ru.po
+++ b/po/ru.po
@@ -13,8 +13,8 @@ msgstr ""
 "Project-Id-Version: Russian (Desktop)\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2025-09-02 00:08+0000\n"
-"PO-Revision-Date: 2025-08-30 01:55+0000\n"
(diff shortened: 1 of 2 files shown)


danirabbit:
@leolost2605 Typically we've added comparison functions to the class that they act on to stay in line with GLIb, which isn't like exactly the same but it seems similar enough conceptually to me to do it the same way. For example: https://valadoc.org/glib-2.0/GLib.DateTime.compare.html

danirabbit:
@leolost2605 thats a good idea! I'll do that. Thanks :)

PR #2346, opened 2025-09-03
AppInfoView: don't allow focusing title labels

### LINKED ISSUE #1513: Text cursor should not appear for app names
## What Happened

In AppCenter, after clicking on a specific app, I'm presented with a page containing details of the app. When I click on the app name beside the icon, both of which are at the top of the page, a text cursor appears. However, the cursor should not appear because it is not a text field. The app name cannot be changed.

## Expected Behavior

The cursor should not appear because it is not a text field. The app name cannot be changed.

## Steps to Reproduce

1. Open AppCenter
2. Click on any app
3. Click on the app name at the top of the page
4. Look for the text cursor

## Platform Information

Screenshot from 2021-04-08 13-53-56

danirabbit:
Fixes #1513

### DIFF
--- a/src/Views/AppInfoView.vala
+++ b/src/Views/AppInfoView.vala
@@ -150,6 +150,7 @@ public class AppCenter.Views.AppInfoView : Adw.NavigationPage {
         };
 
         var app_title = new Gtk.Label (package.name) {
+            can_focus = false,
             selectable = true,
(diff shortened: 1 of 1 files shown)


leolost2605 (approved):
Can't test rn but LGTM

PR #2347, opened 2025-09-08
InstalledPackageRowGrid: put version number inline with name

danirabbit:
Put version info inline with the name, but smaller and dim

## BEFORE
Screenshot from 2025-09-08 14 10 55

## AFTER

Screenshot from 2025-09-08 14 08 25

### DIFF
--- a/data/appcenter.metainfo.xml.in
+++ b/data/appcenter.metainfo.xml.in
@@ -73,6 +73,7 @@
         </ul>
       </description>
       <issues>
+        <issue url="https://github.com/elementary/appcenter/issues/36">Sort installed tab by date installed</issue>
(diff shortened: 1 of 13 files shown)


jeremypw (approved):
Code is improvement and works as expected. Change makes sense.

jeremypw:
This partially fixes the issue but only gives info on Flatpaks? This is a fundamental design decision in Appcenter I guess and will become less important in future but now cannot see when Files, Code, Terminal and other important system apps  have been updated in Appcenter.  Is there any intention to expose this info?

jeremypw:
@danirabbit What I meant was is there any way of exposing the equivalent info elsewhere than Appcenter for components that do not appear there.

danirabbit:
@jeremypw there isn't. Non-Flatpak packages are considered to be part of the operating system image

PR #2363, opened 2025-10-11
Package transactions: Always set error

leolost2605:
The first three commits make sure we always set an error instead of returning false for the three package transactions. This way we make sure we never fail silently for the user. Also it simplifies not having to deal with a success variable as well.
There was also bug where != was used instead of ==. The comment said we don't want to overwrite existing errors but the code checked if there was an existing error and if there WAS it overwrote it instead of the other way round.

The fourth commit ignores the return value from the package ops since it will now always return true or set an error so it assumes success if no error is set.

And the last commit is a little cleanup that changes the backend methods to `void`.

This also serves as preparation for fixing cancelling in a way that we get immediate feedback when cancelling instead of just a flicker that looks like it doesn't cancel.

Can use rebase to merge.

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -1575,6 +1575,7 @@ public class AppCenterCore.FlatpakBackend : Object {
 
         var bundle = package.component.get_bundle (AppStream.BundleKind.FLATPAK);
         if (bundle == null) {
+            warning ("Failed to install package: component %s has no flatpak bundle", package.component.get_name ());
             job.error = new IOError.FAILED (_("Component has no flatpak bundle"));
             job.results_ready ();
             return;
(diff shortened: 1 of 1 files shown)


src/Core/FlatpakBackend.vala
```diff
@@ -1625,15 +1625,14 @@ public class AppCenterCore.FlatpakBackend : Object {
 
         var bundle = package.component.get_bundle (AppStream.BundleKind.FLATPAK);
         if (bundle == null) {
-            job.result = Value (typeof (bool));
-            job.result.set_boolean (false);
+            job.error = new IOError.FAILED (_("Component has no flatpak bundle"));
             job.results_ready ();
```
lenemter: Shouldn't this produce a critical/warning in the console as well?

src/Core/FlatpakBackend.vala
```diff
@@ -1752,8 +1739,7 @@ public class AppCenterCore.FlatpakBackend : Object {
 
         unowned var bundle = package.component.get_bundle (AppStream.BundleKind.FLATPAK);
         if (bundle == null) {
-            job.result = Value (typeof (bool));
-            job.result.set_boolean (false);
+            job.error = new IOError.FAILED (_("Component has no flatpak bundle"));
```
lenemter: Same here

PR #2373, opened 2025-10-31
Use activate functions for Package row parents

### LINKED ISSUE #2371: Revert change from list/flowbox activate to click controller
we've broken keyboard navigation by changing app row items to use a click controller instead of using the activate signal of list/flowbox/gridview/etc

We should stick to those activate signals to make sure all the possible activation cases are covered for various input methods

danirabbit:
Fixes #2371 

Can be rebase merged. Fixes non-mouse activation methods

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -85,3 +85,4 @@ zh_TW
 zu
 tr_TR
 nb_NO
+zh_Hant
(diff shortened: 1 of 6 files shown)


leolost2605 (commented):
Can we use the actionable interface that listboxrow and flowboxchild implement?

PR #2377, opened 2025-11-20
Update Flathub search URL to new format with locale support

srueegger:
The Flathub website has been redesigned with a new URL structure for search. This updates the dynamic search link to use the new format: https://flathub.org/{locale}/apps/search?q={query}

The system locale is automatically detected and the language code is extracted (e.g., "de" from "de_DE.UTF-8") to provide localized search results. Falls back to "en" if locale cannot be determined.

### DIFF
--- a/src/Views/SearchView.vala
+++ b/src/Views/SearchView.vala
@@ -145,7 +145,14 @@ public class AppCenter.SearchView : Adw.NavigationPage {
 
     private void search () {
         if (search_entry.text.length >= VALID_QUERY_LENGTH) {
-            var dyn_flathub_link = "<a href='https://flathub.org/apps/search/%s'>%s</a>".printf (search_entry.text, _("Flathub"));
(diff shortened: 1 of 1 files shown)


srueegger:
@ryonakano  That's a good question that I hadn't thought of. Currently, you also get an Error 404 if the language is not available on Flathub.

I see two possible solutions:

1. Here is a list of all available languages: https://flathub.org/en/languages

We can create a config file with the languages and use an if query to check whether the OS language matches one of the languages – if so, everything is fine; if not, simply fall back to English.
The disadvantage of this solution is that we will have to keep updating this config file.

2. We send a quick request to the URL using curl. If we receive a 404 status code, we use EN as a fallback. The disadvantage of this solution is that we establish a connection to Flathub from the computer without the user being aware of it.

Do you see any other options, and which solution would you prefer?

PR #2389, opened 2025-12-11
FlatpakBackend: don't feaure daily apps

### LINKED ISSUE #1715: The AppCenter allows for a confusing second installation of some pre-installed apps.
## Problem
Some Elementary applications (like Calendar, Tasks, Mail and Videos) that come pre-installed with the OS can be installed on top of their current installation for a second time through the AppCenter. Those apps are shown as already installed from the "elementary Updates" source, but they can be installed again via the "appcenter"  or "appcenter (system-wide)" source. 

This is not friendly for the new user as it does not make it clear whether the app is completely installed on their system. Furthermore, the names used for those two sources ("elementary Updates" and "appcenter") make the choice between the two difficult since the OS is called Elementary but the application used for the updates is called AppCenter. The lack of easily accessible(popular) documentation on this topic makes the experience even more confusing.

By installing the application for a second time things become even worse. More specifically, after installing the app for a second time, when the application is launched there are now two different windows showing up in the dock for the same application. More importantly, one of them is clickable and by clicking it it fires up another instance of the same application. This can be fixed by uninstalling the application from the AppCenter, however it is already obvious that the user experience is not ideal.

### Even bigger problem
The "issue" is even more prominent for the Elementary Calculator. Specifically for that app, the AppCenter does not show it as installed at all, letting you install it for a second time as if it was never installed in the first place. Of course by installing it for a second time the same issues arise as for the other apps. Again, by uninstalling Calculator things get fixed again, however in the case of this specific app the AppCenter looks like Calculator was never installed and uninstalling it will not actually remove it from the system ever.

## Possible fix
Since those apps are considered as parts of Elementary's suite of applications, I think it is a good idea to show whether they are installed or not in a clearer and more intuitive way. A good example of that is the Files or the Music applications. Those apps are clearly shown to be installed through the AppCenter with no confusion about the source that they came from, and at the same time AppCenter still lets you uninstall them if you want to.

### Steps to Reproduce

1. Install Elementary OS
2. Fully update the system to the latest version
3. Open AppCenter
4. Search for Calculator
5. Install Calculator (it is shown as not installed)
6. Run Calculator
- OR -
4. Go to the "Installed" tab
5. Go to the AppCenter's Calendar, Tasks, Mail or Videos page
6. Select "appcenter" as the source
7. Install the application
8. Run the application

### Expected Behavior

The Calculator application should be shown as installed in AppCenter as it comes pre-installed with the OS.

In my opinion, it would be nicer if all applications that come pre-installed with the system were given a clearer indication that they are already fully and correctly installed. The choice of the source could also get some improvement with clearer names for the sources that further clarify the differences between them. Finally, an approach like the one used for the Music application (for example) seems better to me for  programmes that come pre-installed with the OS. 

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
This doesn't fully resolve #1715 but at least make sure we don't show these on the homepage. While we're here, avoid some nesting

### DIFF
--- a/src/Core/FlatpakBackend.vala
+++ b/src/Core/FlatpakBackend.vala
@@ -581,7 +581,7 @@ public class AppCenterCore.FlatpakBackend : Object {
             }
 
             unowned var bundle = package.component.get_bundle (FLATPAK);
-            if (bundle?.get_id ().has_suffix ("daily")) {
+            if (bundle?.get_id ()?.has_suffix ("daily")) {
                 continue;
             }
 


src/Core/FlatpakBackend.vala
```diff
@@ -558,15 +558,34 @@ public class AppCenterCore.FlatpakBackend : Object {
...
+            if (
+                package.normalized_component_id == "io.elementary.calendar" ||
+                package.normalized_component_id == "io.elementary.mail" ||
+                package.normalized_component_id == "io.elementary.tasks"
+            ) {
+                continue;
+            }
+
+            unowned var bundle = package.component.get_bundle (FLATPAK);
+            if (bundle?.get_id ().has_suffix ("daily")) {
```
leolost2605: if bundle is null this would still call has_suffix on null wouldn't it?

I'm not sure if that would work because then there could be if (null) which would work but idk if vala allows it:
```suggestion
            if (bundle?.get_id ()?.has_suffix ("daily")) {
```

PR #2390, opened 2025-12-11
Tests: disable Soup test

danirabbit:
These tests never pass because https://httpbin.org/ is 503


ryonakano:
@danirabbit I personally think we should keep these tests so that we don't introduce bugs around here. What do you think about pulling and using the official docker image of httpbin in CI? #2391

danirabbit:
If it's working now I don't mind closing this PR. It was just really annoying for a while because every PR required admin override to merge 😅

PR #2391, opened 2026-01-01
Actions: Pull and run httpbin as a service container

ryonakano:
## Changes Summary
- Run httpbin as a service container instead of using the officially deployed one in CI so that the build and test won't fail when the official one is down
  - Add a build option to set the HTTP server used in the integration tests

FYI: https://docs.github.com/en/actions/tutorials/use-containerized-services/use-docker-service-containers

## Checklist
- [x] The CI passes
- [x] Local test using the default `https://httpbin.org/post` stil works (`meson setup builddir --prefix=/usr -Dintegration_tests=true && ninja -C builddir/ test`)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,8 +18,8 @@ jobs:
     container:
       image: ghcr.io/elementary/docker:${{ matrix.version }}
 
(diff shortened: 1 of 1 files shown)


ryonakano:
Closing as the officail httpbin works now.

PR #2394, opened 2026-01-22
Send feedback on uninstall

### LINKED ISSUE #2393: User feedback on "Uninstall"
### Problem

Clicking on "Uninstall" does... Nothing. There is no feedback something is happening. The menu just closes.
And if i reopen the menu and flatpak hasnt finished removing, its still there. So it feels bugged.

So uh. Did it do anything ?

There is also this bug: https://github.com/elementary/applications-menu/issues/612
Which also makes that feature confusing: Is anything going on ? Is it the bug ? Should i redo on terminal ?

### Proposal

Uninstall(5)

Show a progressbar overlaying the icon.
Grey out the icon and/or icon (here i did shadowing it) - Make it clear this is going out

I considered a red X as emblem instead, as alternative information, but progressbar is used in the dock to show something is going on.

I was going for a greyed out label in context menu, "Uninstall in progress", but i suppose a "cancel uninstallation" would be more... friendly? (Else we could straight up remove any uninstall functionality in appmenu, let people do it in appcenter, but let it be done here is quicker path)

### Prior Art (Optional)

no

danirabbit:
Fixes #2393

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -5,7 +5,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-10-13 06:00+0000\n"
+"POT-Creation-Date: 2026-01-29 16:32+0000\n"
 "PO-Revision-Date: 2022-04-20 09:08+0000\n"
 "Last-Translator: David Hewitt <davidmhewitt@gmail.com>\n"
 "Language-Team: Afrikaans <https://l10n.elementary.io/projects/appcenter/"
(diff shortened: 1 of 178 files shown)


src/Application.vala
```diff
@@ -322,6 +322,31 @@ public class AppCenter.App : Gtk.Application {
...
+                }
+
+                // Check if window is focused
+                if (active_window != null && active_window.is_active) {
+                    ((MainWindow) active_window).send_uninstalled_toast (package);
+                    break;
+                }
+
+                var notification = new Notification (_("The app has been uninstalled"));
+                notification.set_body (_("“%s” has been installed").printf (package.name));
```
vjr: Shouldn't this read "uninstalled" instead of "installed" ?

PR #2437, opened 2026-07-27
Memory leak: Application views + screenshots never release memory to the system

lewisgoddard:
Every AppInfoView (and all of its screenshots) stayed alive after being popped from the navigation view, so browsing leaked hundreds of MB that were never reclaimed. (I initially thought this was just a screenshot issue.)

Now AppInfoView is free'd within the app memory pool when navigating away, and closing the window releases all free'd memory to the system.

**Before:** Browing apps brought memory usage to 1.4GiB, not free'd on close.
**After:** Browsing apps brought memory usage to ~800MiB, free'd to ~500MiB on close.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -14,6 +14,9 @@
 * with this program. If not, see http://www.gnu.org/licenses/.
 */
 
+[CCode (cheader_filename = "malloc.h")]
+extern int malloc_trim (size_t pad);
+
 public class AppCenter.MainWindow : Gtk.ApplicationWindow {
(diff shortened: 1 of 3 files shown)


ryonakano:
How did you find this solution? `malloc_trim()` is a way to forcibly frees memory which I personally feel like a workaround.

lewisgoddard:
Only destroying the app views on close seemed to leave the memory allocated to the app, so I used `malloc_trim()`. I'm not super farmiliar with Vala, so I used what I know from C, hence the import. Would love to know what Vala method I probably should be using.

PR #2439, opened 2026-08-07
UpdatesView is a dialog

danirabbit:
Screenshot from 2026-08-10 11 17 05

* Updates and installed is a dialog
* Double click activate because it takes you away from this window

Todo:
- [x] Check for updates action is broken
- [x] Navigating to app pages is busted. Should close dialog
- [ ] Something about window title when scrolling?
- [ ] What's New should probably be a subpage instead of a dialog here?

### DIFF
--- a/src/Views/AppListUpdateView.vala
+++ b/src/Views/AppListUpdateView.vala
@@ -66,7 +66,6 @@ public class AppCenter.UpdatesDialog : Gtk.ApplicationWindow {
         updatable_header.append (update_all_button);
 
         list_box = new Gtk.ListBox () {
-            activate_on_single_click = true,
             hexpand = true,
         };
         list_box.bind_model (flatpak_backend.updatable_packages, create_row_from_package);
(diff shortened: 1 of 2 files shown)


src/Views/AppListUpdateView.vala
```diff
@@ -154,51 +157,49 @@ public class AppCenter.Views.AppListUpdateView : Adw.NavigationPage {
...
         toolbarview.add_top_bar (headerbar);
         toolbarview.add_css_class (Granite.STYLE_CLASS_VIEW);
 
+        titlebar = new Gtk.Grid () { visible = false };
         child = toolbarview;
+        hide_on_close = true;
         /// TRANSLATORS: the name of the Installed Apps view
         title = C_("view", "Installed");
+        default_height = 600;
+        default_width = 400;
```
danirabbit, line 186: Probably needs to be wider. Some apps have long names or version numbers

wpkelso: I'm in support of going wider. Perhaps some trickery with a small `default_width` and a `min-width` set in css, since we don't have a clamp/max-width? I think my ideal would put it to maybe ⅔rds the application window width?

PR #2444, opened 2026-08-26
FlatpakBackend: DRY mkdir if not exist

danirabbit:
Also add an early return. View with whitespace changes hidden


src/Core/FlatpakBackend.vala
```diff
@@ -1199,43 +1197,39 @@ public class AppCenterCore.FlatpakBackend : Object, Backend {
...
 
-                var remote_icons_folder = appstream_dir.get_child ("icons");
-                if (!remote_icons_folder.query_exists ()) {
-                    continue;
-                }
+            var local_icons_folder = dest_folder.get_child ("icons");
+            var remote_icons_folder = appstream_dir.get_child ("icons");
+            try {
+                mkdir_if_not_exists (local_icons_folder);
+                mkdir_if_not_exists (remote_icons_folder);
```
leolost2605, line 1211: Is it intentional that on main we `continue` if the remote folder doesn't exist but here it seems we create it?

leolost2605 (commented):
One question, apart from that makes sense :)

src/Core/FlatpakBackend.vala
danirabbit: Yes? Like I'm not sure why we were stopping there before instead of just making the folder?

leolost2605: Hmm I think it makes sense to just stop. As I understand it the remote icons folder is managed by flatpak. So if it doesnt exist why bother making links to something that is of no use because even if we create it there wont be any icons in there? We might as well just stop.
Though I must say I havent analyzed that part of the flatpak backend very much I always just accepted it as some black magic so I might be missing some bigger picture 😅

